PR #10, opened 2023-05-15
Add tests for Specimin

LoiNguyenCS:
Professor, I have written some test cases for `specimin`. I think `specimin` returns the correct result for `globalvariables`, `superclass`, and `threelayerfunctioncall`. For `callexistingmethod`, I'm not sure if that's the correct output. At the same time, I am not sure what the correct output is. So I haven't added the test file to test `callexistingmethod` yet, I have only added the input for testing so that we can discuss further.


src/test/java/org/checkerframework/specimin/GlobalVariables.java
```diff
@@ -0,0 +1,19 @@
+package org.checkerframework.specimin;
+
+import org.junit.Test;
+
+import java.io.IOException;
+
+/**
+ * This test checks that a simple Java file with no dependencies whatsoever and a single target
+ * method is returned unaltered by specimin.
```
kelloggm, line 9: For each new test, this comment should be modified to explain the test's purpose.

src/test/resources/callexistingmethod/expected/com/example/Simple.java
```diff
@@ -0,0 +1,10 @@
+package com.example;
+import java.util.Random;
+
+public class Simple {
+    public void test() {
+        Random random = new Random();
+        int randomNumber = random.nextInt(100);
```
kelloggm, line 7: We should discuss this case in our meeting tomorrow. I see two options:
* generate + populate a source file for `java.util.Random`, and don't link against the Java standard library
* generate + populate a stub file for `java.util.Random`, and continue to link against the standard library

The current spec doesn't fully require the latter, but I think it's the right thing to do exactly for classes that are in the JDK. For non-JDK classes that aren't from the given source code, I expect we'll want to do the former (so we'll need to support both eventually).

kelloggm: To motivate this a bit further, we're planning to use this to try out different combinations of annotations on the dependencies of the target method, so the stub file would be created specifically for that. We can leave that as a TODO for now.

src/test/resources/globalvariables/input/com/example/Simple.java
```diff
@@ -0,0 +1,9 @@
+package com.example;
+
+class Simple {
+    static int x = 5;
```
kelloggm, line 4: We also need a version of this test where `x` is defined in a different class.

src/test/resources/superclass/expected/com/example/Simple.java
```diff
@@ -0,0 +1,15 @@
...
+    public void printMessage() {
+        throw new Error();
+    }
+}
+
+public class Simple extends BigSimple {
+    @Override
+    public void printMessage() {
+        super.printMessage();
+        System.out.println("It's me");
```
kelloggm, line 13: This expected output isn't correct (or, at least, isn't a great test), because this won't compile without an implementation for `System.out.println`. Since that's in the JDK, it's probably okay not to print anything for now, but it would be better to avoid having these kind of complications if we can avoid it.

kelloggm (changes requested):

LoiNguyenCS:
Professor, I have updated the test files based on your review. Do you think that we should delete the `callexistingmethod` from this PR so that we can add them later in another PR where we have the codes to make a copy of the method as we discussed in our meeting?

kelloggm:
> Professor, I have updated the test files based on your review. Do you think that we should delete the `callexistingmethod` from this PR so that we can add them later in another PR where we have the codes to make a copy of the method as we discussed in our meeting?

Yes, let's move those tests into another branch for now. You can then write the code to handle them in that branch and make a PR with both the tests and the code when it is ready.

LoiNguyenCS:
Professor, I have deleted the `callexistingmethod`.

src/test/resources/superclass/input/com/example/Simple.java
```diff
@@ -0,0 +1,15 @@
+package com.example;
+
+class BigSimple {
+    public void printMessage() {
+        String x = "Hello";
```
kelloggm, line 5: Why even include this line of code at all? The method is `void`, so its body can be empty.

(And, regarding our conversation the other day about `System.out.println`, the same thing applies to `java.lang.String`: it is technically a class in the JDK, and therefore a dependency if you include it here.)

kelloggm (approved):

LoiNguyenCS:
Professor, I have updated the test for superclass. Please take a look. Thank you.

.DS_Store
kelloggm, line 1: Please don't commit binary files (you can add `.DS_Store` to the project's `.gitignore` file to avoid this problem in the future).

kelloggm (approved):
Code looks good (and is ready to merge), but there are extraneous binary files committed which need to be removed before merging.

.DS_Store
LoiNguyenCS: Professor @kelloggm , I have deleted the binary files. I think I accidentally add them during one of my commits. Thank you for pointing out.

kelloggm (approved):
LGTM

.DS_Store
kelloggm: Please make another PR to add this to the .gitignore so that it doesn't happen in the future.

PR #18, opened 2023-07-03
Add signature checker

LoiNguyenCS:
Professor, this is the PR to add signature checker to Specimin. Please take a look. Thank you

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -116,11 +116,16 @@ class JavaTypeCorrect {
 
   /**
    * This method takes the name of a class and converts it to the @ClassGetSimpleName type according
-   * to Checker Framework. If the name is already a simple name, this class will not alter it.
+   * to Checker Framework. If the name is already a simple name, this class will not alter it. To
+   * recall a bit, this method belongs to JavaTypeCorrect, where the main job is to correct the
+   * synthetic return type of unsolved methods. UnsolvedSymbolVisitor has take good note of the
+   * details of those synthetic types, so a simple name should be sufficient,
    *
    * @param className the name of the class to be converted
    * @return the simple name of the class
    */
+  // the code is self-explanatory, essentially the last element of a class name is the simple name
+  // of that class
   @SuppressWarnings("signature")
   public static @ClassGetSimpleName String toSimpleName(String className) {
     String[] classNameParts = className.split("[.]");
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -116,10 +116,8 @@ class JavaTypeCorrect {
 
   /**
    * This method takes the name of a class and converts it to the @ClassGetSimpleName type according
-   * to Checker Framework. If the name is already a simple name, this class will not alter it. To
-   * recall a bit, this method belongs to JavaTypeCorrect, where the main job is to correct the
-   * synthetic return type of unsolved methods. UnsolvedSymbolVisitor has take good note of the
-   * details of those synthetic types, so a simple name should be sufficient,
+   * to Checker Framework. If the name is already in the @ClassGetSimpleName form, this method will
+   * not make any changes
    *
    * @param className the name of the class to be converted
    * @return the simple name of the class
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -7,6 +7,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
 import org.checkerframework.checker.signature.qual.ClassGetSimpleName;
+import org.checkerframework.checker.signature.qual.DotSeparatedIdentifiers;
 
 /**
  * This class uses javac to analyze files. If there are any incompatible type errors in those files,
@@ -104,13 +105,18 @@ class JavaTypeCorrect {
      * 2. error: incompatible types: found <type1> required <type2>
      */
     if (errorMessage.contains("cannot be converted to")) {
-      typeToChange.put(
-          toSimpleName(splitErrorMessage[4]),
-          toSimpleName(splitErrorMessage[splitErrorMessage.length - 1]));
+      // since this is from javac, we know that these will be dot-separated identifiers.
+      @SuppressWarnings("signature")
+      @DotSeparatedIdentifiers String incorrectType = splitErrorMessage[4];
+      @SuppressWarnings("signature")
+      @DotSeparatedIdentifiers String correctType = splitErrorMessage[splitErrorMessage.length - 1];
+      typeToChange.put(toSimpleName(incorrectType), toSimpleName(correctType));
     } else {
-      typeToChange.put(
-          toSimpleName(splitErrorMessage[5]),
-          toSimpleName(splitErrorMessage[splitErrorMessage.length - 1]));
+      @SuppressWarnings("signature")
+      @DotSeparatedIdentifiers String incorrectType = splitErrorMessage[5];
+      @SuppressWarnings("signature")
+      @DotSeparatedIdentifiers String correctType = splitErrorMessage[splitErrorMessage.length - 1];
+      typeToChange.put(toSimpleName(incorrectType), toSimpleName(correctType));
     }
   }
 
@@ -123,9 +129,10 @@ class JavaTypeCorrect {
    * @return the simple name of the class
    */
   // the code is self-explanatory, essentially the last element of a class name is the simple name
-  // of that class
+  // of that class. This method takes the input from the error message of javac, so we know that
+  // className will be a dot-separated identifier.
   @SuppressWarnings("signature")
-  public static @ClassGetSimpleName String toSimpleName(String className) {
+  public static @ClassGetSimpleName String toSimpleName(@DotSeparatedIdentifiers String className) {
     String[] classNameParts = className.split("[.]");
     if (classNameParts.length < 2) {
       return className;
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -6,6 +6,7 @@ import java.io.InputStreamReader;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import org.checkerframework.checker.signature.qual.ClassGetSimpleName;
 
 /**
  * This class uses javac to analyze files. If there are any incompatible type errors in those files,
@@ -29,7 +30,7 @@ class JavaTypeCorrect {
    * This map is for type correcting. The key is the name of the current incorrect type, and the
    * value is the name of the desired correct type.
    */
-  private Map<String, String> typeToChange;
+  private Map<@ClassGetSimpleName String, @ClassGetSimpleName String> typeToChange;
 
   /**
    * Create a new JavaTypeCorrect instance. The directories of files in fileNameList are relative to
@@ -49,7 +50,7 @@ class JavaTypeCorrect {
    *
    * @return the value of typeToChange
    */
-  public Map<String, String> getTypeToChange() {
+  public Map<@ClassGetSimpleName String, @ClassGetSimpleName String> getTypeToChange() {
     return typeToChange;
   }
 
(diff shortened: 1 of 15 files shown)


src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
```diff
@@ -29,7 +30,7 @@ class JavaTypeCorrect {
    * This map is for type correcting. The key is the name of the current incorrect type, and the
    * value is the name of the desired correct type.
    */
-  private Map<String, String> typeToChange;
+  private Map<@ClassGetSimpleName String, @ClassGetSimpleName String> typeToChange;
```
kelloggm, line 33: It makes me very nervous that these are simple names, given our discussion just a few minutes ago about programs that use two classes with the same simple name but different fully-qualified names. For now, this is fine, but we should keep this in mind as we're testing to see if that example works correctly.

src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
```diff
@@ -103,9 +104,29 @@ private void updateTypeToChange(String errorMessage) {
...
+
+  /**
+   * This method takes the name of a class and converts it to the @ClassGetSimpleName type according
+   * to Checker Framework. If the name is already a simple name, this class will not alter it.
+   *
+   * @param className the name of the class to be converted
+   * @return the simple name of the class
+   */
+  @SuppressWarnings("signature")
+  public static @ClassGetSimpleName String toSimpleName(String className) {
```
kelloggm, line 125: Since it's under a warning suppression, we have to choose the annotations in this method without the use of the checker.

With that in mind, I think the parameter should be `@FullyQualifiedName String`. This method definitely does not produce a simple name for arbitrary string input, which is what the type says now.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -225,13 +228,48 @@ public static boolean calledByAnIncompleteSyntheticClass(MethodCallExpr method)
...
+   * that there is warning suppression here. It is safe to claim that if we split
+   * a @FullyQualifiedName name by dot ([.]), then the last part is the @ClassGetSimpleName part.
+   *
+   * @param fullyQualifiedName a @FullyQualifiedName classname
+   * @return the @ClassGetSimpleName version of that class
+   */
+  public static @ClassGetSimpleName String fullyQualifiedToSimple(
+      @FullyQualifiedName String fullyQualifiedName) {
+    @SuppressWarnings("signature")
+    @ClassGetSimpleName String simpleName = fullyQualifiedName.substring(fullyQualifiedName.lastIndexOf(".") + 1);
```
kelloggm, line 256: If you move the `@SuppressWarnings` to the method's declaration, you can remove the `@ClassGetSimpleName` annotation on this line.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -225,13 +228,48 @@ public static boolean calledByAnIncompleteSyntheticClass(MethodCallExpr method)
...
+   * Given the name of an unsolved method, this method will return the name of the synthetic return
+   * type for that method. The name is in @ClassGetSimpleName form
+   *
+   * @param methodName the name of a method
+   * @return that name in @ClassGetSimpleName form
+   */
+  public static @ClassGetSimpleName String returnNameForMethod(String methodName) {
+    String capitalizedMethodName =
+        methodName.substring(0, 1).toUpperCase() + methodName.substring(1);
+    @SuppressWarnings("signature")
```
kelloggm, line 270: Here and elsewhere, each warning suppression should have a short comment justifying it. Here, I think the justification is something like `// adding ReturnType to a method name always creates a synthetic class name`

kelloggm (changes requested):

LoiNguyenCS:
Professor, I have added codes and test cases for special class call such as `org.second.Calculator()`. For those special calls, Specimin will rely on the call itself to figure out the class and package name. For other cases, Specimin will rely on the class name and the import statements in the input.

src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
```diff
@@ -116,11 +116,16 @@ private void updateTypeToChange(String errorMessage) {
 
   /**
    * This method takes the name of a class and converts it to the @ClassGetSimpleName type according
-   * to Checker Framework. If the name is already a simple name, this class will not alter it.
+   * to Checker Framework. If the name is already a simple name, this class will not alter it. To
+   * recall a bit, this method belongs to JavaTypeCorrect, where the main job is to correct the
+   * synthetic return type of unsolved methods. UnsolvedSymbolVisitor has take good note of the
+   * details of those synthetic types, so a simple name should be sufficient,
```
kelloggm, line 122: This Javadoc comment seems confusing to me, and I'm not sure exactly what it is trying to say. 

As I understand it, the purpose of this method is to take a Java simple or fully-qualified name, and turn it into just the simple name. If that's correct, just that would be sufficient Javadoc. Further, though, the specification should also say the same thing. Right now, the parameter type is just `String`, but I don't think this method works unless the input is [dot-separated](https://checkerframework.org/api/org/checkerframework/checker/signature/qual/DotSeparatedIdentifiers.html).

I assume that adding an annotation to the parameter causes the checker to reject code elsewhere. I'd be okay with leaving the parameter of this method unannotated (if that's easier at call sites) if and only if this class adds some defensive coding to check that the input is actually a dot-separated identifier or a simple name already. As it is, if I passed this an actual arbitrary string (as opposed to a class name), it will not fail but will instead mangle that string, which will be a hard-to-debug problem.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -391,6 +400,115 @@ public void createMissingClass(UnsolvedClass missedClass) {
...
+
+  /**
+   * This method checks if a string is capitalized
+   *
+   * @param string the string to be checked
+   * @return true if the string is capitalized
+   */
+  public static boolean isCapital(String string) {
+    String capitalizedString = toCapital(string);
+    return string.equals(capitalizedString);
```
kelloggm, line 421: This also returns true if the input is all uppercase (e.g., `"LIKE THIS"`) or if the string starts with a non-letter character (e.g., `5-ary` would be considered capitalized by this code). Are these quirks intentional?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -391,6 +400,115 @@ public void createMissingClass(UnsolvedClass missedClass) {
...
+  }
+
+  /**
+   * This method checks if a string is a class path.
```
kelloggm, line 425: I don't think this comment is accurate. This method checks if a string _has the form_ of a class path: that is, whether it _could_ be a classpath. It does not actually check that the string does represent a real class.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -391,6 +400,115 @@ public void createMissingClass(UnsolvedClass missedClass) {
...
+
+  /**
+   * Given a class path, this method will create a simple instance of UnsolvedClass for that class
+   * path
+   *
+   * @param classPath the class path
+   * @return the corresponding instance of UnsolvedClass
+   */
+  public static UnsolvedClass getSimpleSyntheticClassFromClassPath(String classPath) {
+    if (!isAClassPath(classPath)) {
```
kelloggm, line 444: Rather than `class path` here to describe this, I'd be more comfortable with this code if it was using one of the Signature Checker annotations to describe what this is. I _think_ your definition of `class path` is equivalent to a fully-qualified name (but I'm not 100% sure, which is why I'd like it if we use an annotation, which is definitely unambiguous).

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -391,6 +400,115 @@ public void createMissingClass(UnsolvedClass missedClass) {
...
+      throw new RuntimeException(
+          "Check with isAClassPath first before using getSimpleSyntheticClassFromClassPath");
+    }
+    String[] elements = classPath.split(".");
+    // the last element of a class path is the simple name of that class
+    @SuppressWarnings("signature")
+    @ClassGetSimpleName String className = elements[elements.length - 1];
```
kelloggm, line 451: Use a helper method (with a suppression) rather than suppressing this directly, so that the conversation from "class path" to simple name only needs to happen in one place.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -391,6 +400,115 @@ public void createMissingClass(UnsolvedClass missedClass) {
...
+    }
+  }
+
+  /**
+   * For a method call that is not simple, this method will take that method as input and create corresponding synthetic class
+   * @param methodCall the method call to be taken as input
+   */
+  public void updateClassSetWithNotSimpleMethodCall(String methodCall) {
+    String methodCallWithoutParen = methodCall.replace("()", "");
+    String[] methodParts = methodCallWithoutParen.split("[^a-zA-Z]+");
```
kelloggm, line 486: Method names can include numbers (e.g., `test1()`, `test2()`, etc.). Can you add a test case with a number in the name of a method that is called in a non-simple method call and make sure that this code handles it properly?

We may also need to handle non-ascii characters. Here's the relevant part of the JLS: https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.8 (note that most unicode are `JavaLetter`s, and therefore valid in names, so we need to be able to handle them).

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: Is there a library method that we could use instead? At a minimum, you could get the first character from the string and then check `Character#isUpperCase` directly rather than doing this equality test, which I think avoids most (all?) of the problems I mentioned.

src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
LoiNguyenCS: Professor, can you give me some suggestions on how to do the defensive check? I did try to do that the other day, but I'm not sure how to really check if a string is dot-separated.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: Yes, I think we can switch the codes to check the first character only.

src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
kelloggm: Since we have JavaParser available, one option would be to try to parse the string as if it were a Java expression, and then make sure that the resulting CompilationUnit contains a single identifier.

LoiNguyenCS: Can we instead clarify the Javadoc of this method? Since it takes the type complaining of javac as input (something like "Find /type1/ required /type2/", we can be sure that the input will be a valid string representation of a Java type instead of some ambiguous strings?

kelloggm: I think that would be alright. In that case, though, we should add a type annotation indicating that the input is a valid type (and its format); in this case, I think that would be `@FullyQualifiedName` (or whatever). Then, we'll need to suppress a warning when we convert javac's output to that type, but that will be easy for a human to verify.

LoiNguyenCS: Professor, this method convert a possibly fully-qualified name to a simple name. If the name is simple already, this method will return the name itself. Where should we add the type-qualifiers in this case?

kelloggm: I see. In that case, maybe use the `@DotSeparatedIdentifier` annotation (or one of the others that are lubs in the Signature Checker hierarchy)?

LoiNguyenCS: Got it. Thank you, Professor.

LoiNguyenCS:
Professor, I have updated the codes based on your suggestion. Please take a look. Thank you/

kelloggm (approved):

PR #20, opened 2023-07-11
Add parameter

LoiNguyenCS:
Professor, this is the PR to add codes for unsolved methods with parameters. I accidentally added some unrelated files. My Git seems to work in a very strange way. I apologize for any inconvenience caused by those unrelated files. I have already removed them. Thank you.


src/main/java/org/checkerframework/specimin/UnsolvedMethod.java
```diff
@@ -60,6 +61,29 @@ public void setReturnType(@ClassGetSimpleName String returnType) {
    * @return the content of the method with the body stubbed out
    */
   public String toString() {
-    return "\n    public " + returnType + " " + name + "() {\n        throw new Error();\n    }\n";
+    List<String> letters = new ArrayList<>();
+
+    for (char c = 'a'; c <= 'z'; c++) {
+      letters.add(String.valueOf(c));
+    }
```
kelloggm, line 68: I have some concerns about this approach to naming parameters:
* it is limited to 26 parameters
* it is wasteful: the same List (`letters`) is constructed every time

What about the following alternative strategy: name each parameter `parameter#X`, where `#X` is the loop variable (i.e., `i` in the loop on line 70). This is simpler and I think avoids both problems with the current approach.

kelloggm (changes requested):
This PR appears to (accidentally?) delete lots of existing tests, which needs to be corrected.

src/main/java/org/checkerframework/specimin/UnsolvedMethod.java
LoiNguyenCS: I think that is a good strategy, Professor. I will update the codes.

LoiNguyenCS:
Professor, I have updated the codes based on your suggestion. The commit [30ffe53](https://github.com/kelloggm/specimin/pull/20/commits/30ffe5318827711d916fe8106a13db454790026d) is simply to deal with the files accidentally added earlier.

src/test/resources/syntheticobjectcreation/expected/org/can/not/be/real/Pen.java
```diff
@@ -0,0 +1,7 @@
+package org.can.not.be.real;
+public class Pen {
+
+    public  Pen(org.still.not.real.Ink parameter0) {
```
kelloggm, line 4: Is it possible to fix the way that we print these so that there is not an awkward double-space before the name of the constructor?

kelloggm (changes requested):
The older test files (e.g., `src/test/java/org/checkerframework/specimin/CallJDK.java`, etc.) are still being deleted. To check that you've restored them, you should inspect the "Files changed" tab of the PR. To fix it, you may need to run a `git checkout` command to bring back the old versions of the files. For example, to get the version of `CallJDK.java` on the `main` branch, you'd run `git checkout origin/main -- src/test/java/org/checkerframework/specimin/CallJDK.java`.

LoiNguyenCS:
Professor, those test files are accidentally added, as you can see from path of those files, they are not belong to the `resources` directory of `test`. Sorry for the inconvenience. To be honest, I have no ideas how those files ended up being there.

src/test/resources/syntheticobjectcreation/expected/org/can/not/be/real/Pen.java
LoiNguyenCS: I think we can. I will try that.

kelloggm:
> Professor, those test files are accidentally added, as you can see from path of those files, they are not belong to the resources directory of test. Sorry for the inconvenience. To be honest, I have no ideas how those files ended up being there.

Ah, I misunderstood you before. Thanks for clarifying.

LoiNguyenCS:
Professor, I have updated the codes to get rid of the white space for constructor methods.

kelloggm (approved):

PR #21, opened 2023-07-14
Re add testfiles

LoiNguyenCS:
Professor, for some reasons, I have deleted all of our test files in the past PR. This is the PR to re-add the test files.

### DIFF
--- a/src/test/java/org/checkerframework/specimin/CallJDK.java
+++ b/src/test/java/org/checkerframework/specimin/CallJDK.java
@@ -1,0 +1,19 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.Test;
+
+/**
+ * This test checks that if the targeted method calls a method imported from java.util library,
+ * Specimin will not create synthetic file for that method and only return the targeted file with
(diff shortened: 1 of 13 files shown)


kelloggm (approved):
I pointed this out in #20 but you assured me [here](https://github.com/kelloggm/specimin/pull/20#issuecomment-1631258521) that these files were added accidentally. That you're re-adding them now, I assume, means that that was not correct?

How can we tell that something like this has happened in the future? I would suggest that 1) no changes are made to `main` except via a PR, and 2) we don't ever merge PRs that add or delete large swaths of files unless we are very sure that we intended to add/delete those files.

LoiNguyenCS:
Professor,

> `That you're re-adding them now, I assume, means that that was not correct?

Yes, I have mixed the file paths and got confused.

> I would suggest that 1) no changes are made to main except via a PR, and 2) we don't ever merge PRs that add or delete large swaths of files unless we are very sure that we intended to add/delete those files.

I believe this has been our procedure for a while. But I will by more careful in the future when it comes to add/delete files.




`

PR #22, opened 2023-07-15
Add codes for extension classes

LoiNguyenCS:
Professor, 

This is the PR to add codes for extension classes. At first I tried to make this into two PRs, but since I have found the way for `TargetMethodFinderVisitor` to notice the variables with the super keywords, I think it is easier if we have all of them in a single PR. Please take a look. Thank you.


LoiNguyenCS:
Sorry for the long list of commits. Since this PR is quite complicated, I'm not confident enough to do squash commit :)

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -187,6 +328,305 @@ public Visitable visit(MethodCallExpr method, Void p) {
...
+    } else {
+      updateUnsolvedClassWithVariables(
+          expr.asFieldAccessExpr().getNameAsString(),
+          parentClass,
+          classAndPackageMap.getOrDefault(parentClass, this.currentPackage));
+    }
+  }
+
+  /**
+   * This method will add a new variable declaration to a synthetic class. This class is mainly used
```
kelloggm, line 600: In the second sentence, should "this class" be "this method"? Or did I misunderstand what you mean?

src/test/resources/syntheticsupervariables/input/com/example/Dog.java
```diff
@@ -0,0 +1,14 @@
+package com.example;
+import org.wild.Mammal;
+
+public class Dog extends Mammal {
+    String habitat;
+    boolean canBreathUnderWater;
+    public void Dog() {
```
kelloggm, line 7: Is this supposed to be a constructor? If so, it's not correct: constructors should not have a return type, but this one has `void`.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -156,7 +162,13 @@ public static void main(String... args) throws IOException {
     for (Entry<String, CompilationUnit> target : parsedTargetFiles.entrySet()) {
       // If a compilation output's entire body has been removed, do not output it.
       if (isEmptyCompilationUnit(target.getValue())) {
-        continue;
+        boolean isASyntheticReturnType = target.getKey().contains("ReturnType");
```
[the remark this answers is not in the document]
LoiNguyenCS: I think we can try having `UnsolvedSymbolVisitor` to tell us the list of "ReturnType" classes it created. I will check to see if that approach works..

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -38,14 +51,34 @@
  */
 public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
 
+  /**
+   * The parent class of this current class file. If there is no parent class, then the value of
+   * this variable is an empty string
```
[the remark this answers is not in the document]
LoiNguyenCS: I think I will choose to clarify the javadoc for this field. I will change it to "The unsolved parent class of this current class file.", so that people will understand that this field is used to create synthetic parent class if needed.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -134,6 +139,23 @@ public Visitable visit(ClassOrInterfaceDeclaration decl, Void p) {
     return result;
   }
 
+  @Override
+  public Visitable visit(ConstructorDeclaration method, Void p) {
+    String constructorMethodAsString = method.getDeclarationAsString(false, false, false);
+    String methodName =
+        this.classFQName
+            + "#"
+            + constructorMethodAsString.substring(constructorMethodAsString.indexOf(' ') + 1);
```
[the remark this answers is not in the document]
LoiNguyenCS: Professor,

> Does this rely on the string representation having an access modifier (e.g., public)? If so, that's not okay, because no access modifier is valid Java (meaning "package private").

It does not rely on the modifier. The methodName will look something like: ```com.example.Car#Car()```. I will update the comments.

LoiNguyenCS:
Professor, I have updated the codes based on your comments. Please take a look. Thank you.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -142,6 +142,7 @@ public Visitable visit(ClassOrInterfaceDeclaration decl, Void p) {
   @Override
   public Visitable visit(ConstructorDeclaration method, Void p) {
     String constructorMethodAsString = method.getDeclarationAsString(false, false, false);
+    // the methodName will be something like this: "com.example.Car#Car()"
```
kelloggm, line 145: This does not fully address my concern, which was about `constructorMethodAsString` and the `indexOf` operation on line 149. Why is it correct to only take the part of that string after the first space?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -589,30 +610,30 @@ public void updateSyntheticClassForSuperCall(Expression expr) {
...
     }
   }
 
   /**
-   * This method will add a new variable declaration to a synthetic class. This class is mainly used
-   * for unsolved parent class. The declaration of the variable in the parent class will be the same
-   * as the declaration in the child class since Specimin does not have access to much information.
-   * If the variable is not found in the child class, Specimin will create a synthetic class to be
-   * the type of that variable.
+   * This method will add a new field declaration to a synthetic class. This method is mainly used
```
kelloggm, line 621: "mainly" here is ambiguous. Is it only ever used for unsolved parent classes, or are there other intended users?

Maybe a better word here is "intended", as in "it is intended to be used when the parent class is unsolved" or something similar.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -589,30 +610,30 @@ public void updateSyntheticClassForSuperCall(Expression expr) {
...
+   * for unsolved parent class. The declaration of the field in the parent class will be the same as
```
kelloggm, line 622: Throughout, it might be better to use the term "superclass" rather than "parent class". "Superclass" is a more technical word, but I think it is correct to describe what you mean by "parent class" and so should be preferred, rather than using a new term.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
LoiNguyenCS: You're right. That `indexOf` operation does not seem to make sense in this context. I think I put it there by mistake.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: I'm a bit confused. As I understand, I should keep "parent class" but explain in the document what I mean by "parent class". So I should add a line to this Javadoc saying that a parent class is the super class of the currently visited class. Is that correct, Professor?

kelloggm: > I'm a bit confused. As I understand, I should keep "parent class" but explain in the document what I mean by "parent class".

It depends on what exactly you mean by "parent class":
* if "parent class" means exactly the same as "superclass", you should remove all mentions of "parent class" and say "superclass" instead, because "superclass" is the standard term
* if "parent class" means something slightly different than "superclass", you should explain what you mean in the Javadoc

I'm not sure which of those is true, which I think is why my comment was confusing - I was trying to cover both possibilities.

LoiNguyenCS: Thank you for the explanation.

LoiNguyenCS:
Professor, I have updated the codes. Please help me review it. Thank you.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -71,6 +104,15 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
...
   private String chosenPackage = "";
 
+  /** This set has fully-qualified class names that come from jar files input */
+  private Set<@FullyQualifiedName String> classesFromJar = new HashSet<>();
+
+  /**
+   * This set has the fully-qualfied name of the synthetic return types created by this instance of
+   * UnsolvedSymbolVisitor
+   */
+  private Set<String> syntheticReturnTypes = new HashSet<>();
```
kelloggm, line 114: I think this field (and the others in this class) can also be `final`. Adding `final` is the best practice in Java unless the field will be re-assigned, so we should get into the habit of doing so.

kelloggm (approved):
This is ready to be merged as soon as `final` is added to those fields.

LoiNguyenCS:
Professor, I have added the final keyword.

kelloggm (approved):

PR #23, opened 2023-09-20
Codes for fields from super class

LoiNguyenCS:


LoiNguyenCS:
Professor, these are the codes and test cases for undeclared variables in a class. I assume that if a class has an undeclared variable, then that variable must be a field of the super class of that class. Please take a look.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -200,10 +221,21 @@ public Visitable visit(ExplicitConstructorInvocationStmt expr, Void p) {
...
     if (caller instanceof SuperExpr) {
       usedClass.add(caller.calculateResolvedType().describe());
     }
     return super.visit(expr, p);
   }
+
+  @Override
+  public Visitable visit(NameExpr expr, Void p) {
+    if (!declaredNames.contains(expr.getNameAsString())) {
+      usedClass.add(parentClass);
```
kelloggm, line 237: This concerns me, because I don't think we have a guarantee that all variable declarators are visited before all names. In particular, the following is legal Java code:
```
public class OutOfOrder {
    public static void use() {
	System.out.println(a);
    }

    static String a = "hello world";
}
```

In this code, `a` is used before it is declared, so I think Specimin might be tricked by this case. Can you add a test case like this to make sure that your approach works, even on code with a non-standard order of declarations?

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
```diff
@@ -108,6 +108,26 @@ public void updateMethodByReturnType(
...
 
+  /**
+   * This method updates the types of fields in this class
+   *
+   * @param currentType the current type
+   * @param correctType the desired type
+   */
+  public void updateFieldByType(String currentType, String correctType) {
+    for (String fieldExpression : classFields) {
+      String[] elements = fieldExpression.split(" ");
```
kelloggm, line 119: What if the field has a type annotation on the type? Specimin needs to preserve type annotations (do we have a test for that?); if there is one, I'd expect more than one space.

For example, what if the original field declaration is something like `@Nullable Foo foo;`?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -343,7 +357,7 @@ public Visitable visit(Parameter parameter, Void p) {
     try {
       parameter.resolve().describeType();
       return super.visit(parameter, p);
-    } catch (UnsolvedSymbolException e) {
+    } catch (Exception e) {
```
kelloggm, line 360: Why the change to catch an arbitrary exception? I'd prefer to avoid catching `Exception`, since that will also catch e.g., `NullPointerException` and other run time problems, not just unsolved symbols. If there are two kinds of exceptions you want to catch, you can use the `|` syntax described here: https://docs.oracle.com/javase/8/docs/technotes/guides/language/catch-multiple.html

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -603,11 +620,18 @@ public void updateSyntheticClassForSuperCall(Expression expr) {
...
+    } else if (expr instanceof FieldAccessExpr) {
       updateUnsolvedClassWithFields(
           expr.asFieldAccessExpr().getNameAsString(),
           superClass,
           classAndPackageMap.getOrDefault(superClass, this.currentPackage));
+    } else if (expr instanceof NameExpr) {
+      updateUnsolvedClassWithFields(
+          expr.asNameExpr().getNameAsString(),
+          superClass,
+          classAndPackageMap.getOrDefault(superClass, this.currentPackage));
```
kelloggm, line 632: Since this block is the same as the one above, you should just check `expr instanceof FieldAccessExpr || expr instanceof NameExpr` and avoid duplicating this code.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
LoiNguyenCS: I don't expect to see any annotation here, since these fields are inside synthetic classes. Synthetic classes are created by Specimin for unsolved symbols. These symbols are unsolved because the corresponding class files do not exist in the root directory physically. Specimin will create synthetic versions for these classes so that we will end up with a minimized version of the input codes that can be compiled.

LoiNguyenCS: So if this is the input codes:

```
package org.example;

class Child extends Parent {
  public @NonNull String getLastName() {
     return lastName;
   }
}
```
Then the synthetic class for the Parent class will be: 
```
package org.example;

class Parent{
  public String lastName = null;
}
```

As I understand, it's the job of the user to look that this result and realize that the field `lastName` in the real class file (not the synthetic file) of the Parent class must not be null. So he must look up the codes of the Parent class to make sure that the field `lastName` is not null so that his type reasoning can be valid.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
LoiNguyenCS: Thank for pointing this out. I did not realize that a variable can be used before it is declared. I will try to think of another way to get around this.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: Professor, they are actually different. For `FieldAccessExpr`, the first parameter for the method `updateUnsolvedClassWithFields` is `expr.asFieldAcessExpr().getNameAsString()`, while for `NameExpr`, the first parameter is `expr.asNameExpr().getNameAsString()`

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -117,6 +125,13 @@ public Set<String> getTargetMethods() {
 
   @Override
   public Visitable visit(ClassOrInterfaceDeclaration decl, Void p) {
+    if (decl.getExtendedTypes().isNonEmpty()) {
+      // note that since Specimin does not have access to the class paths of the project, all the
+      // unsolved methods related to inheritance will be placed in the parent class, even if there
+      // is a grandparent class and so forth.
+      parentClass = decl.getExtendedTypes().get(0).resolve().getQualifiedName();
```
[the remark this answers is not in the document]
LoiNguyenCS: I haven't thought of this case. If there are inner classes, it will definitely break the algorithm. Thank you for pointing out. I will try change parentClass to a map rather than a String to see if it helps.

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
kelloggm: Thanks, this makes sense. Can you add a comment at this point in the code that reminds the reader that `fieldExpression` is guaranteed not to have an annotation, because it's part of a synthetic class? That's why it is safe to assume that the result of `split` has exactly two elements.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: Ah, you're right. Even so, there is too much duplicate code here, so it's hard for a reader to see what's different. I suggest adding one line to each block that computes the thing that's different (e.g., `String nameAsString = expr.asNameExpr().getNameAsString()` vs `expr.asFieldAcessExpr().getNameAsString()`) to draw the attention of a reader to the difference.

LoiNguyenCS:
Professor, I have updated the program based on your suggestions. Please take a look.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -164,8 +164,8 @@ public static void main(String... args) throws IOException {
       if (isEmptyCompilationUnit(target.getValue())) {
         boolean isASyntheticReturnType = addMissingClass.isASyntheticReturnType(target.getKey());
         boolean isASyntheticSuperClass =
-            !addMissingClass.getSuperClass().equals("")
-                && target.getKey().contains(addMissingClass.getSuperClass());
+            !addMissingClass.getSuperClass().isEmpty()
+                && addMissingClass.getSuperClass().contains(target.getKey());
```
kelloggm, line 168: It's better style to avoid calling the same method twice in a conditional. In this case, you can abstract `addMissingClass.getSuperClass()` into a local variable on the line before.

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
```diff
@@ -108,6 +108,28 @@ public void updateMethodByReturnType(
...
+  /**
+   * This method updates the types of fields in this class
+   *
+   * @param currentType the current type
+   * @param correctType the desired type
+   */
+  public void updateFieldByType(String currentType, String correctType) {
+    for (String fieldExpression : classFields) {
+      // fieldExpression is guaranteed not to have an annotation, because it's part of a synthetic
+      // class
```
kelloggm, line 120: While this comment is useful, it doesn't directly explain what it is relevant to.

The key problem I'm concerned about here is the unguarded `[0]` and `[1]` accesses to the resulting array on lines 122 and 123 below. These would be unsafe if `fieldExpression` were an arbitrary String. However, they are safe because `fieldExpression` is guaranteed to have the form "TYPE FIELD_NAME", where both "TYPE" and "FIELD_NAME" have no spaces.

The comment here is a justification for why "TYPE" in the paragraph above has no spaces: there are no annotations, so it is a simple type name. However, that connection isn't clear from the comment, so I think in the current form it just makes this code more confusing (since it's not clear to a reader how it is related).

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -55,11 +58,14 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
    * The unsolved superclass of this current class file. If there is no unsolved superclass, then
    * the value of this variable is an empty string
    */
-  private @ClassGetSimpleName String superClass = "";
+  private final Map<String, @ClassGetSimpleName String> classAndItsParent = new HashMap<>();
```
kelloggm, line 61: This change also requires a change to the comment about this field, which is now inaccurate (a map cannot have the value of the empty string).

A comment on a map that is a field should always at a minimum explain what the keys and the values of the map represent.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -55,11 +58,14 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
...
 
   /** The package of this class */
   private String currentPackage = "";
 
+  /** The name of the class currently visited */
+  private String className = "";
```
kelloggm, line 67: Is this a simple or fully-qualified name? (The docs should say, or you could document it with an annotation.)

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -182,8 +193,8 @@ private void setclassAndPackageMap() {
    *
    * @return superClass the value of superClass
    */
-  public String getSuperClass() {
-    return superClass;
+  public Collection<@ClassGetSimpleName String> getSuperClass() {
+    return classAndItsParent.values();
```
kelloggm, line 197: It is surprising that a method called `getSuperClass` returns a collection. The documentation should explain what the meaning of the returned collection is, and how to actually get the name of the super class.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -17,15 +17,12 @@
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
kelloggm, line 20: Avoid wildcard imports. (You can and should change your IDE's settings, which are probably what triggered this change - I have IntelliJ set to automatically insert a wildcard import only if there are more than 10,000 imports from a single package, which never occurs in practce.)

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -55,16 +55,18 @@
 public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
 
   /**
-   * The unsolved superclass of this current class file. If there is no unsolved superclass, then
-   * the value of this variable is an empty string
+   * This map associates class names with their respective superclasses. The keys in this map
+   * represent the classes of the currently visited file. Due to the potential presence of inner
+   * classes, there may be multiple pairs of class and superclass entries in this map. This map can
+   * also be empty if there are no superclasses involved in the currently visited file.
```
kelloggm, line 61: It's impossible for there to be no superclasses: all classes implicitly have `java.lang.Object` as their superclass. I think you can clarify this text to fix this: it's not a problem that `Object` isn't in this map.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -189,9 +191,11 @@ private void setclassAndPackageMap() {
   }
 
   /**
-   * Get the value of superClass
+   * Get the collection of superclasses. It's important to note that due to the potential presence
```
kelloggm, line 194: You can drop "It's important to note that" from this comment, since the fact that you added this information to the comment at all should convey that it is important on its own.

PR #24, opened 2023-09-28
separate minimization part

LoiNguyenCS:
Professor, this is the PR to separate the main method of SpeciminRunner as we've discussed. Please let me know if there is anything I need to change.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
+++ b/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
@@ -69,9 +69,9 @@ public class SpeciminRunner {
   }
 
   /**
-   * This method provides a simplified way for users to perform the minimization job using Specimin.
-   * Users can call this method directly instead of invoking the main method of Specimin to initiate
-   * the minimization process.
+   * This method acts as an API for users who want to incorporate Specimin as a library into their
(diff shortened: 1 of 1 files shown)


src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -59,10 +59,33 @@ public static void main(String... args) throws IOException {
...
+        options.valueOf(outputDirectoryOption));
+  }
 
-    String root = options.valueOf(rootOption);
-    List<String> targetFiles = options.valuesOf(targetFilesOption);
-    List<String> jarPaths = options.valuesOf(jarPath);
+  /**
+   * This method provides a simplified way for users to perform the minimization job using Specimin.
+   * Users can call this method directly instead of invoking the main method of Specimin to initiate
+   * the minimization process.
```
kelloggm, line 74: This comment needs to mention that this is intended as an API for users who want to use Specimin as part of their own project. You've got most of that in the comment, but you should come out and say it directly. In particular, the phrases "API" and "use Specimin as a library" both need to appear, because those are what potential users will search for.

kelloggm (changes requested):

LoiNguyenCS:
Professor, I have updated the javadoc.

kelloggm (approved):

PR #25, opened 2023-09-29
Maven Publishing

LoiNguyenCS:


LoiNguyenCS:
Professor, this is my attempt to publish Specimin to a local Maven repository. I must admit that I find Gradle's [documentation](https://docs.gradle.org/current/userguide/publishing_maven.html) a bit difficult to grasp, so I am not very confident about this PR. There are sections in the build file that are puzzling to me, and I've added comments to highlight those areas.

As for publishing Specimin to Maven Central, I couldn't find clear instructions in Gradle's documentation (or maybe I misunderstood it). I found upon this [link](https://www.albertgao.xyz/2018/01/18/how-to-publish-artifact-to-maven-central-via-gradle/) , which suggests that we need to first publish our project to an approved repository like Sonatype OSS Nexus. I think this step might involve several authentication processes, so I believe it might be best if you handle it. However, if you prefer, I'd be happy to give it a shot.

kelloggm (changes requested):
Let's focus this PR on just the changes necessary to get `./gradlew publishToMavenLocal` working. You can test whether this works by running `./gradlew publishToMavenLocal` and then checking if there is a jar file for specimin in your local Maven repository, which will be located at `~/.m2/repository`.

As an example, [this build file](https://github.com/typetools/templatefora-checker/blob/master/build.gradle) contains just what's needed for local publication.

With respect to publishing onto Maven Central: I'd like to do that, too, but it is much less urgent.

LoiNguyenCS:
Professor, I have updated the build file to be able to do local maven publishing. I have tried publishing Specimin myself and there are jar files in my local maven repository. So I think we're good to go.

kelloggm (approved):

LoiNguyenCS:
Thank you for helping me remove those unused lines.

PR #29, opened 2023-10-04
removing comments from input codes

LoiNguyenCS:


LoiNguyenCS:
Professor, this is the PR to remove all the comments from the input codes.

There are some past test files that include comments, so I have run them through this new version of Specimin to have the new expected results for them.

I have also modified `SpeciminTestExecutor` to ignore blank lines and white spaces. As `PrintWriter` has a bit different in spacing compared to `LexicalPreservingPrinter`, I think ignoring blank lines and spacing is a better method compared to re-write the expected results for every test file.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -191,20 +187,20 @@ public static void performMinimization(
...
-        Files.write(
-            targetOutputPath,
-            LexicalPreservingPrinter.print(target.getValue()).getBytes(StandardCharsets.UTF_8));
+        // Create any parts of the directory structure that don't already exist.
+        Path dirContainingOutputFile = targetOutputPath.getParent();
+        // This null test is very defensive and might not be required? I think getParent can
+        // only return null if its input was a single element path, which targetOutputPath
+        // should not be unless the user made an error.
+        if (dirContainingOutputFile != null) {
+          Files.createDirectories(dirContainingOutputFile);
```
kelloggm, line 198: Why were these lines moved inside the `try` statement? It's good style to keep the size of `try` blocks to a minimum.

src/test/java/org/checkerframework/specimin/SpeciminTestExecutor.java
```diff
@@ -90,6 +90,8 @@ public static void runTest(
       builder.command(
           "diff",
           "-qr",
+          "-w",
+          "-B",
```
kelloggm, line 94: Why do you need `-w` and `-B`? I thought `-w` implied `-B`, but I might be misremembering the various `diff` options.

kelloggm (changes requested):

src/test/java/org/checkerframework/specimin/SpeciminTestExecutor.java
LoiNguyenCS: When I played around with these option, I noticed that "-w" will ignore white spaces but does not ignore blank lines, and as `diff` compares line by line, we will also need "-B" to ignore blank lines.

kelloggm: I've probably just misremembered - I should have checked myself. Sorry for the confusion.

kelloggm (approved):

PR #30, opened 2023-10-11
Add FieldDeclarationVisitor

LoiNguyenCS:


LoiNguyenCS:
Professor,

This is the PR to add the visitor that will tell us all declared fields in a Java file. Please take a look. Thank you.

src/main/java/org/checkerframework/specimin/FieldDeclarationsVisitor.java
```diff
@@ -0,0 +1,50 @@
...
+ * file. It's important to note that this visitor is intended to be used in conjunction with the
+ * UnsolvedSymbolVisitor, so both visitors will traverse the same Java file.
+ * Thus, @ClassGetSimpleName names for involved classes should be sufficient.
+ */
+public class FieldDeclarationsVisitor extends VoidVisitorAdapter<Void> {
+  /**
+   * A mapping of field names to the @ClassGetSimpleName name of the classes in which they are
+   * declared. Since inner classes can be involved, a map is used instead of a simple list.
+   */
+  Map<String, @ClassGetSimpleName String> fieldAndItsClass;
```
kelloggm, line 24: nit: I'd prefer `fieldNameToClassNameMap` or something else that makes it obvious just from the name that this object is a map. The current name is okay, but I think it is ambiguous whether it is intended to represent a single field or many fields just from the name.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -127,6 +127,13 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
    * type of fields in the parent class of the current class.
    */
   private final Set<String> syntheticTypes = new HashSet<>();
+
+  /**
+   * A mapping of field names to the @ClassGetSimpleName name of the classes in which they are
+   * declared.
```
kelloggm, line 133: This comment should be more specific: this is the fields of *the class currently being visited and its inner classes* - that is, the fields defined in the same file being visited and not in super classes.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -293,13 +300,22 @@ public Visitable visit(ExplicitConstructorInvocationStmt node, Void arg) {
 
   @Override
   public Visitable visit(NameExpr node, Void arg) {
+    //  This method explicitly handles NameExpr instances that represent fields of classes but are
```
kelloggm, line 303: nit: extra space at the start of this comment

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -293,13 +300,22 @@ public Visitable visit(ExplicitConstructorInvocationStmt node, Void arg) {
...
+    // not explicitly shown in the code. For example, if "number" is a field of a class, then
+    // "return number;" is an expression that this method will address. If the NameExpr instance is
+    // not a field of any class, or if it is a field of a class but is explicitly referenced, such
+    // as "Math.number," we handle it in other visit methods.
```
kelloggm, line 307: What if `number` is a local variable, as in:
```
int number = 5;
return number;
```

I don't see anything in this code that accounts for that case. Can you add a test case to make sure no synthetic super class field is generated for locals?

kelloggm (changes requested):
Overall looks very good, but I have one concern about the difference between local variable accesses and field accesses, since both are syntactically the same but the code doesn't seem to account for the possibility that a `NameExpr` might be a local variable.

LoiNguyenCS:
Professor @kelloggm,

> Overall looks very good, but I have one concern about the difference between local variable accesses and field accesses, since both are syntactically the same but the code doesn't seem to account for the possibility that a NameExpr might be a local variable.

Thank you for bringing this to my attention. I have forgotten that unsolved `NameExpr `could also include unsolved local variables. I think the problem will now get more complicated since we have to differentiate between unsolved fields of the current class, unsolved local variables, and unsolved fields of superclass.

I will see what I can do to solve this problem.

kelloggm:
> unsolved local variables

Fortunately, local variables are just that - local. So, it should be relatively easy to check what is in-scope.

LoiNguyenCS:
> Fortunately, local variables are just that - local. So, it should be relatively easy to check what is in-scope.
Thanks for telling me. I was about to upgrade `FieldDeclarationsVisitor` to make sure there are no unsolved fields left before passing the input file to `UnsolvedSymbolVisitor`. But your approach sounds much more doable 💯

LoiNguyenCS:
I was looking for methods to find the scope of a variable using JavaParser when I found this [issue](https://github.com/javaparser/javaparser/issues/2673). It seems to be that to know the scope of a variable, we need to trace back its declaration, which would require the variable to be solvable. 

I've also tried getting the `parentNode` of the `NameExpr`, but that doesn't work either. For example, no matter if `shot` is an unsolved local variable, field of the current class, or field from the superclass, its `parentNode` will be `return shot;`. If we keep tracing back, no matter what `shot` is, we will end up with the method declaration of `takeAShot`, which is not useful at all.
`
```
public int takeAShot() {
...
return shot;
}
```

kelloggm:
> I was looking for methods to find the scope of a variable using JavaParser when I found this https://github.com/javaparser/javaparser/issues/2673. It seems to be that to know the scope of a variable, we need to trace back its declaration, which would require the variable to be solvable.

There does seem to be some kind of `getDeclaration()` API. Using that, we could probably check whether something is a local variable declaration.

An alternative would just be to keep a local variable scope table ourselves, similar to what is mentioned in [this comment](https://github.com/javaparser/javaparser/issues/2257#issuecomment-638311359): effectively, we add an entry to this table every time we visit a local variable declaration, and we clear the table every time we exit a scope block. This might require us to be a bit careful to manage nested blocks, but I think it would be relatively easy for us to implement and we should do so.

LoiNguyenCS:
Professor,

I would like to request more time to address this issue, which appears to be a much more complex than originally anticipated. Several difficulties have arisen.

The first challenge is the `getDeclaration` function, which only works for solvable symbols.

The second complication arises from our use of JavaParser, which can not inform us when we exit a specific scope in the code.

The third and perhaps the trickiest part involves handling special local variables, such as those within nested if-else statements. For a local variable within a method, we can conveniently set the key in a Map (let's refer to it as Map A) as the variable's name, with the value being the method's name. However, the scenario becomes different when dealing with constructs like if-else, while, try-catch blocks, and the like. Determining what the key's value should be in such cases is not straightforward.

I've been experimenting with a potential solution where the value of a key in Map A represents the span of lines. For example, if method B has a local variable 'c,' the key in Map A would be 'c,' and its value would be denoted as 'x-y,' where 'x' is the starting line of the method B's declaration, and 'y' is the ending line. This approach, while promising, is quite tiring for me. It requires writing specific code for each special case, such as if-else statements, while loops, try-catch blocks, etc. Working with line numbers also adds complexity and demands careful handling.

When I got tired of dealing with the first solution, I have also tried a second one. In this second solution, I attempted to use the UnsolvedSymbolVisitor to resolve every local variable before dealing with any super variables. My hypothesis is based on the fact that the runtime of the UnsolvedSymbolVisitor is determined by the longest unsolved call statement in the input code. For example, if the longest unsolved call statement in the input code is 'a.b.c.d.e.x()', then UnsolvedSymbolVisitor will be called 6 times.. If we were to run the UnsolvedSymbolVisitor a number of times equivalent to the length of the longest unsolved call statement, then every local variable should, theoretically, be resolved. While this approach appears to work for my test cases, I cannot definitively prove its correctness.

LoiNguyenCS:
Professor, I have added a simple symbol table to Specimin. Please take a look. Thank you.

kelloggm (changes requested):
I think there are some scoping constructs that can contain local variables that this PR doesn't yet handle, especially `BlockStmt`s. You should go over the whole list of Java Parser Nodes and make sure that there are not others, too.

Further, I think the logic here is broken in the case of anonymous classes. Please add a test with an anonymous class inside a method to be sure.

LoiNguyenCS:
Professor, I have updated the program for it to work with anonymous classes. Please take a look. Regarding the warnings of ErrorProne, I will fix them in a separate PR. Thank you.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -291,15 +320,133 @@ public Visitable visit(ExplicitConstructorInvocationStmt node, Void arg) {
...
+    Statement thenStmt = (Statement) n.getThenStmt().accept(this, arg);
+    localVariables.pop();
+    if (condition == null || thenStmt == null) {
+      return null;
+    }
+    n.setCondition(condition);
+    @SuppressWarnings(
+        "nullness") // elseStmt could actually be null, according to the document of setElseStmt.
+    // This part is to avoid the false warnings of Checker Framework
+    @NonNull Statement nullableElseStme = elseStmt;
```
kelloggm, line 354: This warning suppression doesn't make much sense. I think you're saying that `setElseStmt`'s parameter can be null, in which case this code should be fine.

It's not true that `nullableElseStme`'s type in `@NonNull Statement`, so writing a warning suppression to that effect is an error.

The easiest solution to this warning is a stub file for `setElseStmt` that marks its parameter as `@Nullable`. Please add one, and then remove this warning suppression and local variable.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -291,15 +320,133 @@ public Visitable visit(ExplicitConstructorInvocationStmt node, Void arg) {
...
+    localVariables.pop();
+    return node;
+  }
+
+  @Override
+  public Visitable visit(VariableDeclarator decl, Void p) {
+    boolean isAField = false;
+    if (!decl.getParentNode().isEmpty()) {
+      if (decl.getParentNode().get() instanceof FieldDeclaration) {
+        isAField = true;
```
kelloggm, line 419: Rather than two `if` statements, just set `isAField`'s value to the conjunction of the two conditions.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -693,6 +888,23 @@ public void updateUnsolvedClassWithFields(
     }
   }
 
+  /**
+   * This method checks if a variable is local.
+   *
+   * @param fieldName the name of the variable
+   * @return true if that variable is local
+   */
+  public boolean isALocalVar(String fieldName) {
```
kelloggm, line 897: It is confusing that the name of the parameter to this method is `fieldName`, since the parameter might or might not be the name of a field. I suggest `variableName` or something like that, instead.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -291,15 +320,133 @@ public Visitable visit(ExplicitConstructorInvocationStmt node, Void arg) {
...
+      HashSet<String> currentListOfLocals = localVariables.pop();
+      currentListOfLocals.add(decl.getNameAsString());
+      localVariables.push(currentListOfLocals);
+    }
+    return super.visit(decl, p);
+  }
+
   @Override
   public Visitable visit(NameExpr node, Void arg) {
+    String fieldName = node.getNameAsString();
```
kelloggm, line 432: This is not always the name of a field, so we shouldn't call it `fieldName`. Just `name` would be fine.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -877,11 +1089,24 @@ public void updateMissingClass(UnsolvedClass missedClass) {
     while (iterator.hasNext()) {
       UnsolvedClass e = iterator.next();
       if (e.getClassName().equals(missedClass.getClassName())) {
+
+        // add new methods
         for (UnsolvedMethod method : missedClass.getMethods()) {
-          if (!method.getReturnType().equals("")) {
+          boolean alreadyHad = false;
```
kelloggm, line 1095: A brief comment explaining why this boolean is necessary would be helpful. I think the reason is that if the same method is used multiple times in the target(s), it will appear more than once in this loop: once for each call site.

kelloggm (changes requested):
Overall, this PR is in pretty good shape and my comments are relatively minor.

kelloggm:
> Regarding the warnings of ErrorProne, I will fix them in a separate PR.

When you do, please also reconfigure CI so that it fails any time that ErrorProne warns.

LoiNguyenCS:
Professor,
I have updated the program based on your suggestions.

kelloggm (approved):

PR #34, opened 2023-11-01
Solve error prone warnings

LoiNguyenCS:


kelloggm:
> While it is possible to mark all warnings, including those not originating from Error Prone, as errors, this approach may be challenging to maintain

I think we should mark all warnings as errors (i.e., pass the `-Werror` flag to javac), even if there will be a bit of maintenance pain. The advantage of blocking builds on ErrorProne's warnings is too good to pass up.

We should also update our version of the ErrorProne gradle plugin (it's currently 2.0.2, [latest](https://plugins.gradle.org/plugin/net.ltgt.errorprone) is 3.1.0). That plugin should automatically prevent ErrorProne itself from causing warnings from javac. If we get other warnings from javac, I'd like to know about them.

LoiNguyenCS:
Professor, I have update the `build.gradle` to use the newest version of ErrorProne and also mark every warning as an erros.

kelloggm (approved):
Thanks for these fixes! I think the change to `-Werror` will be very helpful going forward to keep us honest about listening to ErrorProne.

PR #35, opened 2023-11-01
Fix bugs with static nested classes

LoiNguyenCS:


LoiNguyenCS:
Professor,

When working with the issues from Checker Framework, I've realized that `Specimin` doesn't work when there is a static nested class. I've made a small change, and I think the program is working now.

kelloggm (approved):
LGTM, thanks. Nice catch.

Issue #37, opened 2023-11-07
Java Parses Exception when attempting to pass (int[], int) as method params


jonathan-m-phillips:
## Issue:
Below are the commands provided, the class/method targeted, and the exception thrown. I am not positive if I am running Specimin incorrectly here, or if I am running into a bug.

### Command:
% ./gradlew run --args='--outputDirectory "/my/output/dir" --root "my/project/root/java" --t
argetFile "ArraySorter.java" --targetMethod "ArraySorter#getSortedArrayLargerThanIndex(int[], int)"'

### Class:
```
public class ArraySorter {

    public static int[] getSortedArrayLargerThanIndex(int[] array, int index) {
        if (index < 0 || index >= array.length) {
            return new int[0];
        }

        int count = 0;
        for (int i = 0; i < array.length; i++) {
            if (array[i] > array[index]) {
                count++;
            }
        }

        int[] resultArray = new int[count];

        int j = 0;
        for (int i = 0; i < array.length; i++) {
            if (array[i] > array[index]) {
                resultArray[j++] = array[i];
            }
        }

        Arrays.sort(resultArray);

        return resultArray;
    }
}
```

### Exception:
> Task :run FAILED
Exception in thread "main" java.lang.UnsupportedOperationException: ResolvedArrayType{PrimitiveTypeUsage{name='int'}} is not a Reference Type
        at com.github.javaparser.resolution.types.ResolvedType.asReferenceType(ResolvedType.java:142)
        at org.checkerframework.specimin.TargetMethodFinderVisitor.visit(TargetMethodFinderVisitor.java:199)
        at org.checkerframework.specimin.TargetMethodFinderVisitor.visit(TargetMethodFinderVisitor.java:32)
        at com.github.javaparser.ast.body.MethodDeclaration.accept(MethodDeclaration.java:98)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
        at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
        at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:243)
[...]

PR #39, opened 2023-11-11
Fix bugs with array type

LoiNguyenCS:


LoiNguyenCS:
Professor,

This is the PR to fix the issue #37 that Jonathan has. I'm not sure if this is a minor bug of JavaParser or not. This is the code snippet where we get the bug:

```
  Type returnType = method.getType();
   if (returnType instanceof ReferenceType) {
        usedClass.add(returnType.resolve().asReferenceType().getQualifiedName());
   }
```

The problem arises when `returnType` is an array type. In such cases, JavaParser incorrectly considers it as a reference type, leading to a bug when trying to resolve and cast it.

The solution is to resolve `returnType` first before checking if it is an instance of reference type.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -195,7 +195,7 @@ public Visitable visit(MethodDeclaration method, Void p) {
       targetMethods.add(method.resolve().getQualifiedSignature());
       unfoundMethods.remove(methodName);
       Type returnType = method.getType();
-      if (returnType instanceof ReferenceType) {
+      if (returnType.resolve() instanceof ResolvedReferenceType) {
```
kelloggm, line 198: Please add a comment here explaining that this is working around a JavaParser bug. (The comment should explain exactly what you explained in the comment you opened on this PR.)

kelloggm (changes requested):

kelloggm:
> I'm not sure if this is a minor bug of JavaParser or not

I'd call this a bug in JavaParser. Unfortunately, we have to work around it here.

kelloggm (approved):

Issue #41, opened 2023-11-13
Issues Compiling with Checker Framework on Minimized ASHE Classes


jonathan-m-phillips:
I am encountering compilation errors with the Checker Framework when minimizing classes from the [ASHE project](https://github.com/jonathan-m-phillips/ASHE_Automated-Software-Hardening-for-Entrypoints), specifically the ASHE#run(String, String, String) method. The minimization process seems to disrupt the dependencies within the project, particularly when the target class relies on other internal classes.

### Arguments
**NOTE:** The arguments are different than the _main_ branch on GitHub. The directory has changed on my local environment, which is why the typical would not use edu.njit.jerse.**_ashe_**.ASHE#run(String, String, String), but rather edu.njit.jerse.ASHE#run(String, String, String).
I used the following arguments for the Specimin minimization process:

`--args=--outputDirectory "my/temp/dir" --root "my/root/to/ASHE_Automated-Software-Hardening-for-Entrypoints/src/main/java" --targetFile "edu/njit/jerse/ashe/ASHE.java" --targetMethod "edu.njit.jerse.ashe.ASHE#run(String, String, String)"`

### Relevant Code Snippet from ASHE.java
```
public class ASHE {

    private static final Logger LOGGER = LogManager.getLogger(ASHE.class);

    public void run(String root, String targetFile, String targetMethod) throws IOException, ExecutionException, InterruptedException, TimeoutException {
        JavaCodeCorrector corrector = new JavaCodeCorrector();
        String speciminTempDir = corrector.minimizeTargetFile(root, targetFile, targetMethod);
        final String sourceFilePath = speciminTempDir + "/" + targetFile;
        boolean errorsReplacedInTargetFile = corrector.fixTargetFileErrorsWithGPT(sourceFilePath, targetMethod);
        if (!errorsReplacedInTargetFile) {
            if (corrector.checkedFileError(sourceFilePath).isEmpty()) {
                return;
            } else {
                throw new RuntimeException("Errors were not replaced with GPT response.");
            }
        }
        String methodName = JavaCodeParser.extractMethodName(targetMethod);
        String originalFilePath = root + "/" + targetFile;
        boolean isOriginalMethodReplaced = MethodReplacementService.replaceOriginalTargetMethod(sourceFilePath, originalFilePath, methodName);
        if (!isOriginalMethodReplaced) {
            throw new RuntimeException("Original method was not replaced.");
        }
    }
}
```

### Exception Observed
[...]

PR #42, opened 2023-11-13
Add static method

LoiNguyenCS:


LoiNguyenCS:
Professor,

When I was writing the paper for Specimin, I realized that I forgot to handle static method calls in the input program, which is quite embarrassing. This pull request addresses that by adding the necessary code for unsolved static method calls. Additionally, I noticed the existence of the `tempDirJDK` folder, which I had created a while ago for bug testing purposes. I believe it was accidentally added at some point in the past, so this PR includes the removal of that folder.

In this updated version of Specimin, wildcard imports are no longer allowed. The reason is that we need to distinguish between class and method names in unsolved method calls, such as `A.c()`. The only reliable way to determine whether `A` refers to a class or a method is to look at the import statements. 

I think the logic of this solution might not apply if there is a method or class instance with the same name as a class among the import statements. But I think the odd is low, since usually people don't name methods or class instances with a capitalized name like they do for classes.

src/test/resources/unsolvedstaticmethod/expected/com/example/Class.java
```diff
@@ -0,0 +1,8 @@
+package com.example;
+
+public class Class {
+
+    public ProcessReturnType process() {
```
kelloggm, line 5: Shouldn't this be `static`?

kelloggm (changes requested):
Nice catch! Please also add a test for a more complex static call (e.g., one where the programmer uses the fully-qualified name of the class).

LoiNguyenCS:
Professor,

I've made some changes to the code.

My initial implementation was incorrect. The problem was somewhat concealed because I named my test class `Class`, coincidentally matching some built-in Java classes. Thus, the program ran without triggering any error.

I've also introduced a more complex method call in the test file.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1216,7 +1219,7 @@ public static boolean isAClassPath(String potentialClassPath) {
    * @param fullyName the fully-qualified name of the class
    * @return the corresponding instance of UnsolvedClass
    */
-  public static UnsolvedClass getSimpleSyntheticClassFromFullyQualifiedName(
+  public UnsolvedClass getSimpleSyntheticClassFromFullyQualifiedName(
```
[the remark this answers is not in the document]
LoiNguyenCS: Ah, I was trying another approach earlier. I will re-add `static`

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1254,12 +1294,17 @@ public static boolean unsolvedAndNotSimple(MethodCallExpr method) {
...
+   * @param methodAgruments the list of agruments for this method call
    */
-  public void updateClassSetWithNotSimpleMethodCall(MethodCallExpr method) {
-    String methodCall = method.toString();
+  public void updateClassSetWithNotSimpleMethodCall(
+      String methodCall, List<String> methodAgruments) {
+    // As this code involves complex string operations, we'll use a method call as an example,
+    // following its progression through the code.
+    // Suppose this is our method call: com.example.MyClass.process()
+    // At this point, our method call become: com.example.MyClass.process
```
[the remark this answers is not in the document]
LoiNguyenCS: Thank you for pointing this out. I've changed the replace call into a substring call.

LoiNguyenCS:
Professor,

I have updated the codes and the test file.

LoiNguyenCS:
I don't know why the CI is not running. I think this is a [bug](https://github.com/orgs/community/discussions/31587) of GitHub. But the solutions discussed in that bug report don't seem to be related to our case.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1254,43 +1294,58 @@ public static boolean unsolvedAndNotSimple(MethodCallExpr method) {
 
   /**
    * For a method call that is not simple, this method will take that method as input and create
-   * corresponding synthetic class
+   * corresponding synthetic class.
    *
-   * @param method the method call to be taken as input
+   * @param methodCall the method call to be taken as input
+   * @param methodAgruments the list of agruments for this method call
```
kelloggm, line 1300: typo: "arguments"

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1254,43 +1294,58 @@ public static boolean unsolvedAndNotSimple(MethodCallExpr method) {
 
   /**
    * For a method call that is not simple, this method will take that method as input and create
-   * corresponding synthetic class
+   * corresponding synthetic class.
    *
-   * @param method the method call to be taken as input
+   * @param methodCall the method call to be taken as input
```
kelloggm, line 1299: This string needs to be a particular format. I think it needs to be a static invocation: in particular, it must not have a receiver, since you use a substring call to find the first `(` and expect the result to be the beginning of the arguments list. If there was a receiver like the following, this would fail:
"foo.getBar().methodCall(...)" - if we passed this whole string, the substring call below will only find "foo.getBar", not "foo.getBar().methodCall", as we intend.

So, please give the exact specification that you expect here in this Javadoc.

kelloggm (changes requested):
I'm still concerned about the string manipulation.

Also, I'm not sure that this code will handle [static imports](https://docs.oracle.com/javase/8/docs/technotes/guides/language/static-import.html) correctly. Can you also add a test that statically imports a method and make sure that the code handles it properly?

LoiNguyenCS:
Professor,

> I'm still concerned about the string manipulation.

I have renamed the related methods and added clarification in the Javadoc to make sure that the expected input will be passed to this string manipulation call.

> Also, I'm not sure that this code will handle [static imports](https://docs.oracle.com/javase/8/docs/technotes/guides/language/static-import.html) correctly. Can you also add a test that statically imports a method and make sure that the code handles it properly?

Unfortunately, I wasn't aware of this Java feature until today. Updating Specimin for static imports might involve extensive modifications, such as changing how it reads import statements and updates `classAndPackageMap`. I suggest addressing this in a separate PR, ideally after resolving issue #41.
`

kelloggm:
> Updating Specimin for static imports might involve extensive modifications, such as changing how it reads import statements and updates classAndPackageMap. I suggest addressing this in a separate PR, ideally after resolving issue https://github.com/kelloggm/specimin/issues/41.

They are rare, so that's fine. Please open an issue for this so we don't forget.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1293,14 +1293,15 @@ public boolean unsolvedAndNotSimple(MethodCallExpr method) {
...
   /**
-   * For a method call that is not simple, this method will take that method as input and create
-   * corresponding synthetic class.
+   * Creates a synthetic class corresponding to a static method called by a qualified class name.
+   * Ensure to check with {@link #isAnUnsolvedStaticMethodCalledByAQualifiedClassName} before
+   * calling this method.
    *
-   * @param methodCall the method call to be taken as input
-   * @param methodAgruments the list of agruments for this method call
+   * @param methodCall the method call to be used as input
```
kelloggm, line 1300: I would still prefer that the `@param` Javadoc here explains exactly the requirements on `methodCall` in string terms.

kelloggm: That is, "must contain one or more dot separated identifiers, followed by a single pair of parentheses containing arguments" or something similar

kelloggm (approved):
When the requested changes are implemented, you can merge without requiring another review from me.

Issue #38, opened 2023-11-10
Exception while running SPECIMIN on cassandra project. 


tahiat:
Repo:[]( https://github.com/jacek-lewandowski/cassandra/commit/7b3c4ce8e5d6da41791492903ec367b8857cd57d)

ExecutionCommand: `./gradlew run --args='--outputDirectory "issue6077" --root "/Users/tahiatul/Documents/SPECIMEN/ISSUES/issue_6077/cassandra/src/java" --targetFile "org/apache/cassandra/index/sasi/conf/IndexMode.java" --targetMethod "org.apache.cassandra.index.sasi.conf.IndexMode#getMode(ColumnMetadata, Map<String, String>)"'`

Original Issue in CF: https://github.com/typetools/checker-framework/issues/6077

Stack trace:

```
Exception in thread "main" java.lang.IllegalStateException: InstantiationException | IllegalAccessException is not ClassOrInterfaceType, it is UnionType
        at com.github.javaparser.ast.type.Type.asClassOrInterfaceType(Type.java:176)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:568)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:68)
        at com.github.javaparser.ast.body.Parameter.accept(Parameter.java:113)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:208)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:397)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:68)
        at com.github.javaparser.ast.stmt.CatchClause.accept(CatchClause.java:77)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
        at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
        at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:947)
[...]

PR #45, opened 2023-11-15
Crash fix for UnionType parameter in Issue #38

tahiat:
**Issue statement:** Casting `UnionType` parameter as `ClassOrInterfaceType` to resolve the parameter causes a crash in `UnsolvedSymbolVisitor.java`. 

**Current Fix** Since both `InstantiationException` and `IllegalAccessException` are built in Java language, they should have been resolved successfully instead of throwing an exception. To resolve this I get each element of UnionType and resolve them separately. 

**Threat**
since InstantiationException , IllegalAccessException is built in java, resolve is successful. Custom exception types need to be handled.


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -544,7 +545,14 @@ public Visitable visit(MethodCallExpr method, Void p) {
   @Override
   public Visitable visit(Parameter parameter, Void p) {
     try {
-      parameter.resolve().describeType();
+      if (parameter.getType() instanceof UnionType) {
+        UnionType unionType = parameter.getType().asUnionType();
+        for (var param : unionType.getElements()) {
+          param.resolve().describe();
```
LoiNguyenCS, line 551: As you can see in the original codes, the purpose of this try block is to check if a parameter is unsolved. If it is, the parameter will be sent to the catch block where Specimin will create a synthetic class file for it. So I think we also need to do the samething with `param`. If it is unsolved (that is to say, an exception is thrown when we try to resolve it), `param` should be sent to the catch block so Specimin can create a synthetic class file for it.

LoiNguyenCS (changes requested):
I think overall this code is good. But as you say, this current version will only work if both the two types in the union type is solvable, which is not guaranteed in real life. So we should also consider to create synthetic class file if one or both of the types are unsolved.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
tahiat: I have pushed a commit addressing this comment

tahiat:
```
package com.example;

class Simple {
    // Target method.
    void bar() {
        try{

        }
        catch (ClassNotFoundException e){

        }
    }
}
```
As expected, resolving `ClassNotFoundException` raises exception. However, the `classAndPackageMap` does not contain  `ClassNotFoundException` as the key.

kelloggm:
> ```
> package com.example;
> 
> class Simple {
>     // Target method.
>     void bar() {
>         try{
> 
>         }
>         catch (ClassNotFoundException e){
> 
>         }
>     }
> }
> ```
> 
> As expected, resolving `ClassNotFoundException` raises exception. However, the `classAndPackageMap` does not contain `ClassNotFoundException` as the key.
> 
> Adding the `import java.lang.ClassNotFoundException` resolves this issue. But it is unlikely to to have this import statement in the class file.
> 
> I believe this issue will occur for every java.lang types that not imported.

I don't understand the problem that you are raising here. All java.lang types are automatically imported by javac at the start of each file.

Based on @LoiNguyenCS's previous comment, you should write test cases that 1) uses an exception class that's defined as part of the input, and 2) uses an exception class that is unsolvable (i.e., is from a library that would presumably be on the classpath when actually compiling the code). Please re-assign me as a reviewer once you've done this.

src/test/resources/unsolvedexception/expected/com/example/Simple.java
```diff
@@ -0,0 +1,11 @@
+package com.example;
+
+public class Simple {
```
kelloggm, line 3: This expected output is not compilable, because `CustomException` is never defined. A synthetic class should be created for it.

Maybe this is what you were referring to when you mention #47 when we were talking on the phone a few minutes ago?

I would prefer not to merge this PR until the test cases are correct: it is an anti-pattern in software engineering to commit known-incorrect test oracles. If we need to fix #47 first (so that we can create a synthetic class for `CustomException` here), we should delay merging this PR until then. Please update this test with the correct expected output.

src/test/java/org/checkerframework/specimin/Issue38Test.java
```diff
@@ -0,0 +1,15 @@
+package org.checkerframework.specimin;
```
kelloggm, line 1: I think there should be one more test runner (for the `unsolvedexception` test).

kelloggm (changes requested):

src/test/resources/unsolvedexception/expected/com/example/Simple.java
tahiat: Yes professor. I will then solve the dependent issue first then merge this one.

src/test/java/org/checkerframework/specimin/Issue38Test.java
tahiat: I did not add this since it will fail because of the issue I discussed in the meeting. I will fix the dependent issue and add the testrunner.

kelloggm:
@tahiat #46 has been merged into this PR's branch, but the tests still fail. Can you investigate (and fix) the failures please?

tahiat:
Looking into it

tahiat:
Even if the `CustomException` class is used as part of the input, the `para.resolve()` throws exception in `visit(Parameter parameter, Void p) ` method in `UnsolvedSymbolVisitor.java` . To mitigate this I added a check if the parameter is from Catch clause. If that check is true I used `para.getType().resolve()` to see if symbol solver can resolve this. 


ALSO brought some change from Loi's unmerged PR.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -219,19 +220,44 @@ public Visitable visit(MethodDeclaration method, Void p) {
...
+          try {
+            paramType = param.resolve();
+            paraTypeFullName =
+                paramType.asReferenceType().getTypeDeclaration().get().getQualifiedName();
+            usedClass.add(paraTypeFullName);
+          } catch (
+              Exception
+                  e) { // This visitor should not have an unsolvable parameter type. Throwing if
+            // exception occurs.
+            throw new Error(e.getMessage());
```
kelloggm, line 237: Why catch and rethrow? If you're not going to add any information to the exception, you should just let it pass up the call stack. Adding a wrapper like this is not helpful.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -219,19 +220,44 @@ public Visitable visit(MethodDeclaration method, Void p) {
...
-          String typeParameterValueName = typeParameterValue.describe();
-          if (typeParameterValueName.contains("<")) {
-            // removing the "<...>" part if there is any.
-            typeParameterValueName =
-                typeParameterValueName.substring(0, typeParameterValueName.indexOf("<"));
+      Type type = para.getType();
+      String paraTypeFullName = "";
+      ResolvedType paramType;
+      if (type.isUnionType()) {
+        for (var param : para.getType().asUnionType().getElements()) {
```
kelloggm, line 227: Personally, I'm not a fan of `var`, since it is sometimes important to know the type of a variable. I'd prefer that we avoid it, even in new code.

kelloggm: It's not any easier to write (it still takes keystrokes!), but it makes the code harder to read by obscuring information. So, it's a poor tradeoff for code that will need to be maintained (= all code we write for research!).

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -219,19 +220,44 @@ public Visitable visit(MethodDeclaration method, Void p) {
...
+            usedClass.add(paraTypeFullName);
+          } catch (
+              Exception
+                  e) { // This visitor should not have an unsolvable parameter type. Throwing if
+            // exception occurs.
+            throw new Error(e.getMessage());
+          }
+        }
+      } else {
+        if (para.getParentNode().get() instanceof CatchClause) {
```
kelloggm, line 241: This only works if the immediate parent is a catch clause. A catch clause can contain arbitrary additional nesting blocks - for example, it might contain one or more levels of `if` statements or loops. Please add a test case where a catch block has multiple levels of parents before references one of the exception parameters, and then fix this test so that it handles arbitrary nesting.

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
```diff
@@ -50,6 +53,21 @@ public UnsolvedClass(@ClassGetSimpleName String className, String packageName) {
...
   }
 
+  /**
+   * Create an instance of UnsolvedClass
+   *
+   * @param className the name of the class
+   * @param packageName the name of the package
+   */
+  public UnsolvedClass(
+      @ClassGetSimpleName String className, String packageName, boolean exceptionType) {
```
kelloggm, line 63: Rather than updating all call sites of this constructor, it would be better to add a second constructor that has the same signature as the constructor before your PR, and which calls this constructor with `exceptionType` set to false.

src/test/java/org/checkerframework/specimin/Issue38Test.java
kelloggm: I don't think this is done @tahiat

kelloggm (changes requested):
> ALSO brought some change from Loi's unmerged PR.

When making a comment like this, you should link the specific PR you're referring to rather than being vague. In fact, Loi has multiple open, unmerged PRs, so this is ambiguous (which made this harder to review).

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
tahiat: Current code has two constructors. 
```
public UnsolvedClass( @ClassGetSimpleName String className, String packageName)
public UnsolvedClass(@ClassGetSimpleName String className, String packageName, boolean exceptionType)
```
The second one is called where the `exceptionType` is `true`. In all other cases the first one is called. The default value of the `isExceptionType` is set to `false`;

I think you suggested this, am I right?

kelloggm: Partially. The problem in the current implementation is actually the other (unchanged) constructor: it should call this constructor explicitly rather than re-implementing the assignment of fields. In other words, my suggestion is already partially (but not completely) implemented.

tahiat: public UnsolvedClass(
      @ClassGetSimpleName String className, String packageName, boolean exceptionType) {
    this(className, packageName);
    this.isExceptionType = exceptionType;
  }

kelloggm: The dependency should go the other way: the other constructor's implementation should be:
```
public UnsolvedClass(@ClassGetSimpleName String className, String packageName) {
  this(className, packageName, false);
}
```

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
tahiat: Professor, I have used the actual type instead of `var`

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
tahiat: Added a separate test case with additional nesting blocks. Since only the parameter of the catch is targeted, I thinks it's it safe to use this way. 

```
try {
} catch (ExceptionType e) { // --> para.resolve() fails in this case
}
```

when ExceptionType is used in a method parameter, para.resolve() works normally. I have created an issue on java parser repo for this case.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
tahiat: updated accordingly

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
tahiat: Thanks Professor. updated accordingly.

tahiat:
Few points we have discussed in today's meeting. 

(1) Add a test case with arbitrary number of  nested blocks inside catch clause and check if that test case completes successfully. --> Added a test case. It's safe to use the  `para.getParentNode().get() instanceof CatchClause` check for resolving the parameter. 

(2) When creating a unsolved class for an exception, current implementation check if the reference type is from catch block. Professor, you suggested if I can check if any of it's parent is Exception type. Yes I can check that. 

(3) I observed that if `parameter.getType().resolve()` is used, it successfully resolve the exception type in catch clause and all test cases passed except the 'typevar-collision' testcase. Since `param.getType().resolve()` is specifically used to resolve the type of the parameter, It throws exception because there symbol solver can't distinguish between com.T and typed parameter.  However in this case, `parameter.resolve()` successfully resolve that it's a type variable. Below is the input for typevar-collision testcase.
```
package org;
// This import isn't used.
import com.T;
public class Foo<T> {
    T useT(T t) {
        return t;
    }
}
```

tahiat:
It will be good if we merge this PR after PR #57

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -220,19 +222,37 @@ public Visitable visit(MethodDeclaration method, Void p) {
...
+      ResolvedType paramType;
+      if (type.isUnionType()) {
+        for (ReferenceType param : para.getType().asUnionType().getElements()) {
+          paramType = param.resolve();
+          paraTypeFullName =
+              paramType.asReferenceType().getTypeDeclaration().get().getQualifiedName();
+          usedClass.add(paraTypeFullName);
+        }
+      } else {
+        if (para.getParentNode().isPresent() && para.getParentNode().get() instanceof CatchClause) {
```
kelloggm, line 236: This `if` statement must be explained with a comment. Ideally, that comment will link to the JavaParser bug/discussion that you and I talked about in our meeting yesterday @tahiat - you said you'd asked the JavaParser people about this, and they told you it should work (but it doesn't). Substantiate and explain that claim here where you work around it in a comment.

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
```diff
@@ -172,7 +188,18 @@ public void updateFieldByType(String currentType, String correctType) {
...
+      sb.append("public class ")
+          .append(className)
+          .append(getTypeVariablesAsString())
+          .append(" extends Exception")
+          .append(" {\n");
+    } else {
+      sb.append("public class ")
+          .append(className)
+          .append(getTypeVariablesAsString())
+          .append(" {\n");
```
kelloggm, line 201: This code is too-duplicated: there should only be one call to `append` for each string. If necessary, you can split this chain of calls up to insert the "extends Exception" in the right place, but you should not duplicate the whole chain.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
tahiat: updated

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
tahiat: added comments here. 
parameter resolution works on method and construction but not on catch clause.

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
kelloggm: rather than leaving a comment like "updated", you can resolve the comment. This has the same semantics (it indicates to the reviewer that you've done what was asked), but it does not generate an email notification and is easier for you. Reviewers generally do not need an email notification about each of their comments, so it is better (and more polite) to save yourself the effort.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -330,4 +328,16 @@ public Visitable visit(NameExpr expr, Void p) {
     }
     return super.visit(expr, p);
   }
+
+  /**
+   * @param type unionType parameter
```
kelloggm, line 333: Please provide a Javadoc that summarizes the purpose of the method for every method that you write (the build system should also warn you that this is missing)

kelloggm (changes requested):
Also, please merge from `main` so that you incorporate the changes from #57 before you request another review.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
tahiat: build system did not show any warning for this method. I guess method description is not forced. I added method description. Added documentation for a method that was not part of the PR

tahiat:
Professor, branch is updated with main

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
kelloggm: > build system did not show any warning for this method. I guess method description is not forced

That's probably a misconfiguration in the build system. I'll look into it later today.

> Added documentation for a method that was not part of the PR

That shouldn't be necessary (now or ever). But, you should do so for new code.

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
```diff
@@ -121,7 +121,11 @@ public void addFields(String variableExpression) {
     this.classFields.add(variableExpression);
   }
 
-  /** This method sets the number of type variables for the current class */
+  /**
+   * This method sets the number of type variables for the current class
+   *
+   * @param numberOfTypeVariables number of type variable in this class.
+   */
```
kelloggm, line 128: This change is harmless, but it wasn't necessary. I'm going to leave it in, but if we were in less of a hurry-up mode right now I'd insist you PR it separately.

kelloggm (approved):

Issue #43, opened 2023-11-13
Parameters aren't considered local variables, leading to a crash


kelloggm:
Consider the following program (based on #40, but in a form suitable for a specimin test case):
```
import java.util.List;
import java.util.ArrayList;
import com.github.javaparser.ast.NodeList;
import com.github.javaparser.ast.ImportDeclaration;

public class Issue40 {
    List<String> importStatement;

    void setclassAndPackageMap() {

    }

    public void setImportStatement(NodeList<ImportDeclaration> listOfImports) {
        List<String> currentImportList = new ArrayList<>();
        for (ImportDeclaration importStatement : listOfImports) {
            String importAsString = importStatement.getNameAsString();
            currentImportList.add(importAsString);
        }
        this.importStatement = currentImportList;
        this.setclassAndPackageMap();
    }
}
```

When I add this to Specimin's test suite (see the `issue40` branch), it crashes with the following stacktrace:
```
java.lang.RuntimeException: Unfound parent for this class: Issue40
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.getParentClass(UnsolvedSymbolVisitor.java:1342)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.updateSyntheticClassForSuperCall(UnsolvedSymbolVisitor.java:846)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:441)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:68)
	at com.github.javaparser.ast.expr.NameExpr.accept(NameExpr.java:79)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:519)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.stmt.ForEachStmt.accept(ForEachStmt.java:86)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
	at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
[...]

PR #46, opened 2023-11-16
add codes for tricky parameters

LoiNguyenCS:


LoiNguyenCS:
Professor,

This is the PR to deal with tricky parameters, which hopefully will fix issue #43.

This one feels a bit like going down a rabbit hole. I know it's usually better to split changes into different PRs, but doing that might take a lot of time for me. To make up for it, I've added more comments to explain things better.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -165,7 +166,9 @@ public static void performMinimization(
     // correct the types of all related files before adding them to parsedTargetFiles
     JavaTypeCorrect typeCorrecter = new JavaTypeCorrect(root, relatedClass);
     typeCorrecter.correctTypesForAllFiles();
-    addMissingClass.updateTypes(typeCorrecter.getTypeToChange());
+    Map<@ClassGetSimpleName String, @ClassGetSimpleName String> typeToChange =
```
kelloggm, line 169: Should the name of this variable be `typesToChange`?

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -187,7 +190,20 @@ public static void performMinimization(
       if (isEmptyCompilationUnit(target.getValue())) {
         // target key will have this form: "path/of/package/ClassName.java"
         String classFullyQualfiedName = target.getKey().replace("/", ".");
-        classFullyQualfiedName = classFullyQualfiedName.replace(".java", "");
+        // this condition is actually always true
```
kelloggm, line 193: Then why is it necessary to check it? Maybe we should have an else block that throws an exception if it somehow is false?

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -207,6 +208,21 @@ public Visitable visit(MethodDeclaration method, Void p) {
     return result;
   }
 
+  @Override
+  public Visitable visit(Parameter para, Void p) {
+    if (insideTargetMethod) {
+      ResolvedType paraType = para.resolve().getType();
+      if (paraType.isReferenceType()) {
+        String paraName = paraType.asReferenceType().getTypeDeclaration().get().getQualifiedName();
```
kelloggm, line 216: The name of this variable is misleading. `paraName` suggests that it is the name of the parameter. (For example, for this method's parameter `Parameter para`, I'd expect `paraName` to be the string `para`.) Actually, I think this is the name of the the type of the parameter. So maybe something like `paraTypeName` would be clearer?

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
```diff
@@ -27,6 +27,9 @@ public class UnsolvedClass {
    */
   private final String packageName;
 
+  /** This field checks if this class has a placeholder */
+  private boolean hasAPlaceHolder = false;
```
kelloggm, line 31: What is a "placeholder"? That's not a standard term, so you should define it.

I see later in the class that you add a type variable `<T>` to the class declaration if this is true. Does that mean that "placeholder" here is a synonym for type variable?

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
```diff
@@ -149,7 +165,11 @@ public void updateFieldByType(String currentType, String correctType) {
   public String toString() {
     StringBuilder sb = new StringBuilder();
     sb.append("package ").append(packageName).append(";\n");
-    sb.append("public class ").append(className).append(" {\n");
+    if (hasAPlaceHolder) {
```
kelloggm, line 168: This logic doesn't seem to handle classes with more than one type variable, such as `java.util.Map`. Can you add a test case where an unsolved class has two different type variables?

kelloggm: My other concern about this is that this code will never add an `extends` clause to the type variable, even if one is warranted. Maybe that's not possible if the class is unsolved, though.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -263,6 +264,21 @@ public Visitable visit(PackageDeclaration node, Void arg) {
...
 
+  @Override
+  public Node visit(ImportDeclaration decl, Void arg) {
+    @SuppressWarnings(
+        "signature") // since this is the content of an import statement, it will be a qualified
+    // class path
+    @FullyQualifiedName String declAsString = decl.getNameAsString();
+    // if there is no jar paths included, every imported class is unsolved, apart from those
+    // belonging to the Java language.
+    if (!classesFromJar.isEmpty() || declAsString.startsWith("java")) {
```
kelloggm, line 275: Users are also allowed to add classes to the `java` namespace. For example, consider a program with this import:
```
import javaproject.MyClass;
```

This is totally valid (and I've seen students do similar things plenty of times...), but this check will misclassify such an import as solved.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -544,8 +590,28 @@ public Visitable visit(MethodCallExpr method, Void p) {
...
+      try {
+        resolvedParameter.describeType();
+        return super.visit(parameter, p);
+      } catch (IllegalArgumentException e) {
+        // following the above explanation, typeName will be in this form path.to.A<path.to.B>
+        String typeName = parameter.getType().asString();
+        @SuppressWarnings(
+            "signature") // since this type is included among the import statements, we can expect
+        // that when it is used, it will be in its simple form.
+        @ClassGetSimpleName String typeToAddPlaceHolder = typeName.substring(0, typeName.indexOf("<"));
```
kelloggm, line 607: This code should check for commas in the part between the `<`s.

Also, note that this could be recursive: for example, consider the type `List<Map<String, Object>>`. We need to be able to handle arbitrarily-deep type parameters here.

src/test/resources/unsolvedparameter/input/com/example/Simple.java
```diff
@@ -0,0 +1,13 @@
+package com.example;
+
+import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.ImportDeclaration;
+
+public class Simple {
+
+    public void lengthImportStatement(NodeList<ImportDeclaration> listOfImports) {
+        for (ImportDeclaration importDecl: listOfImports) {
```
kelloggm, line 9: Please create a version of this test case without the for loop, to make sure that `ImportDeclaration` is still included even if its only use is in a type variable instantiation.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
LoiNguyenCS: Thank you for telling me. I've been searching the internet and people call that `<T>` part placeholder for generic types, so I thought "placeholder" was a standard term.

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
LoiNguyenCS: > My other concern about this is that this code will never add an extends clause to the type variable, even if one is warranted. Maybe that's not possible if the class is unsolved, though.

I've also thought about it.  I don't think we can figure out if an unsolved class should have an `extends` or `implements` clause.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: As I understand, since this is generic type, we don’t really need to care about how deep it can go.

For example, if this our unsolved type: `ImportDeclaration<List<Map<String, String>> `, then a synthetic declaration such as `ImportDeclaration<T>` should be enough to get the program compuled.

kelloggm:
That's only true if the types with which the generic is instantiated are
all solvable. Consider e.g. `NodeList<Node<Import>>` as an example. In this
case, we may need to create synthetic classes for all three of NodeList,
Node, and Import.

I agree that the synthetic class for NodeList doesn't care about how many
levels there are, though.

Martin

LoiNguyenCS:
Professor, 

I have found a cool visit method of JavaParser that can help us deal with those class types and type variables recursively. That is the visit method for `ClassOrInterfaceType`. There is a weird bug accompanying this method, so I have to add a check to workaround.

I have also deleted the visit method for import statements. I think it is unsafe to assume that if there is no jar paths, every class imported must be unsolved.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -207,6 +208,28 @@ public Visitable visit(MethodDeclaration method, Void p) {
...
 
+  @Override
+  public Visitable visit(Parameter para, Void p) {
+    if (insideTargetMethod) {
+      ResolvedType paraType = para.resolve().getType();
+      if (paraType.isReferenceType()) {
+        String paraTypeFullName =
+            paraType.asReferenceType().getTypeDeclaration().get().getQualifiedName();
+        usedClass.add(paraTypeFullName);
+        for (ResolvedType typeVariable : paraType.asReferenceType().typeParametersValues()) {
```
kelloggm, line 219: Following the API, I'd name the loop variable `typeParameterValue` rather than `typeVariable`. (I also suggest changing the associated variables, like `typeVariableFullName` -> `typeParameterValueFullName`.)

kelloggm: I'm also concerned that this code might be overspecialized to the case where classes that have type parameters are used as parameters. Can you add a test case where the target method also _defines_ the type variable in use, such as:
```
<T> void foo(List<T> list) { ... }
```

In particular, I'd like a test like this so that we can be certain that this code won't be triggered (creating a class named "T"!) in that case.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -541,10 +572,63 @@ public Visitable visit(MethodCallExpr method, Void p) {
     return super.visit(method, p);
   }
 
+  @Override
+  public Visitable visit(ClassOrInterfaceType typeExpr, Void p) {
+    /**
+     * Workaround for a JavaParser bug: When a type is referenced as a class path, like
```
kelloggm, line 578: "as a class path" is imprecise here. I think you mean "when a type is referenced by its fully-qualified name"

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -541,10 +572,63 @@ public Visitable visit(MethodCallExpr method, Void p) {
...
+     * com.example.Dog dog, JavaParser considers its package components (com and com.example) as
+     * types, too. This issue happens even when the source file of the Dog class is present in the
+     * codebase.
+     */
+    if (!isCapital(typeExpr.getName().asString())) {
```
kelloggm, line 583: This is going to cause us problems if there is a user-written class that begins with a non-capital, I think. That's allowed in Java (but rare, because convention says class names should begin with capitals). However, if we're going to make this assumption about the input, we need to enforce it: that is, Specimin needs to crash if one of the input files contains a class with a lowercase name. Please add a test case that demonstrates that crash, once you've added it.

kelloggm (changes requested):

kelloggm:
@LoiNguyenCS since you aren't feeling well, I'm going to address my own comments on this PR

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
kelloggm: No extraneous class is created, but Specimin does crash in this case, because it looks for the class `T` (i.e., a class with the type variable's name). I'll commit the test on this branch. (I don't think I'll have time to actually fix the bug right now, but maybe I'll do that in a bit.)

LoiNguyenCS:
Thank you, Professor. I appreciate it.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: 3c414c21216c89a421d4f039cb6642849b47d27b

kelloggm:
@LoiNguyenCS I think I've fixed the issues I found with this PR. You should look over my changes; some of them are fairly extensive (I needed to add additional plumbing to keep track of what type variables are in scope at each point). If you don't have any concerns, I think this can be merged.

kelloggm (approved):
LGTM (I wrote a lot of the code)

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -541,10 +602,66 @@ public Visitable visit(MethodCallExpr method, Void p) {
...
+      UnsolvedClass classToUpdate;
+      int numberOfArguments = 0;
+      String typeRawName = typeExpr.getElementType().asString();
+      if (typeArguments.isPresent()) {
+        numberOfArguments = typeArguments.get().size();
+        // without any type argument
+        typeRawName = typeRawName.substring(0, typeRawName.indexOf("<"));
+      }
+      if (isTypeVar(typeRawName)) {
+        // If the type name itself is an in-scope type variable, just return without attempting
```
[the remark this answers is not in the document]
kelloggm: > Aren't MiddleName and LastName in-scope type variables?

No. They are types that are being substituted for type variables, but they are not (necessarily) type variables themselves. This is like a pseudo-assignment, but at the type level: a method or class can declare some type variables (the "LHS" of the "type-level pseudo-assignment"), and then when that method/class is called/instantiated, it is necessary to provide a "RHS": that is, some concrete type that actually fills in for the type variable at that particular use of the method/class. (Of course, that can also be a type variable, in which case the process would continue: at the uses of whatever element contains that type variable, that type variable would need to be substituted, possibly with another typevar, etc. The complexity of this process is why Java generics is part of why Java generics are [technically Turing complete](https://dl.acm.org/doi/10.1145/3093333.3009871).)

kelloggm:
@LoiNguyenCS I've addressed your comments. Please take another look when you have a chance.

LoiNguyenCS (commented):
Professor,

This changes look good to me. I think the CI should be able to run successfully after we merge #51.

kelloggm:
@LoiNguyenCS CI is still failing (Checker Framework is complaining about signature checker errors - I'm guessing a bad interaction with the removal of signature annotations in #51).

LoiNguyenCS:
Professor, it was a consequence of removing annotations. But it's pretty simple to fix.

LoiNguyenCS:
Professor,

Somehow the `Anonymous` test and `TypeVarSimple` test are failing. I will try to figure out the problems.

kelloggm:
Regarding the failures in the `AnonymousClass` test - have you pulled the changes from #52 into this branch? If not, we should do that (that might be what's responsible).

kelloggm (approved):
LGTM, thanks

PR #51, opened 2023-11-22
Script that checks that all test oracles are compilable

kelloggm:
I expect CI to fail. @LoiNguyenCS you should check the failures here and then make PRs that correct the bugs that it exposes. Once CI on this PR is green, we can merge it and then enforce this requirement for tests going forward.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
@@ -1,0 +1,63 @@
+package org.checkerframework.specimin;
+
+import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
+import com.github.javaparser.ast.type.ClassOrInterfaceType;
+import com.github.javaparser.ast.visitor.ModifierVisitor;
+import com.github.javaparser.ast.visitor.Visitable;
+import com.github.javaparser.resolution.UnsolvedSymbolException;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * A visitor that traverses a Java file's AST and creates a map, associating the file name with the
+ * set of types used inside it.
+ */
+public class GetTypesFullNameVisitor extends ModifierVisitor<Void> {
+
+  /** The directory path of the Java file. */
+  private String fileDirectory = "";
+
+  /**
+   * A map that associates the file directory with the set of fully qualified names of types used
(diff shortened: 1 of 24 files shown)


LoiNguyenCS:
Professor, I have a little concern about this script. 

This script appears to compile each program as an isolated component. For instance, consider a scenario with two files: `com/example/Child.java` and `sample/pack/MyType.java`. Let's assume that the `Child` class imports the `MyType` class. Attempting to compile `com/example/Child.java` independently may result in an error because javac lacks information about the location of `MyType.java`.

LoiNguyenCS:
I've figured out the problems. It was due to how Specimin handles import statements. If a class is deleted because it is not used by the target methods, Specimin should delete the import statements for that class in the final output too. Otherwise we've got an "unfound" error.

LoiNguyenCS:
Professor,

I've made two changes on this branch. Specimin will now:
1. Remove import statements that import unused classes.
2. Try to use fully-qualified names to correct types, instead of simple names.

I am not sure why the CI is failing. Please take a look. Thank you.

LoiNguyenCS:
When I tried running "typecheck_test_outputs.sh" in my local machine, I also got an error message without any further explanation.

LoiNguyenCS:
Professor @kelloggm,

I have fixed the failure of CI.

LoiNguyenCS (approved):
LGTM. (since I wrote the codes for these latest commits)

src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
```diff
@@ -0,0 +1,63 @@
...
+   */
+  private Map<String, Set<String>> fileAndAssociatedTypes = new HashMap<>();
+
+  /**
+   * Get the map of files' directories and types used within those files.
+   *
+   * @return the value of fileAndAssociatedTypes
+   */
+  public Map<String, Set<String>> getFileAndAssociatedTypes() {
+    return fileAndAssociatedTypes;
```
kelloggm, line 34: This should be unmodifiable, using `Collections.unmodifiableMap` (or whatever that method is called).

kelloggm (commented):
I'm concerned that your changes remove signature checker annotations. Given the timeline we're on, I'm okay with doing that for now. However, please open an issue indicating that there is a need to add more signature checker annotations on these classes, which we can return to after the ISSTA deadline.

Once you've done that and made the one other change I requested, this is ready to merge without another review.

Issue #50, opened 2023-11-21
org.checkerframework.specimin.AnonymousClass test is flaky


kelloggm:

kelloggm:
I've observed this a few times while working on specimin. Running `./gradlew clean test` usually (but not always) fixes the problem.

kelloggm:
My guess at the cause: somewhere, Specimin is iterating over some map that doesn't guarantee iteration order. Need to investigate this (with the determinism checker? probably overkill)

kelloggm:
Fixed by #52.

PR #52, opened 2023-11-22
Make synthetic class outputs deterministic

kelloggm:
This fixes the flaky `AnonymousClass` test failures that I noticed (#50), as well as another order-dependent test (`org.checkerframework.specimin.SyntheticSuperVariables`).


LoiNguyenCS (approved):
This is great. I will merge it now.

Issue #56, opened 2023-11-27
 The "UnsolvedSymbolVisitor" generates synthetic classes solely for files specified in the "--targetFiles" arguments.


LoiNguyenCS:
Consider these two files, Foo.java and Baz.java:

**Baz.java:**
```
package com.example;
import com.mypackage.ToolA;

public class Baz {
    public static ToolA getTool() {
        return new ToolA();
    }
}
```
**Foo.java:**

```
package com.example;

class Foo {
    void bar() {
        Baz.getTool();
    }
}
```

If the target file is **Foo.java** and the target method is `com.example.Foo#bar()`, then _UnsolvedSymbolVisitor_ will not create any synthetic file for class `ToolA`. And we will get an `UnsolvedSymbolException` when we try to resolve the return type of `Baz.getTool()` with _TargetMethodFinderVisitor_.

Note: class `ToolA` doesn't need to be there for `Baz.getTool()` to be resolvable by _UnsolvedSymbolVisitor_.

LoiNguyenCS:
The tricky part of this issue is that it is difficult to figure out which files are used by the target methods if the corresponding symbols are not solved. This is one of the limitations of JavaParser. Another solution is simply to add all files existing in the codebase to `UnsolvedSymbolVisitor`, instead of just the files specified in the "--targetFiles" arguments.

PR #58, opened 2023-11-28
Minimize fields

LoiNguyenCS:


LoiNguyenCS:
Professor,

This is the PR to minimize fields so that we don't have to include class files related to the initializers of those fields in the final output.

kelloggm:
@LoiNguyenCS this should fix #56, right? In general, it's always helpful to include a link to the issue that a PR is intended to solve in the issue description; for example, see [this PR](https://github.com/typetools/checker-framework/pull/6280) - notice how the first thing in my comment is a link to the associated issue

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -289,7 +301,11 @@ public Visitable visit(NameExpr expr, Void p) {
             || parentNode.get() instanceof FieldAccessExpr)) {
       ResolvedValueDeclaration exprDecl = expr.resolve();
       if (exprDecl instanceof ResolvedFieldDeclaration) {
-        usedClass.add(exprDecl.asField().declaringType().getQualifiedName());
+        // while the name of the method is declaringType(), it actually returns the class where the
+        // field is declared
```
kelloggm, line 305: This comment appears twice, which suggests to me that this logic should be extracted into a method (with this comment in its Javadoc).

I also notice that here you don't catch a possible exception caused by the use of a fully-qualified name, but above you do. Is that because when visiting a `NameExpr`, you're guaranteed that the expression is a simple name (and therefore unqualified)? If so, that would be good to add as a comment.

src/main/java/org/checkerframework/specimin/MethodPrunerVisitor.java
```diff
@@ -106,4 +118,61 @@ public Visitable visit(ConstructorDeclaration constructorDecl, Void p) {
...
+  @Override
+  public Visitable visit(FieldDeclaration fieldDecl, Void p) {
+    String classFullName = fieldDecl.resolve().declaringType().getQualifiedName();
+    boolean isFinal = fieldDecl.isFinal();
+    Iterator<VariableDeclarator> iterator = fieldDecl.getVariables().iterator();
+    while (iterator.hasNext()) {
+      VariableDeclarator varDecl = iterator.next();
+      String varFullName = classFullName + "#" + varDecl.getNameAsString();
+
+      if (methodsToEmpty.contains(varFullName)) {
```
kelloggm, line 131: I think the name `methodsToEmpty` is out-of-date given that this PR deals with fields. We should change `method` in the name of `methodsToEmpty` to `member` - i.e., it should be `membersToEmpty`, since "members" of a class include both its fields and its methods.

There is a similar problem with the class, which is named `MethodPrunerVisitor`, but has now been augmented to also prune fields. I'd rename it `MemberPrunerVisitor` or maybe just `PrunerVisitor`; the class Javadoc will also need to change appropriately.

src/main/java/org/checkerframework/specimin/MethodPrunerVisitor.java
```diff
@@ -106,4 +118,61 @@ public Visitable visit(ConstructorDeclaration constructorDecl, Void p) {
...
+          return new LongLiteralExpr("0L");
+        case FLOAT:
+          return new DoubleLiteralExpr("0.0f");
+        case DOUBLE:
+          return new DoubleLiteralExpr("0.0");
+        case BYTE:
+          return new IntegerLiteralExpr("0");
+        case SHORT:
+          return new IntegerLiteralExpr("0");
+          // If none of the above cases are true, then this type is a char
```
kelloggm, line 170: I'd prefer an explicit case for `CHAR` and a crash if it's something else, which is more future proof. (Maybe a future version of Java will add another primitive type...)

kelloggm (changes requested):

LoiNguyenCS:
> this should fix https://github.com/kelloggm/specimin/issues/56, right?

Sorry for the ambiguity. This PR should fix #41.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
LoiNguyenCS: > This comment appears twice, which suggests to me that this logic should be extracted into a method (with this comment in its Javadoc).

The code `resolve().asField().declaringType().getQualifiedName()` only applies to `FieldAccessExpr` and `NameExpr` that is a field. If we want to create a new method, we'd have to include several checks, which might not be very efficient.

LoiNguyenCS:
Professor,

I have updated the codes based on your comments. Please take a look.

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * This boolean tracks whether the element currently being visited is inside a target method. It
+   * is set by {@link #visit(MethodDeclaration, Void)}.
+   */
+  private boolean insideTargetMethod = false;
+
+  /**
+   * Creates the pruner. All methods this pruner encounters other than those in its input sets will
+   * be removed entirely. For both arguments, the Strings should be in the format produced by
+   * ResolvedMethodDeclaration#getQualifiedSignature.
```
kelloggm, line 64: The last sentence of this comment isn't true anymore (the strings in `membersToEmpty` can also represent fields).

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
```diff
@@ -0,0 +1,182 @@
...
+   *
+   * @param methodsToKeep the set of methods whose bodies should be kept intact (usually the target
+   *     methods for specimin)
+   * @param membersToEmpty the set of members to be empty
```
kelloggm, line 68: "to be empty" looks like a typo. "to be emptied", maybe? That's the passive voice, though, so it would be better to say "the set of members that this pruner will empty"

kelloggm (approved):
LGTM. After my two minor comments are addressed, you can merge without another review from me. Please also close issue #41 when you do.

PR #60, opened 2023-11-30
fix #56

### LINKED ISSUE #56: The "UnsolvedSymbolVisitor" generates synthetic classes solely for files specified in the "--targetFiles" arguments.

LoiNguyenCS:


kelloggm:
@LoiNguyenCS is this ready for review? You haven't requested a review for this PR, so I have not looked at it.

PR #61, opened 2023-12-01
Creating synthetic classes for unsolved FieldAccessExpr

LoiNguyenCS:


LoiNguyenCS:
Professor,

This PR adds the codes and tests for unsolved `FieldAccessExpr` instances. The `updateClassSetWithQualifiedFieldSignature` method in this PR mainly relies on the existing `updateClassSetWithQualifiedStaticMethodCall` method.

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
```diff
@@ -164,17 +164,29 @@ public void updateFieldByType(String currentType, String correctType) {
...
-            UnsolvedSymbolVisitor.setInitialValueForVariableDeclaration(
-                correctType, correctType + " " + fieldName));
+        if (!isStatic) {
+          newFields.add(
+              UnsolvedSymbolVisitor.setInitialValueForVariableDeclaration(
+                  correctType, correctType + " " + fieldName));
+        } else {
+          newFields.add(
+              UnsolvedSymbolVisitor.setInitialValueForVariableDeclaration(
+                  correctType, "static " + correctType + " " + fieldName));
```
kelloggm, line 188: Since these two calls are otherwise identical, I'd prefer to make `isStatic` a `String` variable rather than a boolean:
* if the field is static, then `isStatic` equals `"static "`
* if the field is non-static, then `isStatic` equals the empty string

We might want to rename the variable, too (maybe to just "static"?)

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1465,6 +1508,24 @@ public boolean unsolvedAndCalledByASimpleClassName(MethodCallExpr method) {
...
+   */
+  public boolean unsolvedFieldCalledByASimpleClassName(FieldAccessExpr field) {
+    try {
+      field.resolve();
+      return false;
+    } catch (UnsolvedSymbolException e) {
+      // this check is not very comprehensive, since a class can be in lowercase, and a method or
+      // field can be in uppercase. But since this is without the jar paths, this is the best we can
+      // do.
+      return Character.isUpperCase(field.getScope().toString().charAt(0));
```
kelloggm, line 1525: I don't fully understand why being uppercase tells us whether the field is unsolved. Is there some other invariant here that I'm missing, such as that unsolved fields always are static and need to be accessed via a class name?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1565,6 +1640,62 @@ public void updateClassSetWithQualifiedStaticMethodCall(
...
+    // following its progression through the code.
+    // Suppose this is our field access expression: com.example.MyClass.myField
+    List<String> fieldParts = Splitter.onPattern("[.]").splitToList(fieldExpr);
+    int numOfFieldParts = fieldParts.size();
+    if (numOfFieldParts <= 2) {
+      throw new RuntimeException(
+          "Need to check this field access expression with"
+              + " isAnUnsolvedStaticFieldCalledByAQualifiedClassName before using this method");
+    }
+    String fieldTypeClassName = toCapital(fieldParts.get(0));
```
kelloggm, line 1663: `fieldTypeClassName` is at best a misleading name for this variable. In your example, for instance, I think its value on this line is "Com`, which is not the name of the field's type? I believe it's the start of the synthetic field type name, but that's not obvious from this name. Maybe a comment here would help.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: Professor, I will rename it. We know the field is unsolved because in the visit method for `FieldAccessExpr`, we check to see if the field is solved first before coming to this check.

LoiNguyenCS:
Professor,

I have updated the codes based on your suggestions. Please take a look. Thank you.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: I'm still not fully happy with this, but I'll let it go for now as tech debt given the need to move quickly

kelloggm (approved):

Issue #63, opened 2023-12-04
specimin execution fails for cf-6019


tahiat:
commands: 
`./gradlew run --args='--outputDirectory "../ISSUES/CF/cf-6019/output/" --root "../ISSUES/CF/cf-6019/input/kafka-sensors/src/main/java/" --targetFile "com/fillmore_labs/kafka/sensors/serde/confluent/interop/Avro2Confluent.java" --targetMethod "com.fillmore_labs.kafka.sensors.serde.confluent.interop.Avro2Confluent#transform(String, byte[])"'`

stack tracke: 

```
> Task :run FAILED
Exception in thread "main" java.lang.RuntimeException: Error calculating the type of parameter schema of method call registerInConfluentSchemaRegistry(topic, schema)
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveMethodAsUsage(JavaParserFacade.java:751)
        at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:523)
        at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:99)
        at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:116)
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getTypeConcrete(JavaParserFacade.java:554)
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:401)
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:383)
        at com.github.javaparser.symbolsolver.JavaSymbolSolver.calculateType(JavaSymbolSolver.java:291)
        at com.github.javaparser.ast.expr.Expression.calculateResolvedType(Expression.java:566)
        at java.base/java.util.Optional.map(Optional.java:265)
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:727)
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:657)
        at com.github.javaparser.symbolsolver.JavaSymbolSolver.toResolvedType(JavaSymbolSolver.java:281)
        at com.github.javaparser.ast.type.VarType.resolve(VarType.java:102)
[...]

PR #66, opened 2023-12-05
Fix a crash in Checker Framework issue

LoiNguyenCS:


LoiNguyenCS:
Professor, this is the PR to fix #63. Here is what I have done:

1. Add codes to handle the "var" keyword.
2. Fix the codes for solving unsolved union types.
3. Add codes so that Specimin will not create a synthetic class for a file that already exists.
4. Add codes to add the Exception instance in the catch clause to the symbol table.

I have created the corresponding test cases for all of them.

LoiNguyenCS:
I've just updated the codes a bit to throw an exception when there is a wildcard import.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -271,6 +276,15 @@ public void setExceptionToFalse() {
     gotException = false;
   }
 
+  @Override
+  public Node visit(ImportDeclaration decl, Void arg) {
+    if (decl.isAsterisk()) {
+      throw new RuntimeException(
+          "A wildcard import statement found. Please use explicit import" + " statements.");
```
kelloggm, line 283: Rather than "please use", which sounds like an order, it would be better for this error message to indicate that **we** don't support wildcard imports. I suggest the following text:
"Specimin does not support wildcard import statements, but a wildcard import was found in " + the path to the file.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -430,7 +445,12 @@ public Visitable visit(VariableDeclarator decl, Void p) {
     Type declType = decl.getType();
     try {
       declType.resolve();
-    } catch (UnsolvedSymbolException | UnsupportedOperationException e) {
+    } catch (RuntimeException e) {
+      // nothing to do here. This type is a var type. It is not unsolved.
```
kelloggm, line 449: What actual exception is thrown when this is `var`?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -580,7 +600,13 @@ public Visitable visit(FieldAccessExpr node, Void p) {
     // if the symbol that called this field is solvable yet this field is unsolved, then the type of
     // the calling symbol is a synthetic class that needs to have a synthetic field updated
     else if (canBeSolved(node.getScope())) {
-      updateSyntheticClassWithNonStaticFields(node);
+      Expression nodeSccope = node.getScope();
```
kelloggm, line 603: typo: "Sccope"

kelloggm (changes requested):

LoiNguyenCS:
Professor, I have updated the codes. I have also add two more modifications:

1. A dedicated method to remove the return type in the declaration of a method, since return type can contain space.
2. Remove every unsolved symbol, since those symbols are not used by target methods.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -183,10 +184,9 @@ public Visitable visit(VariableDeclarator node, Void arg) {
   @Override
   public Visitable visit(MethodDeclaration method, Void p) {
     String methodDeclAsString = method.getDeclarationAsString(false, false, false);
-    // The substring here is to remove the method's return type. Return types cannot contain spaces.
+    // The substring here is to remove the method's return type.
```
kelloggm, line 187: This substring has been moved into `removeMethodReturnType`, so this comment doesn't make sense anymore.

kelloggm (approved):
LGTM other than the one minor comment. I don't need to review again.

PR #68, opened 2023-12-07
Add static imports

LoiNguyenCS:


LoiNguyenCS:
Professor,

This is a PR to add codes to handle static imports, since I've seen these imports in some of the issues of CheckerFramework;.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -58,6 +58,7 @@
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import javax.swing.*;
```
kelloggm, line 61: Two notes about this:
* avoid wildcard imports
* I don't see why you are importing anything from `javax.swing`, which is a GUI library

kelloggm (approved):
Overall, this PR looks good. Please fix the one issue that I noticed, but once you do I do not need to review again before this is merged.

PR #70, opened 2023-12-08
Add codes to handle static blocks and constraint types

LoiNguyenCS:
Professor, this is the PR to handle constraint types and static blocks. For static blocks, I think it's safe to remove all of them.


kelloggm (approved):
LGTM, thanks

PR #72, opened 2023-12-09
Preserve annotations

LoiNguyenCS:
This PR is falling right now because javac requires the class files of annotations in order for the input program to compile successfully.


LoiNguyenCS:
Professor, I have updated the codes.

In this current version of Specimin, the jar files used by annotations will be included in the final output.

Please take a look.

LoiNguyenCS:
LGTM.

PR #78, opened 2023-12-22
Handle used methods in inner classes

LoiNguyenCS:
Professor,

This is the updated codes so that Specimin can preserve inner classes that contain used methods, and remove classes that don't.


kelloggm (approved):
LGTM, thanks

Issue #76, opened 2023-12-11
Casting  EnumDeclaration to ClassorInterfaceDeclaration crashes specimin


tahiat:
command: 
`collect/Maps.java" --targetMethod "com.google.common.collect.Maps#difference(Map<? extends K, ? extends V>, Map<? extends K, ? extends V>, Equivalence<? super V>)"'`

Error: 

```
Exception in thread "main" java.lang.ClassCastException: class com.github.javaparser.ast.body.EnumConstantDeclaration cannot be cast to class com.github.javaparser.ast.body.ClassOrInterfaceDeclaration (com.github.javaparser.ast.body.EnumConstantDeclaration and com.github.javaparser.ast.body.ClassOrInterfaceDeclaration are in unnamed module of loader 'app')
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:591)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:74)
        at com.github.javaparser.ast.body.MethodDeclaration.accept(MethodDeclaration.java:98)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
        at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
        at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:428)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
        at com.github.javaparser.ast.body.EnumConstantDeclaration.accept(EnumConstantDeclaration.java:87)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
        at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
        at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:445)
[...]

PR #81, opened 2023-12-25
Handle methods inside enum constant declarations

LoiNguyenCS:


LoiNguyenCS:
Professor,

This is the PR to fix issue #76. 

While it is better if Specimin can remove all enum constant declarations that are not in use, I've found it to be quite difficult to do with JavaParser. For example, with the method `STATUS.ON.testing()`, it is easy to get the class of `testing()`, which is `STATUS`, but it is difficult to get the related the constant declaration, which is `ON`.

kelloggm (approved):

Issue #75, opened 2023-12-10
Specimin fails on cf-3021


tahiat:
**Command:** 
```
./gradlew run --args='--outputDirectory "../ISSUES/CF/cf-3021/output/" --root "../ISSUES/CF/cf-3021/input/guava/guava/src/" --targetFile "com/google/common/collect/MutableClassToInstanceMap.java" --targetMethod "com.google.common.collect.MutableClassToInstanceMap#checkedEntry(final Entry<Class<? extends B>, B>)"'
```

**Stack trace:** 

```
Exception in thread "main" java.lang.RuntimeException: Method 'cast' cannot be resolved in context cast(getKey(), value) (line: 89) MethodCallExprContext{wrapped=cast(getKey(), value)}. Parameter types: [TypeVariable {JPTypeParameter(K, bounds=[])}, TypeVariable {JPTypeParameter(B, bounds=[])}]
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveMethodAsUsage(JavaParserFacade.java:760)
        at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:523)
        at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:99)
        at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:116)
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getTypeConcrete(JavaParserFacade.java:554)
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:401)
        at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:383)
        at com.github.javaparser.symbolsolver.JavaSymbolSolver.calculateType(JavaSymbolSolver.java:291)
        at com.github.javaparser.ast.expr.Expression.calculateResolvedType(Expression.java:566)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.getArgumentsFromMethodCall(UnsolvedSymbolVisitor.java:1262)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.updateUnsolvedClassWithMethod(UnsolvedSymbolVisitor.java:924)
[...]

PR #83, opened 2023-12-26
Add codes to handle methods from superclasses with unsolved parameters.

LoiNguyenCS:
Professor,

This is the PR to fix the exception in #75.


kelloggm (approved):
LGTM

PR #84, opened 2024-01-10
preserved functionality while refactoring for readability and 

jonathan-m-phillips:
maintenance.

While completing this PR, please ensure to do a thorough CR. There were two key changes to the logic - checking if the parent node is present before calling .get(); The rest of the code should provide the same functionality, just with a more readable approach.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
@@ -156,13 +156,15 @@ public class PrunerVisitor extends ModifierVisitor<Void> {
     ResolvedConstructorDeclaration resolved = constructorDecl.resolve();
     if (methodsToLeaveUnchanged.contains(resolved.getQualifiedSignature())) {
       return super.visit(constructorDecl, p);
-    } else if (membersToEmpty.contains(resolved.getQualifiedSignature())) {
+    }
+
+    if (membersToEmpty.contains(resolved.getQualifiedSignature())) {
       constructorDecl.setBody(StaticJavaParser.parseBlock("{ throw new Error(); }"));
       return constructorDecl;
-    } else {
-      constructorDecl.remove();
-      return constructorDecl;
     }
+
+    constructorDecl.remove();
+    return constructorDecl;
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -273,21 +273,24 @@ public static void performMinimization(
...
+          // deciding whether to entirely remove a compilation unit.
+          continue;
         }
-      } else {
+
+        // If the node is a ClassOrInterfaceDeclaration, check if it has members.
+        if (child instanceof ClassOrInterfaceDeclaration cdecl) {
+          if (!cdecl.getMembers().isEmpty()) {
+            return false;
+          }
```
kelloggm, line 287: If this falls through, this function will just return false below, which I don't think is desirable. I think we can simplify this further:
* if `cdecl.getMembers().isEmpty()` is true, continue
* otherwise, return false

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -245,9 +245,11 @@ else if (!type.isUnknownType()) {
         }
 
         if (paramType.isReferenceType()) {
-          String paraTypeFullName =
-              paramType.asReferenceType().getTypeDeclaration().get().getQualifiedName();
-          updateUsedClassWithQualifiedClassName(paraTypeFullName);
+          if (paramType.asReferenceType().getTypeDeclaration().isPresent()) {
```
kelloggm, line 248: Specimin should probably fail loudly (i.e., crash) if `getTypeDeclaration()` is not present, so I'm not sure that this change is a good idea. I'd approve it if you also added an `else` clause that throws a better exception than the `NoSuchElementException` that the code would throw otherwise, but I'd prefer to avoid Specimin silently failing.

src/main/java/org/checkerframework/specimin/UnsolvedMethod.java
```diff
@@ -78,29 +78,31 @@ public void setStatic() {
...
+      + returnTypeInString
+      + name
+      + "("
+      + arguments
+      + ") {"
+      + System.lineSeparator()
+      + "        throw new Error();"
+      + System.lineSeparator()
+      + "    }"
+      + System.lineSeparator();
```
kelloggm, line 106: Why not convert the rest of this method to use `StringBuilder` too? If we're going to do that in the first half of the method, we should do it everywhere while we're at it, I think.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
```diff
@@ -206,28 +208,17 @@ public Visitable visit(FieldDeclaration fieldDecl, Void p) {
...
-          return new IntegerLiteralExpr("0");
-        case CHAR:
-          return new CharLiteralExpr("'\u0000'");
-        default:
-          throw new RuntimeException("Unexpected primitive type: " + fieldType);
-      }
-    } else {
-      return new NullLiteralExpr();
+      return switch (primitiveType) {
+        case BOOLEAN -> new BooleanLiteralExpr(false);
```
kelloggm, line 212: CI is failing because google-java-format is expecting Java 11 code. There are ways to work around it, but I'd prefer to be more portable and just stick to Java 11 features only: I'd like Specimin to run on as many JDKs as possible, since that makes life easier for all of us. (Switch expressions are a Java 14 feature, so that means we can't use them in Specimin if its language level is 11, which it is.)

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
jonathan-m-phillips: I completely overlooked this during the refactor! Thanks..

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
jonathan-m-phillips: Ahhh, I see. Thanks

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
jonathan-m-phillips: Since we are using Java 11, I will remove the instanceof.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
jonathan-m-phillips: You are definitely right - silently failing is a bad idea. It might be beneficial to have a specific exception message to explain to the developer just exactly what is happening. But without running into this exception firsthand, I will rely on the default exception. Deleting this if statement.

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
jonathan-m-phillips: reverting

kelloggm:
Closed in favor of #89

PR #89, opened 2024-01-19
Jp refactor

jonathan-m-phillips:
Using Java 11, consists of the same functionality, but creating better readability for future maintenance. Spotless ran and formatted without errors.

Note: Since I now have commit rights, when I pushed, it went directly to this repository. I am not sure if I did something wrong, but we can complete the PR here if that works.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
@@ -156,13 +156,15 @@ public class PrunerVisitor extends ModifierVisitor<Void> {
     ResolvedConstructorDeclaration resolved = constructorDecl.resolve();
     if (methodsToLeaveUnchanged.contains(resolved.getQualifiedSignature())) {
       return super.visit(constructorDecl, p);
-    } else if (membersToEmpty.contains(resolved.getQualifiedSignature())) {
+    }
+
+    if (membersToEmpty.contains(resolved.getQualifiedSignature())) {
       constructorDecl.setBody(StaticJavaParser.parseBlock("{ throw new Error(); }"));
       return constructorDecl;
-    } else {
-      constructorDecl.remove();
-      return constructorDecl;
     }
+
+    constructorDecl.remove();
+    return constructorDecl;
   }
 
   @Override
@@ -206,28 +208,17 @@ public class PrunerVisitor extends ModifierVisitor<Void> {
   private Expression getBasicInitializer(Type fieldType) {
     if (fieldType.isPrimitiveType()) {
       PrimitiveType.Primitive primitiveType = ((PrimitiveType) fieldType).getType();
-      switch (primitiveType) {
-        case BOOLEAN:
-          return new BooleanLiteralExpr(false);
-        case INT:
-          return new IntegerLiteralExpr("0");
-        case LONG:
-          return new LongLiteralExpr("0L");
-        case FLOAT:
-          return new DoubleLiteralExpr("0.0f");
-        case DOUBLE:
-          return new DoubleLiteralExpr("0.0");
-        case BYTE:
-          return new IntegerLiteralExpr("0");
-        case SHORT:
-          return new IntegerLiteralExpr("0");
-        case CHAR:
-          return new CharLiteralExpr("'\u0000'");
-        default:
-          throw new RuntimeException("Unexpected primitive type: " + fieldType);
-      }
-    } else {
-      return new NullLiteralExpr();
+      return switch (primitiveType) {
+        case BOOLEAN -> new BooleanLiteralExpr(false);
+        case INT, BYTE, SHORT -> new IntegerLiteralExpr("0");
+        case LONG -> new LongLiteralExpr("0L");
+        case FLOAT -> new DoubleLiteralExpr("0.0f");
+        case DOUBLE -> new DoubleLiteralExpr("0.0");
+        case CHAR -> new CharLiteralExpr("'\u0000'");
+        default -> throw new RuntimeException("Unexpected primitive type: " + fieldType);
+      };
     }
+
+    return new NullLiteralExpr();
   }
 }
--- a/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
@@ -208,17 +208,28 @@ public class PrunerVisitor extends ModifierVisitor<Void> {
   private Expression getBasicInitializer(Type fieldType) {
     if (fieldType.isPrimitiveType()) {
       PrimitiveType.Primitive primitiveType = ((PrimitiveType) fieldType).getType();
-      return switch (primitiveType) {
-        case BOOLEAN -> new BooleanLiteralExpr(false);
-        case INT, BYTE, SHORT -> new IntegerLiteralExpr("0");
-        case LONG -> new LongLiteralExpr("0L");
-        case FLOAT -> new DoubleLiteralExpr("0.0f");
-        case DOUBLE -> new DoubleLiteralExpr("0.0");
-        case CHAR -> new CharLiteralExpr("'\u0000'");
-        default -> throw new RuntimeException("Unexpected primitive type: " + fieldType);
-      };
+      switch (primitiveType) {
+        case BOOLEAN:
+          return new BooleanLiteralExpr(false);
+        case INT:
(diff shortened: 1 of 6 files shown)


src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -27,11 +27,7 @@
 import com.github.javaparser.resolution.types.ResolvedReferenceType;
 import com.github.javaparser.resolution.types.ResolvedType;
 import com.google.common.base.Splitter;
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
kelloggm, line 30: Please avoid wildcard imports (our house style is no wildcard imports)

src/main/java/org/checkerframework/specimin/FieldDeclarationsVisitor.java
```diff
@@ -30,6 +30,9 @@ public FieldDeclarationsVisitor() {
 
   @Override
   public void visit(FieldDeclaration decl, Void p) {
+    // Caution: Calling .get() on an Optional without checking if it's present.
+    // This will throw NoSuchElementException if the Optional returned by getParentNode()
+    // is empty.
```
kelloggm, line 35: I don't find this comment very helpful - presumably this code was written based on an assumption that this call to `.get()` will never fail.

In this case, a `NoSuchElementException` can only occur here if there is a FieldDeclaration that does not have a parent node. I don't think that's possible in Java (at least not in Java 11, which is the language level that Specimin targets). So, the code is safe as-is. (A comment that explains why, like what I just wrote here, would be helpful.)

src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
```diff
@@ -39,6 +39,9 @@ public Map<String, Set<String>> getFileAndAssociatedTypes() {
   public Visitable visit(ClassOrInterfaceDeclaration decl, Void p) {
     // Nested type classes don't have a separate class file.
     if (!decl.isNestedType()) {
+      // Caution: Calling .get() on an Optional without checking if it's present.
+      // This will throw NoSuchElementException if the Optional returned by getFullyQualifiedName()
+      // is empty.
```
kelloggm, line 44: @LoiNguyenCS why is this call to `.get()` safe? I don't know this part of the JavaParser API well enough to say.

kelloggm (changes requested):
@LoiNguyenCS should also review

src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
LoiNguyenCS: I have just checked the [documentation](https://javadoc.io/doc/com.github.javaparser/javaparser-core/3.15.6/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.html) of the `getFullyQualifiedName` method. This call to `.get()` is safe because `.get()` will only return an empty value for these two cases:

1. **`decl` is a local class declaration**, which is not possible, since `decl` is a `ClassOrInterfaceDeclaration`,  not a `LocalClassDeclarationStmt`.
2. **`decl` is not inside any compilation unit**, which is not possible, because if that is the case, we would not be able to visit `decl` with a visitor in the first place.

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
```diff
@@ -156,13 +156,15 @@ public Visitable visit(ConstructorDeclaration constructorDecl, Void p) {
     ResolvedConstructorDeclaration resolved = constructorDecl.resolve();
     if (methodsToLeaveUnchanged.contains(resolved.getQualifiedSignature())) {
       return super.visit(constructorDecl, p);
-    } else if (membersToEmpty.contains(resolved.getQualifiedSignature())) {
```
LoiNguyenCS, line 159: Can you clarify why it's preferable to isolate these conditional statements rather than employing an if-elseif-else clause? I think using such a clause would enhance the clarity of logical connections within the code elements.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -189,7 +185,7 @@ public Visitable visit(MethodDeclaration method, Void p) {
     // TODO: test this with annotations
     String methodName = this.classFQName + "#" + removeMethodReturnType(methodDeclAsString);
     // this method belongs to an anonymous class inside the target method
-    if (insideTargetMethod) {
+    if (insideTargetMethod && method.getParentNode().isPresent()) {
```
LoiNguyenCS, line 188: I think the second condition check was unneeded. `method` will definitely have a parent node, since it is inside another method.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -110,6 +110,9 @@ public static void performMinimization(
     SourceRoot sourceRoot = new SourceRoot(Path.of(root));
     sourceRoot.tryToParse();
     for (CompilationUnit compilationUnit : sourceRoot.getCompilationUnits()) {
+      // Caution: Calling .get() on an Optional without checking if it's present.
```
LoiNguyenCS, line 113: This `.get()` will only fail if the Java source file of the related compilation unit is not available, such as in the case of a jar file. 

We are iterating through compilation units based on a `SourceRoot`. According to this [documentation](https://javadoc.io/static/com.github.javaparser/javaparser-core/3.4.3/index.html?com/github/javaparser/utils/SourceRoot.html), the above case will not happen, as a `SourceRoot` is:

> A collection of Java source files located in one directory and its subdirectories on the file system. Files can be parsed and written back one by one or all together.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -245,6 +241,9 @@ else if (!type.isUnknownType()) {
         }
 
         if (paramType.isReferenceType()) {
+          // Caution: Calling .get() on an Optional without checking if it's present.
```
LoiNguyenCS, line 244: Maybe this is because Specimin is using an old version of JavaParser. According to this [documentation](https://www.javadoc.io/doc/com.github.javaparser/javaparser-core/3.5.0/com/github/javaparser/resolution/types/ResolvedReferenceType.html#getTypeDeclaration--), `.getTypeDeclaration` will not return an Optional.

LoiNguyenCS (commented):
Hi, I have left some comments explaining why some `.get()` statements are safe.

src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
kelloggm: Here and elsewhere, rather than putting this information in a GitHub comment please replace the comment in this PR with the justification for why it's safe.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
jonathan-m-phillips: thanks for letting me know - avoiding moving forward

src/main/java/org/checkerframework/specimin/FieldDeclarationsVisitor.java
jonathan-m-phillips: I will remove comments similar to this one as well.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
jonathan-m-phillips: Yes, this should have been removed in later commits but forgot. Nice catch!

src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
jonathan-m-phillips: @kelloggm is this asking me to replace the comments with why they are safe?

kelloggm: Yes

kelloggm:
Superseded by #257.

PR #92, opened 2024-01-27
Update used classes based on type arguments 

LoiNguyenCS:


LoiNguyenCS:
Professor,

This is the PR to fix a bug of Specimin.

Currently, there is a bug where Specimin fails to consider unsolved types if those types only appear as type arguments. For example, Specimin will overlook unsolved types `BigTree `and `SmallTree` in `Map<BigTree, SmallTree>`.

This is one of the bugs that I have figured out while trying to make Specimin work for [AnalysisResult.java](https://github.com/mernst/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java) (but right now there are still more bugs to fix before that happen)

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -478,6 +487,7 @@ public void updateUsedElementWithPotentialFieldNameExpr(NameExpr expr) {
       String classFullName = exprDecl.asField().declaringType().getQualifiedName();
       updateUsedClassWithQualifiedClassName(classFullName);
       usedMembers.add(classFullName + "#" + expr.getNameAsString());
+      updateUsedClassWithPossibleTypeArgs(exprDecl.getType());
```
kelloggm, line 490: I notice that every place you've added a call to this method, there is a call to `updateUsedClassWithQualifiedClassName` immediately beforehand. Is it possible to combine the two methods? As it is, I think it would be easy for someone to call one without calling the other (and it's difficult for me to verify in a code review that you haven't missed somewhere that `updateUsedClassWithQualfiedClassName` is called, since I'm only looking at the diff)

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
LoiNguyenCS: It's not always the case that these two methods appear together. These two methods only appear together when we are updating the list of used classes based on the types of some used members.

When updating the list of used classes based on the used members themselves, we only call `updateUsedClassWithQualifiedClassName`.

For example, given a method declaration, we will add the class containing that method to the list of used classes. In that case, only `updateUsedClassWithQualifiedClassName` is called. When we update the list of used classes based on the return type of that method declaration, then we will call `updateUsedClassWithQualifiedClassName` and `updateUsedClassWithPossibleTypeArgs` together.

kelloggm: Please update the documentation of both methods to explain this.

LoiNguyenCS:
Professor,

I have updated the documentation. I've tried to keep it shorter than the explanation I wrote in this PR. Please take a look. Thank you.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -499,4 +512,27 @@ public void updateUsedClassWithQualifiedClassName(String qualifiedClassName) {
       updateUsedClassWithQualifiedClassName(potentialOuterClass);
     }
   }
+
+  /**
+   * Given the resolved type of a used member, this method will update the list of used classes
+   * based on the type arguments of the input type accordingly.
+   *
+   * <p>We should always call this method whenever we call {@link
+   * TargetMethodFinderVisitor#updateUsedClassWithQualifiedClassName(String)}.
```
kelloggm, line 521: This documentation is inconsistent with what you said in your [PR comment](https://github.com/kelloggm/specimin/pull/92/files#r1469646362), where you said:
> It's not always the case that these two methods appear together. These two methods only appear together when we are updating the list of used classes based on the types of some used members.

However, this documentation says "we should always call this method whenever we call...", which suggests that the two methods should always appear together.

From your earlier comment, I had understood the requirement to be that a call to this method (`updateUsedClassWithPossibleTypeArgs`) must always _follow_ a call to the other method (`updateUsedClassWithQualifiedClassName`). However, not all calls to that method need to have a call to this method after them. Is that correct? If so, please correct this documentation. Otherwise, please correct my understanding.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -478,13 +487,17 @@ public void updateUsedElementWithPotentialFieldNameExpr(NameExpr expr) {
...
+      updateUsedClassWithPossibleTypeArgs(exprDecl.getType());
     }
   }
 
   /**
    * Updates the list of used classes with the given qualified class name and its corresponding
    * enclosing classes, if applicable (in the case of nested classes).
    *
+   * <p>If we are updating the type of a used member, we should call this method together with
+   * {@link TargetMethodFinderVisitor#updateUsedClassWithPossibleTypeArgs(ResolvedType)}.
```
kelloggm, line 499: "together with" is imprecise, because there is a required (or at least expected) ordering: first call this method, then call `updateUsedClassWithPossibleTypeArgs`.

Please also provide a precise definition of `used member` here so that it is clear to a reader what you mean.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
LoiNguyenCS: There is no ordering required, since these two methods do not rely on each other to work. 

> Please also provide a precise definition of used member here so that it is clear to a reader what you mean.


Given that we've already have a field called `usedMembers`, and the Javadoc of that field has already defined what "members" means, I think we can skip the definition of used member in the Javadoc of this method.

kelloggm: > Given that we've already have a field called usedMembers, and the Javadoc of that field has already defined what "members" means, I think we can skip the definition of used member in the Javadoc of this method.

The definition of "used member" can just be "(i.e., a member of the `usedMembers` field of this class)". The point is just to make sure that a reader knows what you are talking about (and this is a big class with many interacting parts, so it's easy for a non-expert to get lost). Put another way, it's better to make these kind of links explicit rather than expecting the reader to remember that the class has a field with the appropriate name. (Also, we might change the name of the field later, and then this documentation might become out of date without anyone noticing.)

> There is no ordering required, since these two methods do not rely on each other to work.

Fair enough. We can skip this part.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
kelloggm: The documentation on `updateUsedClassWithPossibleTypeArgs` remains inconsistent with what you've said in this comment: the documentation says:
> We should **always** call this method **whenever** we call {@link
>  TargetMethodFinderVisitor#updateUsedClassWithQualifiedClassName(String)}.

(emphasis mine.) This differs from what you're saying in these PRs comments (and in the documentation of the other method), which suggest that calling `updateUsedClassWithPossibleTypeArgs` should only be called when `updateUsedClassWithQualifiedClassName` is called on a used member. The documentation of one of the two methods _must_ change, because currently they contradict each other.

Stepping back, might it be possible to encode this in the program itself rather than require the user to understand the relationship of these two methods? In particular, could `updateUsedClassWithQualifiedClassName` check whether its argument is a member of the `usedMembers` field, and if so, call `updateUsedClassWithPossibleTypeArgs` automatically? If we did that, then you could 1) remove all the new calls to `updateUsedClassWithPossibleTypeArgs` in this PR, 2) remove the complexity in this documentation, and 3) prevent any missed calls in the future automatically.

LoiNguyenCS: I think my PR comments have been unclear.

Given a used member, we need to add to the list of used classes:
1: The class containing the declaration of that member.
2: The class corresponding to the type of that member.

If we are doing the first task, then we only need to call `updateUsedClassWithQualifiedClassName`. If we are doing the second task, then we need to call both `updateUsedClassWithQualifiedClassName` and `updateUsedClassWithPossibleTypeArgs`. So `updateUsedClassWithPossibleTypeArgs` will always be called together with `updateUsedClassWithQualifiedClassName` (the second task), while `updateUsedClassWithQualifiedClassName` might be called independently (in case of the first task).

That's why in the documentation of `updateUsedClassWithQualifiedClassName`, I said that we also need to call `updateUsedClassWithPossibleTypeArgs` when we are updating the type of a used member (which is the second task).

> Stepping back, might it be possible to encode this in the program itself rather than require the user to understand the relationship of these two methods? 

I think it will be quite challenging to write codes to differentiate between the class containing a declaration of a member and the class corresponding to the type of that member.

kelloggm: > If we are doing the second task, then we need to call both updateUsedClassWithQualifiedClassName and updateUsedClassWithPossibleTypeArgs

In that case, we should have a method that calls both whose documentation specifically calls out this case. What I want us to avoid is a maintainer needing to call two methods to cover one case but only call one of the two to cover the other - it is always better to have _one_ method to call that covers each situation.

LoiNguyenCS:
Professor,

I have updated the codes so that we don't have to call two methods together every time we update the list of used classes based on the type of a used element. I have decided to use "element" instead of "member", because in the beginning of `TargetMethodFinderVisitor`, a member is defined to be a method or a field only.

PR #95, opened 2024-01-30
Refactor how type variables are handled by UnsolvedClass

### LINKED ISSUE #87: Not Minimizing a Specific Method - Gets Stuck at Run (Urgent)
When running Specimin on `org.checkerframework.dataflow.analysis.AbstractAnalysis#getCurrentTree()` the process just runs without minimizing, throwing an exception, completing, or exiting in anyway. I tried multiple times and one time waited for over 5 minutes before terminating the process.

Link to the method: [AbstractAnalysis#getCurrentTree()](https://github.com/typetools/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/analysis/AbstractAnalysis.java)

I find this issue to be **urgent** because ASHE can drynrun and iterate over exceptions thrown by Specimin, but if we are getting hung up on issues like this, there is no way to bypass but to manually force a skip of this method (hardcode it). This is not the best solution. There might be more methods that get hung up that we are unaware currently.

`./gradlew run --args='--outputDirectory "tempDir" --root "/my/absolute/path/checker-framework/dataflow/src/main/java" --targetFile "org/checkerframework/dataflow/analysis/AbstractAnalysis.java" --targetMethod "org.checkerframework.dataflow.analysis.AbstractAnalysis#getCurrentTree()"'`

There is a chance I am just calling this method incorrectly. If so, please let me know!

kelloggm:
This simplifies the API and gives me confidence that `getClassName` really is the simple class name.

I made these changes as part of my investigation into #87, but this PR does **not** fix #87.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedClass.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedClass.java
@@ -41,25 +41,33 @@ public class UnsolvedClass {
   private boolean isExceptionType = false;
 
   /**
-   * Create an instance of UnsolvedClass
+   * Create an instance of UnsolvedClass. This constructor correctly splits apart the class name and
+   * any generics attached to it.
    *
-   * @param className the name of the class
+   * @param className the name of the class, possibly followed by a set of type arguments
    * @param packageName the name of the package
    */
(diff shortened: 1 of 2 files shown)


LoiNguyenCS (approved):
LGTM

PR #96, opened 2024-01-30
Don't print type variable signature in name of fields

### LINKED ISSUE #87: Not Minimizing a Specific Method - Gets Stuck at Run (Urgent)

kelloggm:
I confirmed on the full test case for #87 that this fixes the reported problem with missing classes like this one that have a type variable signature in the type of a field:
```
package org.checkerframework.dataflow.analysis;
public class OrgCheckerframeworkDataflowAnalysisAbstractAnalysis<V, S, T>CurrentTreeType {
}
```

This PR removes the "<V, S, T>" from this field's type. Unfortunately, this still doesn't fix #87.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -1132,6 +1132,10 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
   public void updateSyntheticClassWithNonStaticFields(FieldAccessExpr field) {
     Expression caller = field.getScope();
     String fullyQualifiedClassName = caller.calculateResolvedType().describe();
+    int indexOfAngleBracket = fullyQualifiedClassName.indexOf('<');
+    if (indexOfAngleBracket != -1) {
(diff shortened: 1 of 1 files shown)


LoiNguyenCS (approved):
LGTM

PR #97, opened 2024-01-31
Implement equality checks for UnsolvedMethod based on the JLS rules

kelloggm:
This solves one infinite loop that #87 triggers, but not all of them. The new test in this PR causes an infinite loop on `main`.


kelloggm:
@LoiNguyenCS please make sure that when you merge this PR, #87 does not get automatically closed by GitHub (that has happened twice already with PRs related to #87, and I have had to go back and correct it).

LoiNguyenCS (approved):
LGTM

PR #98, opened 2024-01-31
Implement equals and hashcode for UnsolvedClass, removing another infinitely-looping test case

kelloggm:
...but #87 still runs forever

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedClass.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedClass.java
@@ -212,9 +212,9 @@ public class UnsolvedClass {
       return false;
     }
     UnsolvedClass otherClass = (UnsolvedClass) other;
-    // This makes me nervous, because it's not clear that it
-    // allows us to differentiate between anonymous inner classes in the same
-    // outer class. TODO: write some tests for that case.
+    // Note: an UnsovledClass cannot represent an anonymous class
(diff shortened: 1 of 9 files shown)


src/main/java/org/checkerframework/specimin/UnsolvedClass.java
```diff
@@ -204,6 +206,24 @@ public void updateFieldByType(String currentType, String correctType) {
     }
   }
 
+  @Override
+  public boolean equals(@Nullable Object other) {
+    if (!(other instanceof UnsolvedClass)) {
+      return false;
+    }
+    UnsolvedClass otherClass = (UnsolvedClass) other;
+    // This makes me nervous, because it's not clear that it
```
LoiNguyenCS, line 215: An `UnsolvedClass` instance represents a synthetic class to be created by Specimin. And we don't create anonymous synthetic classes, so this case will not happen.

LoiNguyenCS (changes requested):
There is just one minor comment that I think we should remove. Other than that, this PR looks good to me.

src/main/java/org/checkerframework/specimin/UnsolvedClass.java
kelloggm: Thanks for correcting this. I wrote a comment there instead to clarify why this is safe.

LoiNguyenCS (approved):
LGTM

PR #99, opened 2024-01-31
Codes to keep original class files

LoiNguyenCS:
Professor,

This is the PR to enforce that original class files are kept untouched.

Given an original class in the input codebase with generic types, such as `Resolved<T, V>`, if that type is used with unsolved type parameters, such as `Resolved<Unsolved, Unsolved>`, Specimin can confuse `Resolved` to be an unsolved type that needs synthetic classes created. It is rather difficult to solve this confusion of Specimin, since JavaParser does not have any straightforward mechanism to separate a type from its type parameters. It's easier to prevent Specimin to make any changes to original class files. And when the unsolved type parameters are solved by the visit method for `ClassOrInterfaceType`, the problem will disappear automatically.


kelloggm (approved):
LGTM

PR #101, opened 2024-02-04
Handling interfaces

LoiNguyenCS:
Professor,

This is the PR to handle interfaces. This PR does four things:

1. Remove unsolved interfaces.
2. Remove unused interfaces.
3. Preserve used method signatures in solved interfaces.
4. Create synthetic classes for unsolved symbols in solved interfaces.

An interface is consider unsolved if we can not find its source codes in the original input.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
@@ -96,6 +96,13 @@ public class TargetMethodFinderVisitor extends ModifierVisitor<Void> {
   private final Map<String, String> importedClassToPackage;
 
   /**
+   * This map connects the declaration of a method to the qualified name of the interface that
+   * contains it, if any.
+   */
+  private final Map<ResolvedMethodDeclaration, ClassOrInterfaceType>
+      methodDeclarationToInterfaceType = new HashMap<>();
+
+  /**
    * Create a new target method finding visitor.
    *
    * @param methodNames the names of the target methods, the format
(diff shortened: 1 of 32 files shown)


src/test/resources/interfacewithunsolvedsymbols/input/com/example/Baz.java
```diff
@@ -0,0 +1,13 @@
+package com.example;
+
+import org.testing.UnsolvedType;
+
+public interface Baz<T> {
+    void doSomething(T value);
```
kelloggm, line 6: Can we modify this test (or add another test) so that we also check that Specimin doesn't crash when the return type of an interface method that we have to keep is unsolved?

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -430,6 +461,30 @@ public Visitable visit(NameExpr expr, Void p) {
     return super.visit(expr, p);
   }
 
+  /**
+   * Updates the list of used classes based on a resolved method declaration. If the input method
+   * originates from an interface, that interface will be added to the list of used classes. The
+   * determination of whether a method belongs to an interface is based on three criteria: method
+   * name, method return type, and the number of parameters.
```
kelloggm, line 468: Note that this is an over-approximation, to handle generics.

(Technically, equality between methods is defined over the _types_ of the parameters, but here we want to preserve any method that _might_ be overridden, without having to solve type parameters. This will be a good example for the paper of how Specimin is over-approximating the set of things that the verifier might use.)

kelloggm (changes requested):
Overall LGTM, but I had a few minor comments. Please resolve those and then I can quickly review again.

LoiNguyenCS:
Professor, I have added the test case.

kelloggm (approved):
Thanks for adding the new test case. LGTM now.

PR #102, opened 2024-02-05
adding evaluation script in CI/CD pipeline

tahiat:


tahiat:
Professor, python script takes long time to complete. Approximately 25 minutes. We can change the trigger event for this workflow so that it triggers less frequently. Maybe it can be triggered only in push event.

kelloggm:
> Professor, python script takes long time to complete. Approximately 25 minutes. We can change the trigger event for this workflow so that it triggers less frequently. Maybe it can be triggered only in push event.

That's fine with me. If you're concerned about it, I'm okay with changing the CI configuration so that the job only runs on `main` (i.e., only once we merge PRs); that way, we can easily see the current state.

The long term solution is to use pash to make the script faster.

PR #103, opened 2024-02-05
Minimize type parameters

LoiNguyenCS:
Professor,

This is the PR to remove unused type bounds of type parameters in the declaration of a class. After this PR and #101 are merged, Specimin will be able to minimize [AnalysisResult.java](https://github.com/mernst/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/analysis/AnalysisResult.java).

### DIFF
--- a/src/test/java/org/checkerframework/specimin/UnusedTypeParameterBoundTest.java
+++ b/src/test/java/org/checkerframework/specimin/UnusedTypeParameterBoundTest.java
@@ -3,12 +3,15 @@ package org.checkerframework.specimin;
 import java.io.IOException;
 import org.junit.Test;
 
-/** This test checks if Specimin can remove unused type parameters in the declaration of a class. */
-public class UnusedTypeParameterTest {
+/**
+ * This test checks if Specimin can remove unused bounds of type parameters in the declaration of a
+ * class.
+ */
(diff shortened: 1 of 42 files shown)


src/test/java/org/checkerframework/specimin/UnusedTypeParameterBoundTest.java
```diff
@@ -0,0 +1,18 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.Test;
+
+/**
+ * This test checks if Specimin can remove unused bounds of type parameters in the declaration of a
+ * class.
```
kelloggm, line 8: Can you please add a test for generic type bounds that should not be removed, as well. We don't currently have one.

kelloggm: (to check that, I searched our `src/test/resources/` directory for "extends", and then observed that none of the results are type variables)

kelloggm (changes requested):

LoiNguyenCS:
Professor,

I have updated the test case to remove one generic type bound and preserve the other type bound.

kelloggm (approved):
Thanks for the new test case. LGTM!

PR #105, opened 2024-02-06
Failing test case where a used method is not created properly by Specimin

kelloggm:
In particular, Specimin does not create the `org.example.Foo#isGood()` method (I'm not sure why).

The goal of this test was to make sure that Specimin correctly makes methods/fields that need to return booleans (because they are used in if guards) actually have boolean returns (instead of a synthetic return class). That part seems to be working (good!), but I'm not sure why the `isGood()` method isn't created. @LoiNguyenCS can you take a look at this and either 1) fix it, or 2) let me know why my test case isn't correct? Thanks.


LoiNguyenCS:
Professor, thank you for the test. The codes to update types for fields were incorrect. I have fixed that code and also created a separate method for ease of future maintenance.

LoiNguyenCS:
For some reasons I can not request reviews from you for this PR. So I will tag your name here: @kelloggm.

kelloggm (commented):
LGTM, but I can't approve (or review, officially) because I originally opened the PR.

PR #106, opened 2024-02-06
Speed up tests by running them in parallel

kelloggm:
Speeds up `./gradlew test` from ~50 seconds to ~15 seconds on my laptop. (Still too slow, but much better!)


LoiNguyenCS (approved):
LGTM.

PR #107, opened 2024-02-06
Another failing test case

kelloggm:
@LoiNguyenCS this test is based on our conversation yesterday, where you hypothesized that a test case like this one would trigger an infinite loop. It doesn't do that, but the output is not compilable: no `B.java` is produced, but `Simple` still extends `B`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -116,6 +116,11 @@ javadoc {
     }
 }
 
+tasks.withType(Test).configureEach {
+    // Creates half as many forks as there are CPU cores.
+    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
+}
+
 test {
     testLogging {
         showStandardStreams = true
--- a/src/test/java/org/checkerframework/specimin/ExtendsSourceButNotTargetTest.java
+++ b/src/test/java/org/checkerframework/specimin/ExtendsSourceButNotTargetTest.java
@@ -1,0 +1,20 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.Test;
+
+/**
+ * This test checks that Specimin can correctly resolve a test case where there is a three-class
+ * chain of extends clauses from the target file/class to another class that is provided as a source
+ * file (but which is not a target). In other words, this test's goal is to make sure that Specimin
+ * is handling files that are source-available but not targets correctly.
+ */
+public class ExtendsSourceButNotTargetTest {
+  @Test
+  public void runTest() throws IOException {
+    SpeciminTestExecutor.runTestWithoutJarPaths(
+        "extendssourcebutnottarget",
+        new String[] {"com/example/Simple.java"},
+        new String[] {"com.example.Simple#bar()"});
+  }
+}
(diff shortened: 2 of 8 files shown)


LoiNguyenCS:
Professor, this is another bug that I am not aware of. It seems to be about inheritance rather than adding required target files.

The bug I discussed yesterday is when class C has an unsolved symbol, such as this:

```
package com.example;

import org.unsolved.Unsolved;

public class C {
    public Unsolved foo() {

    }
}
```
In such cases, we will get an infinite loop.

LoiNguyenCS:
> It seems to be about inheritance rather than adding required target files.

I have tried running your test with "B.java" and "C.java" included as target files, and the output is still the same. I think the reason is that Specimin only preserve a class where the declaration of a used method resides. Since foo() is not declared in B, Specimin does not realize that B should also be included.

kelloggm:
> The bug I discussed yesterday is when class C has an unsolved symbol, such as this:

Thanks. I must have left that out of the notes that I took during our meeting. I'll look into that version of the test case.

kelloggm:
I've added the other test case to [the `a-extends-b-extends-c-loop` branch](https://github.com/kelloggm/specimin/tree/a-extends-b-extends-c-loop). I don't want to try to investigate that until we fix the inheritance issue here (which will also come up in that test case).

LoiNguyenCS:
Professor,

I think we can create a visitor to somehow keep track of the inheritance chains in some forms of data structures (maybe `ArrayList`?). And when we determine which class to preserve, if we have a method/field from class C that is used in class A, and if the two classes belong to the same inheritance chain, then every class along the inheritance path should also be preserved in the final output.

I'm interested in exploring existing algorithms in compiler design that handle inheritance tracking. If such algorithms exist, I believe we should consider adopting them, as I'm uncertain about implementing my proposed approach.

kelloggm:
> I think we can create a visitor to somehow keep track of the inheritance chains

I think this is overcomplicating the issue. If a preserved class `A` inherits from a non-preserved class `B`, we should detect that and add `B` as a preserved class. We can do this recursively (by just checking the `extends` and `implements` clauses of each preserved class, and going around the main loop in `SpeciminRunner` again by setting `UnsolvedSymbolVisitor#gotException` to true if so. So there's no need for an additional data structure: we can just check the immediate ancestors of each preserved class, and then augment the list of classes to be preserved.

kelloggm:
(In my comment above, _preserved_ is defined as "appearing in the output".)

LoiNguyenCS:
I see your point. At the point of `UnsolvedSymbolVisitor`, we don't know which class should be preserved, but we can implement another visitor to run after `TargetMethodFinderVisitor` to make sure that all the classes in the inheritance chain of a preserved class will also be preserved.

kelloggm:
> I see your point. At the point of UnsolvedSymbolVisitor, we don't know which class should be preserved, but we can implement another visitor to run after TargetMethodFinderVisitor to make sure that all the classes in the inheritance chain of a preserved class will also be preserved.

Sure, that seems reasonable to me. It will be a simple visitor.

LoiNguyenCS:
Professor,

I have added a `InheritancePreserveVisitor` to solve this issue. Please take a look. Thank you.

kelloggm (commented):
The changes LGTM

PR #108, opened 2024-02-07
Codes to update the list of target files more comprehensively

LoiNguyenCS:
Professor,

This is the PR to address the problems where classes with unsolved symbols are not visited by `UnsolvedSymbolVisitor` because they are not considered target files. This PR only add those classes to the list of target files if either of the following conditions applies:

1. They have unsolved fields and methods that are accessed in any of the target files.
2. They are extended by any of the target files.

Please let me know if I am missing any cases.

P/S: I have noticed that we already have a method to trace back to the receiver of a method call, so I only needed to change the name of that method and how it is used a little bit.


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -333,19 +333,19 @@ public Visitable visit(ClassOrInterfaceDeclaration node, Void arg) {
     typeVariables.removeFirst();
 
     NodeList<ClassOrInterfaceType> interfaceList = node.getImplementedTypes();
-    for (ClassOrInterfaceType interfaceType : interfaceList) {
+    // Not sure why getExtendedTypes return a list, since a class can only extends at most one class
+    // in Java.
```
kelloggm, line 337: Does it return the entire list of superclasses? If so, that would be useful to us!

(That is, for `class A extends B`, where `class B extends C`, does calling `getExtendedTypes()` on `A` lead to `[B, C]`?)

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -823,6 +834,21 @@ public Visitable visit(ObjectCreationExpr newExpr, Void p) {
     return super.visit(newExpr, p);
   }
 
+  /**
+   * Converts a qualified class name into a relative file path.
+   *
+   * @param qualifiedName The qualified name of the class.
```
kelloggm, line 840: Qualified names shouldn't have angle brackets. Either change the documentation here to say that angle brackets for type variables are permitted, or remove lines 845-847 (I recommend the former).

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1410,22 +1436,23 @@ public static boolean canBeSolved(Expression expr) {
 
   /**
    * This method takes a MethodCallExpr as an instance, and check if the method involved is called
-   * by an incomplete synthetic class. It should be noted that an incomplete synthetic class is
-   * different from a non-existing synthetic class. In this context, an incomplete synthetic class
-   * is a compilable class but missing some methods.
+   * by an incomplete synthetic class. An incomplete class could either be an original class with
```
kelloggm, line 1439: remove the word "synthetic"

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1410,22 +1436,23 @@ public static boolean canBeSolved(Expression expr) {
...
     if (calledByAnUnsolvedSymbol(method)) {
       return false;
     }
     if (method.getScope().isEmpty()) {
       return false;
     }
     try {
-      method.resolve();
+      // use an additional getReturnType() will check the solvability of the method more
+      // comprehensively.
```
kelloggm, line 1454: This comment is vague. Can you amend it to say in what situations this change is useful?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1503,10 +1530,10 @@ public static boolean calledByAnIncompleteSyntheticClass(MethodCallExpr method)
    * @param missedClass the class to be updated
    */
   public void updateMissingClass(UnsolvedClass missedClass) {
+    String qualifiedName = missedClass.getPackageName() + "." + missedClass.getClassName();
```
kelloggm, line 1533: This should probably be a method in `UnsolvedClass` (i.e., this should be a call to a new method `UnsolvedClass#getQualifiedName()`.

src/test/java/org/checkerframework/specimin/UnsolvedMethodInExistingClassTest.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * This test checks if Specimin can properly update the list of target files to be solved by
+ * UnsolvedSymbolVisitor in case there is an unsolved method from a class that is not added as a
+ * target file.
+ */
+public class UnsolvedMethodInExistingClassTest {
+  @Test
+  public void runTest() throws IOException {
+    SpeciminTestExecutor.runTestWithoutJarPaths(
+        "unsolvedmethodsinexistingclass",
```
kelloggm, line 15: Why is the name of the directory different than the name of the class? (There is an extra `s` after `method` in the directory name.)

kelloggm (changes requested):
Overall LGTM, but there are some minor things that need to be fixed.

kelloggm:
> Please let me know if I am missing any cases.

I didn't see any, but that doesn't mean that there aren't some others we are missing. I have an idea about how we can detect missing cases, but I'll make a separate PR for that.

kelloggm:
> I have an idea about how we can detect missing cases, but I'll make a separate PR for that.

My idea did not work (I was thinking of enforcing an invariant in `vistit(MethodCallExpr)` that one of the `if` statements should trigger as long as `gotException` is going to be set to true, but that causes many tests to fail so it's clearly not an invariant).

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: I have just tested that method with a chain of inheritance. Sadly, it doesn't return the entire list. It only returns the immediate superclass.

kelloggm: Too bad. You can ignore my comment, then.

LoiNguyenCS:
Professor,

> My idea did not work (I was thinking of enforcing an invariant in vistit(MethodCallExpr) that one of the if statements should trigger as long as gotException is going to be set to true, but that causes many tests to fail so it's clearly not an invariant).

In that case, we can detect missing cases by testing Specimin on real-life projects and submit new pull requests to address those cases (adding `Block` as a target file in #87 is an example).

I have also updated the codes based on your review. Please take a look. Thank you.

kelloggm (approved):
LGTM, thanks

PR #114, opened 2024-02-12
Solve infinite loops regarding unsolved extensions

LoiNguyenCS:
Professor,

This is to fix the bug mentioned in #87. The solution is that for every class used inside target classes, if that class extends an unsolved class, we'll add that class to the list of target files.


LoiNguyenCS:
Professor, this PR is ready for review. Apart from your original test case, I have added an additional one to show more clearly why we had this bug.

src/main/java/org/checkerframework/specimin/InheritancePreserveVisitor.java
```diff
@@ -15,6 +16,8 @@ public class InheritancePreserveVisitor extends ModifierVisitor<Void> {
   /** List of classes used by the target methods. */
   public Set<String> usedClass;
 
+  /** List of classes to be added to the list of used classes. */
+  public Set<String> addedClass = new HashSet<>();
```
kelloggm, line 20: The documentation should specify that these strings are fully-qualified names.

src/main/java/org/checkerframework/specimin/InheritancePreserveVisitor.java
```diff
@@ -25,20 +28,20 @@ public InheritancePreserveVisitor(Set<String> usedClass) {
...
   /**
-   * Return the current set of used classes.
+   * Return the set of classes to be added to the list of used classes.
    *
-   * @return The set of used classes.
+   * @return The value of addedClass.
    */
-  public Set<String> getUsedClass() {
-    return usedClass;
+  public Set<String> getAddedClass() {
```
kelloggm, line 35: The name of this method is a bit misleading, since it returns a set but the method name is singular. (Admittedly, this issue is inherited from the previous version of this method, but we should fix it while we're here.) I'd prefer `getAddedClasses()`, and I also suggest that we rename the field accordingly.

kelloggm (approved):
My comments are surface level; I don't need to review again after you address them, so you can go ahead and merge this once you've done so.

LoiNguyenCS:
Professor, 

I have updated the codes to fix a bug that took me a bit of time to figure out.

When a field is used inside the target method, the moment we visit the usage of that field within the target method, we've already traversed pass the declaration of that field. So we need a thorough check to make sure that `UnsolvedSymbolVisitor` will loop at least one more time to solve the type of that field.

kelloggm (approved):
LGTM now

Issue #94, opened 2024-01-29
Specimin crashes Nomulus project 


tahiat:
target [method](https://github.com/tahiat/nomulus/blob/c17a5c489c9cc002c6ea61cf515fe5aac1b100d8/core/src/main/java/google/registry/beam/invoicing/InvoicingPipeline.java#L149) 

Specimin gradle command: `./gradlew run --args='--outputDirectory "../ISSUES/CF/cf-3032/output/" --root "../ISSUES/CF/cf-3032/input/nomulus/core/src/main/java/" --targetFile "google/registry/beam/invoicing/InvoicingPipeline.java" --targetMethod "google.registry.beam.invoicing.InvoicingPipeline.GenerateInvoiceRows#expand(PCollection<BillingEvent>)"'`

Stack trace of the exception:

```
> Task :run FAILED
Exception in thread "main" java.lang.RuntimeException: Need to check the method call with unsolvedAndNotSimple before using isAnUnsolvedStaticMethodCalledByAQualifiedClassName
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.updateClassSetWithQualifiedStaticMethodCall(UnsolvedSymbolVisitor.java:1731)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:659)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:76)
        at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:116)
        at com.github.javaparser.ast.visitor.ModifierVisitor.lambda$visit$58(ModifierVisitor.java:653)
        at java.base/java.util.Optional.map(Optional.java:260)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:653)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:684)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:76)
        at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:116)
        at com.github.javaparser.ast.visitor.ModifierVisitor.lambda$visit$58(ModifierVisitor.java:653)
        at java.base/java.util.Optional.map(Optional.java:260)
[...]

PR #117, opened 2024-02-13
Java names cannot contain spaces

kelloggm:
While this does resolve the crash in #94, unfortunately that full test case now infinitely loops. I hope #114 will also fix it.


LoiNguyenCS (approved):
LGTM

Issue #116, opened 2024-02-13
Specimin fails to minimize the program of cf-6388


tahiat:
Specimin command: 
`./gradlew run --args='--outputDirectory "../ISSUES/CF/cf-6388/output" --root "../ISSUES/CF/cf-6388/input/beam/sdks/java/core/src/main/java/" --targetFile "org/apache/beam/sdk/io/FileIO.java"  --targetMethod "org.apache.beam.sdk.io.FileIO#via(Contextful<Fn<UserT, OutputT>>, Contextful<Fn<DestinationT, Sink<OutputT>>>)"'`

**Stack trace:**
```
Exception in thread "main" java.lang.RuntimeException: Need to check the method call with unsolvedAndNotSimple before using isAnUnsolvedStaticMethodCalledByAQualifiedClassName
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.updateClassSetWithQualifiedStaticMethodCall(UnsolvedSymbolVisitor.java:1814)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:701)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:76)
	at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:116)
	at com.github.javaparser.ast.visitor.ModifierVisitor.lambda$visit$79(ModifierVisitor.java:852)
	at java.base/java.util.Optional.map(Optional.java:260)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:852)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.stmt.ReturnStmt.accept(ReturnStmt.java:78)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
	at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:169)
[...]

PR #118, opened 2024-02-13
Attempt to upgrade JavaParser (many tests are failing)

kelloggm:
@LoiNguyenCS can you look into why? Unfortunately I'm out of time today and I probably won't get another chance to look at this for a day or two.

This upgrade is motivated by the fact that we've been using a version of JavaParser from 2021, and I think that the bug in #116 (see my comment there) may have been fixed in a recent JavaParser version. We should be on the latest JavaParser version regardless, since the library is under active development.


LoiNguyenCS:
Professor,

The reason why the test is failing seems to be that this new version of JavaParser no longer removes comments in the final output as expected. I will create another PR to fix this issue.

LoiNguyenCS:
The PR continues to fail due to the updated JavaParser's behavior, which no longer requires strict resolution of a class to calculate the resolved type of an instance. For example, even if there is no constructor for `Foo` within the `Foo` class, the new JavaParser version can still calculate the resolved type of `new Foo("Starting")`. Consequently, the `canBeSolved` method within `UnsolvedSymbolVisitor` is not precise anymore.

kelloggm:
> no longer requires strict resolution of a class to calculate the resolved type of an instance

Hmm, this is interesting. How difficult do you think it will be to refactor our code to adapt to JavaParser's change here?

LoiNguyenCS:
> How difficult do you think it will be to refactor our code to adapt to JavaParser's change here?

Updating the `UnsolvedSymbolVisitor` could be difficult, because it was developed based on the limitation of the older version of JavaParser, which is the requirement of strict resolution of a class to determine the type of a method or field. Over time, we've found this limitation to be a feature rather than a bug (we use it to check if our synthetic classes are completely resolved). Therefore, I suggest postponing the update until later, perhaps after we've submitted the paper.

kelloggm:
Superseded by #266.

PR #119, opened 2024-02-14
Handle implicit calls from interface or superclass

LoiNguyenCS:
Professor,

This is the PR to create synthetic interfaces and superclasses when methods from them are implicitly called.


LoiNguyenCS:
The CI fails at the evaluation script step. I am not sure why.

src/main/java/org/checkerframework/specimin/UnsolvedMethod.java
```diff
@@ -27,6 +27,12 @@ public class UnsolvedMethod {
   /** This field is set to true if this method is a static method */
   private boolean isStatic = false;
 
+  /**
+   * Indicates whether this instance of UnsolvedMethod represents just a method signature without a
+   * body.
+   */
+  public boolean isJustMethodSignature = false;
```
kelloggm, line 34: Why is this field public and mutable? That combination should not be created without a very good reason. I'd prefer a private final field, if at all possible.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -176,6 +178,16 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
   /** New files that should be added to the list of target files for the next iteration. */
   private final Set<String> addedTargetFiles = new HashSet<>();
 
+  /** Stores the set of method names declared in the currently visiting class. */
+  private final Set<String> declaredMethod = new HashSet<>();
```
kelloggm, line 182: Rather than names, should this set contain signatures? I'm thinking of a test case where there is an unsolved superclass/superinterface that declares two overloads of the same method that have different parameter lists, and the target method uses both. For example, something like:
```
void target(boolean b) {
  if (b) { this.foo(1); } else { this.foo(1, "random string"); }
}
```

Please 1) write such a test case, and then 2) consider whether how set should handle overloads.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -176,6 +178,16 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
...
+
+  /**
+   * Maps the name of a class to the unsolved interface that it implements. If a class implements
+   * multiple unsolved interfaces, only the last one is tracked in this map.
```
kelloggm, line 186: Why is only the last interface tracked? (I think the reason is that Specimin wouldn't be able to tell which interface declares which method anyway, but I'd like to be sure.)

Also, please write at least one test case with two unsolved interfaces, so that this behavior is tested.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -321,6 +344,7 @@ public Visitable visit(PackageDeclaration node, Void arg) {
 
   @Override
   public Visitable visit(ClassOrInterfaceDeclaration node, Void arg) {
+    setDeclaredMethod(node.getMethods());
```
kelloggm, line 347: I'm not sure that this is sufficient, because I think the set isn't properly cleared when a class body is exited. The proper abstraction here is probably a stack of sets, similar to the standard algorithm for tracking scopes.

Please add a test case where there are two or more inner classes that demonstrates this issue. If I'm correct, the set of declared methods that are available in the second inner class will still include any methods that are declared in the first inner class. You'll need to make sure that the second inner class uses an unsolved method from an unsolved superclass or superinterface that has the same name/signature as one of the methods that the first inner class declares to show the bug.

tahiat:
Professor, let me look into parallelizing using Pash today. I have not made any progress on that task.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -355,8 +375,29 @@ public Visitable visit(ClassOrInterfaceDeclaration node, Void arg) {
...
+          continue;
+        } catch (UnsolvedSymbolException e) {
+          unsolvedInterface = new UnsolvedClassOrInterface(typeName, packageName);
+        }
+        // this extended/implemented type is an interface if it is in the declaration of an
+        // interface, or if it is used with the "implements" keyword.
+        boolean typeIsAnInterface =
+            node.isInterface() || implementedTypes.contains(implementedOrExtended);
+        if (typeIsAnInterface) {
+          unsolvedInterface.setIsAnInterface();
```
kelloggm, line 391: This should be set during the constructor call (so that the field in `UnsolvedClassOrInterface` can be final).

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -355,8 +375,29 @@ public Visitable visit(ClassOrInterfaceDeclaration node, Void arg) {
...
+          classToItsUnsolvedInterface.put(className, typeName);
+        }
+        updateMissingClass(unsolvedInterface);
       }
     }
+
+    addTypeVariableScope(node.getTypeParameters());
+    Visitable result = super.visit(node, arg);
+    typeVariables.removeFirst();
```
kelloggm, line 400: Why was this change necessary? Did we just miss this before? If so, I'd expect a test for it (and possibly a separate PR, since it seems unrelated to the bug at hand).

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: Professor,

Regarding the `declaredMethod` field, it only records methods declared in the current class. This allows us to recognize methods from superclasses or superinterfaces, even if they are implicitly called, such as `this.foo()` or `baz()` (by checking if `foo` or `baz` is already declared in the current class). Hence, for your test case, `declaredMethod` remains effective.

But your observation raises an important point about the handling method overloads . If `foo(int)` is declared in the current class, and there is an implicit call for a method from a superclass, such as `foo("hello")`, Specimin will not be able to realize that` foo("hello")` is not declared in the current class because it only relies on method name. So I will update Specimin and add corresponding test case.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: These three lines have already existed. I simply changed their position. Previously, they are placed before the inheritance-handling loop. For this PR, we create new synthetic interfaces/superclasses during inheritance-handling loop, so we need to put these three lines after the loop in order to have JavaParser take the newly created synthetic classes into consideration when visiting other nodes.

LoiNguyenCS:
Professor,

I have updated the code based on your review. Please take a look. Thank you.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: Thanks for the explanation - sorry I missed the corresponding deletions.

src/main/java/org/checkerframework/specimin/UnsolvedClassOrInterface.java
```diff
@@ -60,7 +63,7 @@ public UnsolvedClass(String className, String packageName) {
    * @param packageName the name of the package
    * @param isException does the class represents an exception?
    */
-  public UnsolvedClass(String className, String packageName, boolean isException) {
+  public UnsolvedClassOrInterface(String className, String packageName, boolean isException) {
```
kelloggm, line 66: Rather than duplicating the code in this constructor in the new one, this constructor should be a `this(...)` call, just like the constructor directly above it, that adds `false` to its parameter list.

src/main/java/org/checkerframework/specimin/UnsolvedMethod.java
```diff
@@ -38,6 +44,24 @@ public UnsolvedMethod(String name, String returnType, List<String> parameterList
     this.name = name;
     this.returnType = returnType;
     this.parameterList = parameterList;
+    this.isJustMethodSignature = false;
```
kelloggm, line 47: This constructor should also be chained with the new constructor, rather than having duplicated logic.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -176,6 +178,16 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
   /** New files that should be added to the list of target files for the next iteration. */
   private final Set<String> addedTargetFiles = new HashSet<>();
 
+  /** Stores the sets of method names declared in the currently visiting classes. */
```
kelloggm, line 181: These are no longer method names, so this comment needs to be updated.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -176,6 +178,16 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
...
+  private final ArrayDeque<Set<MethodDeclaration>> declaredMethod = new ArrayDeque<>();
+
+  /**
+   * Maps the name of a class to the unsolved interface that it implements. If a class implements
+   * multiple unsolved interfaces, only the last one is tracked in this map.
```
kelloggm, line 186: This makes me nervous, for a reason that I can't really articulate. I have a suspicion that at some point in the future one of us will come back to this comment and wonder "why didn't we just store the whole list of interfaces? that would be really convenient for whatever task I'm trying to do now".

I think it would therefore be better for maintainability if we store the full list of interfaces here for each class, even if the current implementation only _uses_ the last one. The size of the data structure won't get much larger, and the code won't be all that much more complex. But, I think the resulting data structure will be less surprising to future maintainers (i.e., us in a few months).

kelloggm (changes requested):

LoiNguyenCS:
Professor,

I have updated the codes. Please take a look. Thank you.

kelloggm (approved):
LGTM now, thanks

LoiNguyenCS:
Professor,

Should we merge this PR or wait until we get the CI failure solved?

kelloggm:
> Should we merge this PR or wait until we get the CI failure solved?

Let's disable the evaluation script in CI (in this PR is fine) so that CI stays green.

LoiNguyenCS:
I have temporarily disabled the evaluation script.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -850,6 +907,44 @@ public String qualifiedNameToFilePath(String qualifiedName) {
...
+              methodDeclaredTypesOfParameters.add(
+                  parameterTypeResolved.asReferenceType().getQualifiedName());
+            }
+          } catch (UnsolvedSymbolException e) {
+            // UnsolvedSymbolVisitor will not create any synthetic class at this iteration.
+            return false;
+          }
+        }
+        if (methodDeclaredTypesOfParameters.equals(methodTypesOfParameters)) {
+          return true;
```
kelloggm, line 941: @LoiNguyenCS Sorry to bring this up now, when this has already been merged, but I don't think that this logic is correct. In particular, it fails to take into account subtyping: it is legal to call a method with arguments that are subtypes of the declared types of the method, but this method will return false in that case. For example, consider:
```
class A {}
class B extends A {}

class C {
   void test(A a) { ... }
...
   test(new B());
}
```
The call to `test(new B())` is legal and typechecks, but this method would miss that this call is to a method defined in the class. The problem is that the argument and parameter lists don't exactly match up - each argument is a subtype.

I'm unsure if this is causing any of our problems (I observed it while debugging #152, but it is not the proximate cause of that issue I don't think), but we need to come up with a strategy to handle this case properly.

PR #123, opened 2024-02-18
Try to solve less symbols

LoiNguyenCS:
Professor,

This is the PR to make Specimin solves as less symbols as possible. The best thing about this is that the less symbols `UnsolvedSymbolVisitor` try to solve, the less chance Specimin will crash :-) Hopefully this PR makes the evaluation script runs too.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -185,8 +185,8 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
   private final Map<@ClassGetSimpleName String, List<@ClassGetSimpleName String>>
       classToItsUnsolvedInterface = new HashMap<>();
 
-  /** List of target methods as specified by users. */
-  private final Set<String> targetMethodsNames;
+  /** List of signatures of target methods as specified by users. */
+  private final Set<String> targetMethodsSignatures;
 
   /**
    * Fields and methods that could be called inside the target methods. We call them potential-used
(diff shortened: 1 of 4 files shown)


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -185,16 +185,47 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
...
+
+  /** The qualified name of the current class. */
+  private String currentClassQualifiedName = "";
+
   /**
    * Create a new UnsolvedSymbolVisitor instance
    *
    * @param rootDirectory the root directory of the input files
    * @param setOfExistingFiles the set of existing files in the input codebase
+   * @param targetMethodNames the list of target methods as specified by the user.
```
kelloggm, line 220: Is this signatures (as specified by the user) or is it simple names? Clarify this throughout the documentation.

kelloggm: `UnsolvedSymbolVisitor` makes this clear around line 700 that the intention is for these to be signatures, not names. Please change the documentation to reflect that.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -626,64 +691,56 @@ public Visitable visit(FieldDeclaration node, Void arg) {
...
-      }
+    String methodQualifiedSignature =
+        this.currentClassQualifiedName
+            + "#"
+            + TargetMethodFinderVisitor.removeMethodReturnType(
+                node.getDeclarationAsString(false, false, false));
+    String methodSimpleName = node.getName().asString();
+    if (targetMethodsNames.contains(methodQualifiedSignature)) {
+      // this approach will fail if target method a() is inside target method b(), but I don't see
+      // any reason why anyone would pass the arguments to Specimin that way.
```
kelloggm, line 721: Can you add an input validation check for this in `SpeciminRunner`? It should be easy enough to check for overlap in the names and give the user a helpful error message if they ever do this.

kelloggm (changes requested):
Can you try re-enabling the evaluation logic in CI to see whether this does actually speed things up?

I'm not confident that it will, because the tests run in almost exactly the same amount of time on this branch vs on `main`: 1m 10s in both CI builds.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: Professor,

I've just realized that I was overthinking. Right now Specimin does not support having a method inside an anonymous class as a target method, since Specimin determines the signature of a method base on the qualified name of the current non-anonymous class and the declaration of that method (as we can see in the code).

LoiNguyenCS:
Professor,

Compared to the last time the evaluation script ran successfully, this time it runs 10 minutes faster. But we only have one more passing project, which is quite disappointing.

> because the tests run in almost exactly the same amount of time on this branch vs on main

This is because in our tests, the number of unused fields and methods are minimal, so we can't see the impact of this PR clearly.

kelloggm (approved):

kelloggm:
> But we only have one more passing project, which is quite disappointing.

Even so, this is clearly an improvement. Hopefully some of the others will now be easier to debug.

PR #124, opened 2024-02-19
Add a cache to avoid unnecessary calls to `Files.createDirectories`

kelloggm:
The improvement here is pretty minimal - basically in the noise on the tests - but it was easy enough to do, and I confirmed that this cache does get enough hits on the tests to make it plausible that this will be an improvement.

Don't merge until the eval script has run, because I want to see whether this improves performance on larger targets or not before we merge.


LoiNguyenCS (approved):
LGTM

PR #125, opened 2024-02-19
Improve error messages when there is a wildcard import

kelloggm:


kelloggm:
TODO: this PR produces misleading error messages. For example, if target class A extends B, and B contains a wildcard import, this PR (sometimes?) prints that the offending class is A rather than B. We need a more reliable way to tell the user which input file contains the wildcard import. @LoiNguyenCS any ideas?

LoiNguyenCS:
I think the reason is that `className` is determined at the time we visit the declaration of that class. So when we throw the error message at the time we visit the import statements, the `className` that we are using is not accurate, because we've not reached the declaration of that class yet.

PR #126, opened 2024-02-19
Avoid generating .class files

kelloggm:
I've noticed that the tests produce `.class` files locally sometimes, which is definitely undesirable behavior.

I would have preferred to disable code generation rather than just putting the `.class` files somewhere else (since either way code generation is definitely creating I/O we don't need!), but this at least hides the problem from users.


LoiNguyenCS:
LGTM.

Actually Specimin has a method to delete all the .class files in the end. But if Specimin crashes, then that clearing method is never called.

LoiNguyenCS (approved):

PR #128, opened 2024-02-19
Fix a flaky test

kelloggm:
This took a shockingly long time to debug :(

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -3,8 +3,11 @@ package org.checkerframework.specimin;
 import com.google.common.base.Splitter;
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStreamReader;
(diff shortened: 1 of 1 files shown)


kelloggm:
For us the next time we see something like this, here are some notes about what the symptom looked like:
* the other test (`implicitinterfaceaccess`) was failing every other time I ran the tests locally
* the apparent failure was that a file wasn't being found, in particular a `FileNotFoundException` for `BazReturnType`
* the code in which the exception was thrown was `SpeciminRunner#performMinimization`, around line 207. In particular, this is the part of `performMinimization` that's responsible for parsing newly-added target files:
```
    for (String directory : relatedClass) {
      // directories already in parsedTargetFiles are original files in the root directory, we are
      // not supposed to update them.
      if (!parsedTargetFiles.containsKey(directory)) {
        parsedTargetFiles.put(directory, parseJavaFile(root, directory)); // <- exception here
      }
    }
```

LoiNguyenCS (approved):
LGTM. Thank you for this PR. I would understand why it took you a long time. I will be more careful next time I write these test files.

Issue #88, opened 2024-01-14
UnsolvedSymbolException{context='null', name='UniqueId', cause='null'}


jonathan-m-phillips:
This seems more related to the class than method, but there are multiple methods within this class that are giving me the same exception.
Link to class: [ForwardAnalysisImp.java](https://github.com/typetools/checker-framework/blob/master/dataflow/src/main/java/org/checkerframework/dataflow/analysis/ForwardAnalysisImpl.java)

`--args=--outputDirectory "/my/temp/dir" --root "/my/absolute/to/root/checker-framework/dataflow/src/main/java" --targetFile "org/checkerframework/dataflow/analysis/ForwardAnalysisImpl.java" --targetMethod "org.checkerframework.dataflow.analysis.ForwardAnalysisImpl#performAnalysis(ControlFlowGraph)"`

`org.checkerframework.dataflow.analysis.ForwardAnalysisImpl#performAnalysisBlock(Block)`
`org.checkerframework.dataflow.analysis.ForwardAnalysisImpl#getInput(Block)`
etc...

```
Exception in thread "main" UnsolvedSymbolException{context='null', name='UniqueId', cause='null'}
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.toReferenceType(JavaParserClassDeclaration.java:477)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.getAncestors(JavaParserClassDeclaration.java:368)
	at com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration.getAncestors(ResolvedReferenceTypeDeclaration.java:76)
	at com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration.lambda$static$0(ResolvedReferenceTypeDeclaration.java:126)
	at com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration.getAllAncestors(ResolvedReferenceTypeDeclaration.java:115)
	at com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration.getAllAncestors(ResolvedReferenceTypeDeclaration.java:104)
	at com.github.javaparser.symbolsolver.model.typesystem.ReferenceTypeImpl.getAllAncestors(ReferenceTypeImpl.java:208)
[...]

PR #129, opened 2024-02-20
Reset isInsideAnInterface

LoiNguyenCS:
Professor,

This is one of the reasons that cause the uncompilable bug in #88. When `isInsideAnInterface` is not set to false after finishing visiting that interface, it could make Specimin assuming whichever file it visits afterward is also an interface.


LoiNguyenCS:
No ideas what went wrong. Starting another PR.

PR #130, opened 2024-02-20
remove isInsideInterface

LoiNguyenCS:
Professor,

Turns out that the field `isInsideAnInterface` was buggy; once set to `true`, it remained permanently in that state. So I have removed that field. I have also fixed a test file we had in the past, where a used method should have an empty body.

### DIFF
--- a/src/test/resources/MethodInEnum/expected/com/example/Foo.java
+++ b/src/test/resources/MethodInEnum/expected/com/example/Foo.java
@@ -18,9 +18,7 @@ class Foo {
         }
         ;
 
-        public abstract void testing() {
-            throw new Error();
-        }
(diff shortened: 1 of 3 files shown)


src/test/resources/MethodInEnum/expected/com/example/Foo.java
```diff
@@ -18,9 +18,7 @@ public static void testing() {
         }
         ;
 
-        public abstract void testing() {
-            throw new Error();
-        }
+        public abstract void testing();
```
LoiNguyenCS, line 21: This one was originally a method signature. The old version of Specimin added a `throw new Error()` body by mistake.

PR #131, opened 2024-02-20
Make the printed diff command actually useful for debugging when a test fails

kelloggm:
`-q` hides the output, so when locally debugging you previously needed to copy the command and then remove the `-q` by hand, which was rather irritating.


LoiNguyenCS (approved):
LGTM

PR #132, opened 2024-02-20
Handle fields with multiple variable declarators

kelloggm:
Note that for debugging purposes, this PR already includes the change in #131. Merge #131 before reviewing this PR.

@LoiNguyenCS is there a better way to get the full class name in `PrunerVisitor` that you're aware of? I'm not particularly confident in how I did it, but it seems to work on the test cases.

The change in `TargetMethodFinderVisitor` is necessary because without it, code like:
```
baz(obj1);
```
where `obj1` is a field fail to register `obj1` as a used member. Removing that logic didn't cause a test to fail, so I'm not sure why it was there. @LoiNguyenCS do you know why it was originally present? There's no documentation that I can see, and `git blame` doesn't help (it points to https://github.com/kelloggm/specimin/commit/fac443546259ad285f446a2ae39377ac3d26b4d5, which doesn't have any documentation associated with it that I can find).

### DIFF
--- a/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
@@ -198,25 +198,31 @@ public class PrunerVisitor extends ModifierVisitor<Void> {
     if (insideTargetMethod) {
       return super.visit(fieldDecl, p);
     }
-    try {
-      fieldDecl.resolve();
-    } catch (UnsolvedSymbolException e) {
-      // The current class is employed by the target methods, although not all of its members are
-      // utilized. It's not surprising for unused members to remain unresolved.
-      fieldDecl.remove();
-      return fieldDecl;
-    }
-    String classFullName = fieldDecl.resolve().declaringType().getQualifiedName();
     boolean isFinal = fieldDecl.isFinal();
+    // Fields must have parents, because they must be contained in a class.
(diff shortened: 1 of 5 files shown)


kelloggm:
This fixes a crash reported by the evaluation tool on CF-6077 (Apache Cassandra)

LoiNguyenCS:
Professor,

> @LoiNguyenCS do you know why it was originally present? 

In the `visit` method for `MethodCallExpr`, we check if a method call is invoked by a `NameExpr` that represents a field. Previously, I overlooked the scenario where the field might be a parameter within a method call. Thus, to avoid redundant updates for the same scenario (a field invoking a method), I've incorporated that check into the `visit` method for `NameExpr`.

kelloggm:
> We can use `declaringType().getQualifiedName()`,

Actually we cannot: `declaringType()` only exists on the return type of `resolve()` if the receiver is a `FieldDeclaration`; this method isn't present on the return type of `resolve()` when the receiver is a `VariableDeclarator` (that's why I had to make the change at all - calls to `resolve` on a `FieldDeclaration` with multiple declarators always fail).

LoiNguyenCS:
I see, in that case, I think using multiple `get()` is the way to go.

LoiNguyenCS (approved):
LGTM

kelloggm:
@LoiNguyenCS I did a bit of refactoring based on your last comment, so please re-review

LoiNguyenCS (approved):
LGTM.

PR #133, opened 2024-02-20
Typecheck script broken

kelloggm:
...which I discovered while debugging another problem. I've fixed one test input, but need to run to a meeting. I wanted to make you aware of this problem @LoiNguyenCS (see the CI output from the task that checks whether the expected outputs compile for details).

### DIFF
--- a/src/main/java/org/checkerframework/specimin/FieldDeclarationsVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/FieldDeclarationsVisitor.java
@@ -38,8 +38,8 @@ public class FieldDeclarationsVisitor extends VoidVisitorAdapter<Void> {
         (ClassOrInterfaceDeclaration) decl.getParentNode().get();
     SimpleName classNodeSimpleName = classNode.getName();
     String className = classNodeSimpleName.asString();
-    for (VariableDeclarator var : decl.getVariables()) {
-      fieldNameToClassNameMap.put(var.getNameAsString(), className);
+    for (VariableDeclarator variableDecl : decl.getVariables()) {
+      fieldNameToClassNameMap.put(variableDecl.getNameAsString(), className);
     }
   }
 
--- a/src/main/java/org/checkerframework/specimin/FieldDeclarationsVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/FieldDeclarationsVisitor.java
@@ -1,6 +1,8 @@
 package org.checkerframework.specimin;
 
+import com.github.javaparser.ast.Node;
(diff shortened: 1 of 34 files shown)


kelloggm:
I am working on fixing these.

kelloggm:
@LoiNguyenCS I've fixed all of the issues except for the following two, which a simple change to the test can't fix - I think there's something wrong with how we're handling interfaces:
```
./com/example/Foo.java:3: error: Foo is not abstract and does not override abstract method doSomething(int) in Baz
class Foo implements Baz<String> {
^
1 error
Running javac on interfacewithgenerictype/expected issues one or more errors, which are printed above.
./com/example/Foo.java:3: error: Foo is not abstract and does not override abstract method doSomething(int) in Baz
class Foo implements Baz<String> {
^
1 error
Running javac on interfacewithunsolvedsymbols/expected issues one or more errors, which are printed above.
```

I tried adding an override of the `doSomething(int)` method in the input version of `Foo` (which I think is required to make `Foo` compile - actually, `Foo` should implement all of the interfaces!), but it's not preserved in the output in `Foo` (only in the interface `Baz`). Can you look into this one?

LoiNguyenCS:
Professor,

I think this bug comes back to the problem of handling overriding. For interfaces, we believe a method is overriding a method signature if they have the same return type, name, and **number of parameters**. This leads to some method signatures not removed even if they are not overridden by any target methods.

kelloggm:
> I think this bug comes back to the problem of handling overriding. For interfaces, we believe a method is overriding a method signature if they have the same return type, name, and number of parameters. This leads to some method signatures not removed even if they are not overridden by any target methods.

I agree. However, the output still needs to compile, and the current approach doesn't guarantee that. I think in these cases we either need to make sure that the extra interface methods have default implementations or that the overrides are also preserved. I'm not sure exactly how to accomplish that, which is why I'd like you to look into it.

LoiNguyenCS:
> I think in these cases we either need to make sure that the extra interface methods 

Since we can't know which interface methods are extra, I think we can make a default implementation for all the interface methods that are not removed.

LoiNguyenCS:
> Since we can't know which interface methods are extra, I think we can make a default implementation for all the interface methods that are not removed.

As I can't make changes to this branch and the CI won't run unless I am merging something into the main branch, I will make another PR to implement this solution to the main branch.

LoiNguyenCS:
I think this PR is good for merging.

kelloggm:
@LoiNguyenCS this still needs an approval from you before it will auto-merge

LoiNguyenCS (approved):

PR #138, opened 2024-02-20
Resolve crash when encountering a field inside an enum

kelloggm:
Fixes a crash in CF-6388, although I haven't yet checked if that target has other problems. The proximate cause of the crash was `FieldDeclarationsVisitor`, with the following stack trace:
```
Exception in thread "main" java.lang.ClassCastException: class com.github.javaparser.ast.body.EnumDeclaration cannot be cast to class com.github.javaparser.ast.body.ClassOrInterfaceDeclaration (com.github.javaparser.ast.body.EnumDeclaration and com.github.javaparser.ast.body.ClassOrInterfaceDeclaration are in unnamed module of loader 'app')
	at org.checkerframework.specimin.FieldDeclarationsVisitor.visit(FieldDeclarationsVisitor.java:38)
	at org.checkerframework.specimin.FieldDeclarationsVisitor.visit(FieldDeclarationsVisitor.java:20)
	at com.github.javaparser.ast.body.FieldDeclaration.accept(FieldDeclaration.java:116)
...
```

However, fixing that led me to discover a cascade of other enum-related problems (because I wrote a test case). This PR therefore does all of the following:
* updates the logic in `FieldDeclarationsVisitor` to also consider fields contained inside enums
* preserves all enum constructors in used classes (because enum constants, which are all currently preserved, might use them)
* fixes the logic in `PrunerVisitor#getEnclosingClassName` to also account for enums
* adds two new test cases: `FieldInEnumTest` for the crash, and `UnrelatedEnumTest` to check that my enum constructor preservation technique doesn't preserve _all_ enum constructors (just those in classes that would otherwise be preserved)

### DIFF
--- a/src/main/java/org/checkerframework/specimin/FieldDeclarationsVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/FieldDeclarationsVisitor.java
@@ -1,6 +1,8 @@
 package org.checkerframework.specimin;
 
+import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
+import com.github.javaparser.ast.body.EnumDeclaration;
 import com.github.javaparser.ast.body.FieldDeclaration;
 import com.github.javaparser.ast.body.VariableDeclarator;
 import com.github.javaparser.ast.expr.ObjectCreationExpr;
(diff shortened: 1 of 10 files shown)


LoiNguyenCS (approved):
This PR is great. Thank you.

PR #140, opened 2024-02-21
Solve overriding issue

LoiNguyenCS:
Professor,

This is the PR to handle method overriding. This PR makes me feel a bit proud :-) 


src/main/java/org/checkerframework/specimin/SolveMethodOverridingVisitor.java
```diff
@@ -0,0 +1,129 @@
...
+   */
+  private void checkForOverridingAndUpdateUsedClasses(MethodDeclaration methodDeclaration) {
+    // just a method signature, no need to check for overriding.
+    if (methodDeclaration.getBody().isEmpty()) {
+      return;
+    }
+    BlockStmt methodBody = methodDeclaration.getBody().get();
+    // JavaParser does not support solving overriding, but it does support solving super
+    // expressions. So we make a temporary super expression to figure out if this current method is
+    // overriding.
```
kelloggm, line 91: Clever! Nice plan :)

src/main/java/org/checkerframework/specimin/SolveMethodOverridingVisitor.java
```diff
@@ -0,0 +1,129 @@
...
+   * wrapped inside another node, causing removal failure. This method iterates through the parent
+   * nodes until it successfully removes the specified node.
+   *
+   * <p>If this explanation does not make sense to you, please refer to the following link for
+   * further details: <a
+   * href="https://github.com/javaparser/javaparser/issues/858">https://github.com/javaparser/javaparser/issues/858</a>
+   *
+   * @param node The node to be removed.
+   */
+  private void removeNode(Node node) {
```
kelloggm, line 124: Rather than making this a private instance method of this class, I would make this a static method in some kind of `JavaParserUtils` class (that's not instantiable), because it might be useful elsewhere.

src/main/java/org/checkerframework/specimin/SolveMethodOverridingVisitor.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  /**
+   * Removes a node from its compilation unit. If a node cannot be removed directly, it might be
+   * wrapped inside another node, causing removal failure. This method iterates through the parent
+   * nodes until it successfully removes the specified node.
+   *
+   * <p>If this explanation does not make sense to you, please refer to the following link for
+   * further details: <a
+   * href="https://github.com/javaparser/javaparser/issues/858">https://github.com/javaparser/javaparser/issues/858</a>
```
kelloggm, line 120: This link is very helpful to understand why this is necessary, thanks.

kelloggm (approved):
I have one easy to-carry-out suggestion, but otherwise this PR looks great. I don't need to review again before you merge this once you've moved `removeNode` and made it static.

PR #141, opened 2024-02-21
Handle synthetic types that extend Throwable

LoiNguyenCS:
Professor,

This PR addresses the issue that some synthetic types need to extend `Throwable` in order for the output to compile.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -45,7 +45,7 @@ class JavaTypeCorrect {
   private Map<String, Set<String>> fileAndAssociatedTypes = new HashMap<>();
 
   /** Synthetic types that need to extend Throwable. */
-  private Set<String> typesToExtendsThrowable = new HashSet<>();
+  private Set<String> typesThatExtendThrowable = new HashSet<>();
 
   /**
    * Create a new JavaTypeCorrect instance. The directories of files in fileNameList are relative to
@@ -76,10 +76,10 @@ class JavaTypeCorrect {
   /**
    * Get the names of synthetic classes that should extend Throwable.
    *
-   * @return the value of typesToExtendsThrowable.
+   * @return the value of typesThatExtendThrowable.
    */
-  public Set<String> getTypesToExtendsThrowable() {
-    return typesToExtendsThrowable;
+  public Set<String> getTypesThatExtendThrowable() {
+    return typesThatExtendThrowable;
   }
 
   /**
@@ -170,7 +170,7 @@ class JavaTypeCorrect {
       String incorrectType = splitErrorMessage.get(4);
       String correctType = splitErrorMessage.get(splitErrorMessage.size() - 1);
       if (correctType.equals("Throwable")) {
-        typesToExtendsThrowable.add(incorrectType);
+        typesThatExtendThrowable.add(incorrectType);
       } else {
         typeToChange.put(incorrectType, tryResolveFullyQualifiedType(correctType, filePath));
       }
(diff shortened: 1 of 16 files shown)


src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
```diff
@@ -43,6 +44,9 @@ class JavaTypeCorrect {
    */
   private Map<String, Set<String>> fileAndAssociatedTypes = new HashMap<>();
 
+  /** Synthetic types that need to extend Throwable. */
+  private Set<String> typesToExtendsThrowable = new HashSet<>();
```
kelloggm, line 48: I find the name of this field a bit difficult to parse. Let's change it to something that reads like an English phrase, like `typesThatExtendThrowable`.

kelloggm: (the getter, etc., will need to be updated accordingly, too)

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -2275,6 +2275,30 @@ public void updateTypes(Map<String, String> typeToCorrect) {
...
+   * synthetic classes as needed.
+   *
+   * @param typesToExtendThrowable the set of synthetic types that need Throwable extension.
+   */
+  public void updateTypesToExtendThrowable(Set<String> typesToExtendThrowable) {
+    for (String typeToExtendThrowable : typesToExtendThrowable) {
+      Iterator<UnsolvedClassOrInterface> iterator = missingClass.iterator();
+      while (iterator.hasNext()) {
+        UnsolvedClassOrInterface missedClass = iterator.next();
+        // Class comparison is based on class name and package name only.
```
kelloggm, line 2289: The comparison below is only based on the simple name - not on the package. I think we could pretty easily construct a test where this matters - for example, does the following behave correctly (if it's a target method)?:
```
org.packageone.MyThrowable get() throws org.packagetwo.MyThrowable { return null; }
```

(It might be necessary to move around the order of the actual throwable and the fake throwable, because this loop returns after it finds the first one - so if it finds the right one first, this code will continue to work.)

In general, I don't think we should do class comparisons based on simple names only if at all possible, and this is an example where we should use the fully-qualified name, regardless of whether you can get a test like the above to fail (although I strongly suggest that you try to do so, and add a test for that case regardless).

kelloggm (changes requested):
Overall good, but we need to use fully-qualified names to keep track of the classes that need to extend throwable.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: Professor,

It's quite difficult to use whole qualified names because the error messages in Javac only tell us the simple name of types.

I don't think this current approach can cause any compilation error. The worst case scenario is that if we have two synthetic classes with the same name, and one of them should extends `Throwable`, Specimin will make both of them extend `Throwable`. Specimin doesn't support extension for synthetic classes, apart from `extends Exception` and `extends Throwable`. So we should not have any conflict.

kelloggm: > The worst case scenario is that if we have two synthetic classes with the same name, and one of them should extends Throwable, Specimin will make both of them extend Throwable

I see your point here. Please add a test case to demonstrate this imprecision, at least, so that we have documentation of Specimin's behavior in this scenario.

LoiNguyenCS:
Professor, 

I have added the imprecision test case and edited `updateTypesToExtendThrowable` to avoid `ConcurrentModificationException`.

Please re-review.

kelloggm (approved):
LGTM. Note that I added some more explanations to two comments that I felt where leaving out something important.

PR #142, opened 2024-02-21
Adding dependencies in the Specimin build jar

tahiat:
The default jar does not contain the dependent class files. It throws `ClassNotFoundException`.


tahiat:
tangled with other commit. closing and pushing again

PR #144, opened 2024-02-22
Move to updateUsedClassWithQualifiedClassName

LoiNguyenCS:
Professor,

This is the PR to replace every `usedClass.add(...)` with the call to `updateUsedClassWithQualifiedClassName`. This method call is better because it also updates potential outer classes, in case the used classes are inner.


kelloggm (approved):
LGTM

PR #145, opened 2024-02-22
Add Exception

LoiNguyenCS:
Professor,

This is one of the bug I've figured out while looking at the `CheckerFramework `issue.

I think the comment is clear enough to discourage future maintainers from removing this additional catch, so I did not add a test case.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
@@ -188,7 +188,9 @@ public class TargetMethodFinderVisitor extends ModifierVisitor<Void> {
       try {
         updateMethodDeclarationToInterfaceType(
             interfaceType.resolve().getAllMethods(), interfaceType);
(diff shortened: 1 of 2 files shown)


kelloggm (changes requested):
I'd strongly prefer that we add a test case. If you're unwilling to write one for some reason, please let me know why; I think this PR _could_ be merged without one, but doing so makes me nervous (since this is supposed to fix a crash, no?).

LoiNguyenCS:
Professor,

I will add a test case. But at the moment this PR is not ready to be merged, the reason is that I am still not sure what to do if resolving a type ends up in an `IlegalAgrumentException`.

LoiNguyenCS:
Upon further investigation, I have realized that this is a bug of Specimin rather than a false exception of JavaParser.

The bug happens because for some unsolved types with type variables, Specimin create synthetic versions of them with no type variables.

LoiNguyenCS:
Closing this PR in favor of #146.

PR #146, opened 2024-02-25
Handle IlegalArgumentException

LoiNguyenCS:


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1071,6 +1059,37 @@ public boolean declaredInCurrentClass(MethodCallExpr method) {
     return false;
   }
 
+  /**
+   * Resolves symbols for a given ClassOrInterfaceType instance, including its type variables if
+   * present.
+   *
+   * @param typeExpr The ClassOrInterfaceType instance to resolve symbols for.
+   */
+  private void solvedSymbolsForTypes(ClassOrInterfaceType typeExpr, boolean isAnInterface) {
```
kelloggm, line 1068: I think the name of this method should be "solveSymbols" or "solveSymbolsFor". The "solved" (with a "d") in the current name is quite confusing, since it suggests that the symbols must already be solved when this method is called. I'm also not sure what "for types" is doing in the name: does this not just solve all of the symbols related to the input type? That is, what other types does the method name reference?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1071,6 +1059,37 @@ public boolean declaredInCurrentClass(MethodCallExpr method) {
...
+      String packageName = typeRawName.substring(0, typeRawName.lastIndexOf("."));
+      String className = typeRawName.substring(typeRawName.lastIndexOf(".") + 1);
+      classToUpdate = new UnsolvedClassOrInterface(className, packageName);
+    } else {
+      String packageName = classAndPackageMap.getOrDefault(typeRawName, currentPackage);
+      classToUpdate = new UnsolvedClassOrInterface(typeRawName, packageName);
+    }
+    classToUpdate.setNumberOfTypeVariables(numberOfArguments);
+    if (isAnInterface) {
+      classToUpdate.isAnInterface = true;
```
kelloggm, line 1088: Why can't this be set by the constructor calls on lines 1081 and 1084? If you did that, you wouldn't need to make this field non-final or public, I think, so that would be worthwhile.

src/main/java/org/checkerframework/specimin/UnsolvedClassOrInterface.java
```diff
@@ -46,7 +46,7 @@ public class UnsolvedClassOrInterface {
   private boolean isThrowable = false;
 
   /** This field records if the class is an interface */
-  private final boolean isAnInterface;
+  public boolean isAnInterface;
```
kelloggm, line 49: As a general rule, fields should be at most one of `public` and non-`final`, if at all possible. So, this change makes me very nervous (as it makes this field both).

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
LoiNguyenCS: Professor,

I am not sure about the issue here. It seems to me that implementing a setter method for a private field is not different from simply making that field public.

Could you please provide clarification on this matter?

kelloggm: > It seems to me that implementing a setter method for a private field is not different from simply making that field public.

You wouldn't need a public setter for the field if the only assignment is in the constructor, which is what I'm suggesting. Ideally, fields shouldn't be writable outside the enclosing class at all: that is, a class should control writes to its own fields. Logically this makes sense, because the field should be _encapsulated_ within the class if it logically belongs to that class.

kelloggm: Happy to explain this more in our meeting later today if this explanation doesn't work.

PR #147, opened 2024-02-25
A better way to get packages' names

LoiNguyenCS:
Professor,

This PR handles the other bug that cause `IlegalArgumentException`. I can't make a test for this because I can't isolate the bug. This PR is an attempt to solve cf-3012. So I think we can add cf-3012 as a test case after solving all of its issues,


kelloggm (approved):
LGTM other than one minor comment. This can be merged once my comment is addressed.

PR #155, opened 2024-02-26
Debug logic for infinite loops

kelloggm:
This is the second time I've had to write this logic, so I figured we should include it in UnsolvedSymbolVisitor directly. It shouldn't have a perf impact as long as `DEBUG` is false, but it makes it easy to switch it on when it's needed (when debugging an infinite loop caused by `gotException` being set to true even though nothing is changing).


LoiNguyenCS (approved):
This PR looks great to me.

PR #156, opened 2024-02-27
Add check for infinite loop

LoiNguyenCS:
Professor,

This is the PR to get Specimin to throw an exception whenever `UnsolvedSymbolVisitor` is stuck. While the error message is not very useful, I think it's convenient to know whenever we are in an infinite loop.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
+++ b/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
@@ -125,7 +125,15 @@ public class SpeciminRunner {
     // The set of path of files that have been created by addMissingClass. We will delete all those
     // files in the end.
     Set<Path> createdClass = new HashSet<>();
+    // Set containing paths of files created by all runs of addMissingClass before the current one.
+    Set<Path> createdFilesLastRun = new HashSet<>();
     while (addMissingClass.gettingException()) {
+      if (!createdFilesLastRun.isEmpty()) {
+        if (createdClass.equals(createdFilesLastRun)) {
+          throw new RuntimeException("UnsolvedSymbolVisitor is not making any progress.");
+        }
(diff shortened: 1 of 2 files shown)


kelloggm (commented):
I don't see any issues with this change at a low level. However, I'm concerned about the design: what happens if we forget to update `makeProgress` when we make changes to Specimin in the future?

Moreover, this new field feels like it fulfills a similar purpose to `gotException`, and will fail in the same ways. In general, this sort of design where there is a boolean that needs to be set whenever something is or is not updated is extremely brittle, because it's very easy for the maintenance programmer to allow it to drift out-of-sync. For that reason, I don't think I support merging this PR as-is: I think that while it might provide a bit of short-term relief from the infinite loop problems caused by `gotException` not being set properly, it will create its own long-term problems of a similar kind and therefore won't be worth it in the long run.

If we are going to add a check like this to detect whether we are making progress, it is imperative that it not require us to maintain a separate flag beyond what `UnsolvedSymbolVisitor` already maintains. If instead we were able to extract `UnsolvedSymbolVisitor`'s logical state before it runs, and then compare that to the logical state at the end to set the `makeProgress` variable, I could support a change like this.

@LoiNguyenCS does my concern here make sense? If not, I'm happy to have a meeting to discuss.

LoiNguyenCS:
> @LoiNguyenCS does my concern here make sense?

Professor, I believe that if future maintainers write code to perform additional symbol solving to address certain bugs but forget to update `makeProgress` as needed, they will encounter an exception when executing Specimin on either a relevant test case or the original input where the bug occurs. I think that exception throwing mechanism will remind them to update `makeProgress`.

I would be happy to have a meeting to discuss this issue further. I will be available for tomorrow's afternoon.

LoiNguyenCS:
Closing in favor of #159

PR #159, opened 2024-02-28
A different test for infinite loop

LoiNguyenCS:
Professor,

This PR proposes another way to check for infinite loops. When I was trying to fix a mysterious bus related to `Equivalence.Wrapper`, I realized that comparing the string representations of synthetic classes is already a comprehensive comparison. It includes all the important details like class names, extensions, package names, fields, and methods. Thus, I have decided to not use `Equivalence`.

P/S: The CI has been failing with exit code 143 since the second commit (0330c01), where essentially no changes were applied to Specimin. This implies that the CPU limitation was reached due to changes in `specimin-evaluation` rather than the modifications introduced in this PR.


kelloggm:
@LoiNguyenCS you should be able to run these tests on your laptop via `./gradlew test`. Is that not working for some reason?

LoiNguyenCS:
@LoiNguyenCS you should be able to run these tests on your laptop via ./gradlew test. Is that not working for some reason?

Professor, I need to use a virtual Ubuntu machine to run these tests, which take too long to see the results. Right now Specimin does not support testing for Windows.

kelloggm:
> Right now Specimin does not support testing for Windows.

I forgot that the tests have that limitation, so fair enough. In my
experience, it's worth the time investment to get a sensible local testing
setup, even if that means you need to dual boot. Have you tried this under
WSL? It works for almost all unix programs at this point, and if it doesn't
work for Specimin that might be worth fixing.

LoiNguyenCS:
>  Have you tried this under
> WSL? It works for almost all unix programs at this point, and if it doesn't
> work for Specimin that might be worth fixing.

I am using WSL, and it does work for Specimin. It just takes too long to build and run the tests. Also, it causes my computer to lag while it's running. I prefer the CI because it runs faster, does not interfere with my local computer, and shoots me an email whenever it fails :-)

kelloggm:
@tahiat, we're once again reaching the limits of a single CI job to run specimin eval.

One easy solution might be to split the json file into two and turn this into two CI jobs:
* one for the main evaluation script (i.e., the ones we plan to include in the paper)
* one for other bugs that we've found (e.g., the sp-* targets)

Can you look into fixing this ASAP? We can't merge PRs to Specimin while CI is failing, so if this can't be fixed before this PR is ready to merge we're going to disable the evaluation script again.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -309,6 +309,28 @@ public void setFieldNameToClassNameMap(
...
+  }
+
+  /**
+   * Get the list of synthetic classes created by UnsolvedSymbolVisitor.
+   *
+   * @return the value of missingClass.
+   */
+  public Set<UnsolvedClassOrInterface> getMissingClass() {
+    Set<UnsolvedClassOrInterface> copyOfMissingClass = new HashSet<>();
+    copyOfMissingClass.addAll(missingClass);
```
kelloggm, line 330: Can an `UnsolvedClassOrInterface` object be modified in a later run of `UnsolvedSymbolVisitor`? I think that the answer is yes: new methods can be added, the number of type variables might be updated, etc.

In that case, this copy operation isn't sufficient: you're creating a new set, yes, but the new set still contains the same set of pointers as the old set. So any change _to those objects_ will still be reflected in the new set.

(We may also need to implement a copy constructor for `UnsolvedClassOrInterface`.)

Note after I read the rest of the PR: my concern here is valid in general, but in this specific case the code is already correct, because this copy is transformed into Strings (which are immutable) immediately. However, that's not obvious from the way that the code is written, so I think we should make a change here to make it clear why this is safe despite `UnsolvedClassOrInterface` being mutable: in particular, I suggest that `getStringSetFromSyntheticClassSet` be called on `copyOfMissingClass` here before the result is returned (and the return type of this method (and its name, Javadoc, etc) be changed to `Set<String>`.

src/main/java/org/checkerframework/specimin/WorkDoneByUnsolvedSymbolVisitor.java
```diff
@@ -0,0 +1,56 @@
+package org.checkerframework.specimin;
+
+import java.util.Objects;
+import java.util.Set;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/** A simple class to keep track of the progress of UnsolvedSymbolVisitor. */
+public class WorkDoneByUnsolvedSymbolVisitor {
```
kelloggm, line 8: It is confusing that this class' name ends in `Visitor`, but it is not a visitor. Can we come up with a name that isn't so confusing? Here are some suggestions:
* `UnsolvedSymbolVisitorProgress`
* `UnsolvedSymbolVisitorProgressTracker`
* `UnsolvedSymbolVisitorWork`

kelloggm (changes requested):
Overall I like the implementation strategy in this PR: I think it's a big improvement over storing this information in another boolean, as you'd done in the previous draft.

tahiat:
Professor, let me add a new job for bugs and see if it goes through successfully.

LoiNguyenCS:
Professor, I have updated the PR based on your suggestions. Please take a look. Thank you.

kelloggm (approved):
Thanks, this looks great now :)

PR #160, opened 2024-02-29
Separating  evaluation targets in Specimin's CI

tahiat:


tahiat:
Professor, CI is only running on evaluation targets. This PR can be ignored also since it is not adding anything new.

kelloggm:
Let's close this for now and come back to it once it's fast enough to run with our CI resources.

PR #170, opened 2024-03-05
A more comprehensive check for original class files.

LoiNguyenCS:
Professor,

This is a better check for classes existing in the original codebase. This check can handle situations where a class such as `A` is not in `A.java`.

Additionally, I have updated the result of cf-3020. I have also taken the liberty to update the diff command of CI, since I have been struggled with the difference in the ending line of `target_status.json`.


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -174,8 +174,10 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
    */
   private Map<String, @ClassGetSimpleName String> fieldNameToClassNameMap = new HashMap<>();
 
-  /** The list of existing files in the input codebase. */
-  private Set<Path> setOfExistingFiles;
+  /**
+   * The set of Java classes in the original codebase mapped with their corresponding Java files.
```
kelloggm, line 178: This javadoc should indicate that the keys are fully-qualified names

kelloggm (approved):
LGTM, and I don't need to review again after my comment is addressed.

PR #171, opened 2024-03-05
Update README to ignore receiver parameters

LoiNguyenCS:
Professor,

This PR updates the README so that users don't add receiver parameters to signatures of target methods. Please take a look. Thank you.

Note: the CI is failing because we need to update the result of cf-3020.


kelloggm (approved):
I think it would appropriate to include the change to cf-3020's expected status in this PR. Please do that; once you do and CI is green, I don't need to review again.

PR #173, opened 2024-03-06
Ignore annotations in target method signatures

LoiNguyenCS:
This PR ignores annotations in target method signatures. Due to the difficulty in keeping space consistency after removing annotations, such as the transformation of `byte @Nullable []` to `byte[]` (removing both starting and ending spaces), versus `int x, @Nullable int b`, which becomes `int x, int b` (only removing the ending space), this PR removes all spaces from target method signatures prior to comparison.

### DIFF
--- a/src/test/resources/parameterwithannotations/input/org/testing/UnsolvedType.java
+++ b/src/test/resources/parameterwithannotations/input/org/testing/UnsolvedType.java
@@ -1,3 +1,0 @@
-package org.testing;
-public class UnsolvedType {
-}
--- a/src/test/resources/parameterwithannotations/input/org/testing/UnsolvedType.java
+++ b/src/test/resources/parameterwithannotations/input/org/testing/UnsolvedType.java
@@ -1,0 +1,3 @@
+package org.testing;
+public class UnsolvedType {
+}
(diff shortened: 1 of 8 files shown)


kelloggm:
A corresponding change in the target method for the `cf-3850` issue needs to be made in https://github.com/tahiat/specimin-evaluation. I've created a PR: https://github.com/tahiat/specimin-evaluation/pull/21

LoiNguyenCS:
Thank you very much for your suggestion. I have updated the test and added codes to make the new test work.

src/test/resources/parameterwithannotations/input/org/testing/UnsolvedType.java
```diff
@@ -0,0 +1,3 @@
+package org.testing;
```
kelloggm, line 1: Should this class have been added? The name suggests not.

(This is a tangent, but I've noticed that when Specimin crashes, there are often synthetic files left in the input folder. In a separate PR, it would be great if we could adjust Specimin so that it does all of its work in the output folder, instead, to avoid this problem. This isn't urgent, though, since it's only relevant when debugging a crash.)

kelloggm (changes requested):

LoiNguyenCS:
Thank you, I have removed the file.

kelloggm (approved):

PR #176, opened 2024-03-08
Allow methods in synthetic classes to take a lambda as a parameter

kelloggm:
This is a step towards solving CF-3850 (but is definitely not enough on its own)


kelloggm:
I'm looking into the CI failure

kelloggm:
@LoiNguyenCS this now handles arbitrary method call expressions in target methods at least one of whose parameters is a lambda (or at least, that's the intention). Handling lambdas is complicated because functions aren't first-class types in Java, and lambdas are _not subtypes_ of `java.lang.Object`. Instead, Java has one type hierarchy for functions (called "functional interfaces" in the JLS) for each possible "arity" of lambdas (including zero-arity for both the parameters and the return type):
* "producers" have 0-arity but do return something
* "consumers" can have any arity but don't return anything
* for each positive arity, there is a hierarchy of functional interfaces rooted a single-member interface with the arity+1 type variables; this interface _doesn't actually exist_ for all arities, which is why we have to create our own sometimes in this PR. For example, the 1-arity hierarchy (i.e., the functions that take one parameter and return a value) is rooted at `java.util.function.Function<?, ?>`

This design is, of course, insane. If we could just do something like `'a -> `b` (Ocaml syntax, sorry), this PR would be simple. Alas, Java.

LoiNguyenCS:
Thank you so much for this PR. It would take me forever to figure this one out myself. I am reviewing it. Handling functions in Java is quite new to me, so I am making sure that I understand everything in this PR correctly.

src/main/java/org/checkerframework/specimin/UnsolvedClassOrInterface.java
```diff
@@ -277,6 +277,11 @@ public String toString() {
     StringBuilder sb = new StringBuilder();
     sb.append("package ").append(packageName).append(";\n");
     if (isAnInterface) {
+      // For synthetic interfaces created for lambdas only.
+      if (methods.size() == 1
+          && (className.startsWith("Function") || className.startsWith("Consumer"))) {
```
LoiNguyenCS, line 282: Should we consider renaming these interfaces to names like `SyntheticFunction4` or `SyntheticConsumer3`? I think there could be more "real" interfaces starting with `Function` or `Consumer` compared to `SyntheticFunction` or `SyntheticConsumer`?

src/test/java/org/checkerframework/specimin/LambdaBiFunctionTest.java
```diff
@@ -0,0 +1,18 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.Test;
+
+/**
```
LoiNguyenCS, line 6: Is it intentional that all the tests here have the same Javadoc?

LoiNguyenCS (approved):
This PR looks great. Thank you so much. I only have some minor questions.

src/test/java/org/checkerframework/specimin/LambdaBiFunctionTest.java
kelloggm: I was being lazy :) thanks for calling me out on this, I'll differentiate them so that it's more clear what each test is supposed to cover.

src/main/java/org/checkerframework/specimin/UnsolvedClassOrInterface.java
kelloggm: Good idea!

PR #177, opened 2024-03-09
Handling target methods in non-primary classes

LoiNguyenCS:
Professor,

This PR addresses the weird scenario in Issue689. Here are my findings regarding non-primary classes:

1) Non-primary classes can only be accessed by classes within the same package. (While Oracle's documentation doesn't explicitly state this, it does say that only primary classes can be public)

2) However, if the primary class is declared as public, the non-primary class can only be accessed by classes within the same file. (I found out this behavior during experiments, and I have no ideas why)

Given the following examples:

In the first case, `printMessage()` can be accessed by other classes within the same package:

```java
package com.example;

class Baz {
    int toBeRemovedField;
    void toBeRemovedMethod() {
        System.out.println("Will be removed!");
    }
}

class NonPrimary {
    static void printMessage() {
        System.out.println("This method will be emptied");
    }
}
```

However, in the second case, `printMessage()` can only be accessed within the same Java file:

```java
package com.example;

public class Baz {
    int toBeRemovedField;
    void toBeRemovedMethod() {
        System.out.println("Will be removed!");
    }
}

class NonPrimary {
    static void printMessage() {
        System.out.println("This method will be emptied");
    }
}
```


kelloggm (approved):
Thanks for looking into this. This PR looks good to me, and it will be a solid basis to build on if we need to handle anything else related to non-primary classes.

PR #179, opened 2024-03-10
Handle type parameters with bounds to fix 6030b

LoiNguyenCS:
Professor,

This PR deals with type parameters having bounds. Please take a look. Thank you.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -1294,6 +1294,7 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
             boundOfType.get(index).resolve();
           } catch (UnsolvedSymbolException | UnsupportedOperationException e) {
             if (e instanceof UnsolvedSymbolException) {
+              this.gotException();
               // quoted from the documentation of Oracle: "A type variable with multiple bounds is a
               // subtype of all the types listed in the bound. If one of the bounds is a class, it
               // must be specified first."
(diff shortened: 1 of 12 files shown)


src/test/resources/multipleboundtypeparameter/input/com/example/Simple.java
```diff
@@ -0,0 +1,12 @@
+package com.example;
+
+import org.testing.UnsolvedType;
+import org.testing.UnsolvedTyp2;
```
kelloggm, line 4: typo in this test case (should be "UnsolvedType2")

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1275,6 +1277,39 @@ private boolean insideAnObjectCreation(MethodDeclaration decl) {
...
+          } catch (UnsolvedSymbolException | UnsupportedOperationException e) {
+            if (e instanceof UnsolvedSymbolException) {
+              // quoted from the documentation of Oracle: "A type variable with multiple bounds is a
+              // subtype of all the types listed in the bound. If one of the bounds is a class, it
+              // must be specified first."
+              // If the first bound is also unsolved, it is better to assume it to be a class.
+              if (index == 0) {
+                solveSymbolsForClassOrInterfaceType(boundOfType.get(index), false);
+              } else {
+                solveSymbolsForClassOrInterfaceType(boundOfType.get(index), true);
```
kelloggm, line 1304: Rather than an if-statement with two calls here, I would find this easier to read if you defined a boolean local variable for whether or not we're going to create an interface (based on the index).

kelloggm: Should there also be a call to `gotException()` here?

kelloggm (approved):
Overall looks good, thanks! After addressing my minor comments this can be merged without another review.

PR #181, opened 2024-03-10
A better way to recognize infinite loop

LoiNguyenCS:


LoiNguyenCS:
Not as good as I thought.

PR #183, opened 2024-03-11
checking minimized programs compile status

tahiat:
This PR adds a CI task that compiles all the minimized programs, creates a json file with the compilation status, and compares it with the current compilation status. If the comparison differs it fails the CI. 

We currently have 3 programs that pass compilation. As we fix compilation issue, that status should be change in `src/main/resources/min_program_compile_status.json` file. 

Dependent on [PR](https://github.com/tahiat/specimin-evaluation/pull/20)


.github/workflows/gradle.yml
```diff
@@ -43,3 +43,8 @@ jobs:
           python main.py
       - name: Check evaluation status
         run: diff -uw src/main/resources/target_status.json specimin-evaluation/ISSUES/target_status.json
+      - name: Check targets compile status
+        run: |
+          cd specimin-evaluation
+          sh check_compilation.sh
+          diff <(jq -S . src/main/resources/min_program_compile_status.json) <(jq -S . specimin-evaluation/ISSUES/compile_status.json)
```
kelloggm, line 50: Why is this diff command different from the one on line 45 above (beyond the names of the files, obviously)? I'd expect them to be the same.

tahiat: Professor, in line 45, for the execution of Specimin on input target, we parsed the json array and run minimization on them which maintained the order of items (order in which json array is constructed). 
Here, We are iterating over the directories which do not maintain the order (if a new directory is added, it's order can be in the middle of the existing directories.)

I am making a change to parse the json file and execute them in order.

kelloggm (approved):

PR #184, opened 2024-03-11
Correct types for superclass fields

kelloggm:
If we create synthetic types, then if there are constraints on RHSes of (pseudo-)assignments to the variable in the target method, those constraints won't be taken into account. Object is the top type, so all assignments to the superclass field will succeed if we type the superclass field at Object.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -132,22 +132,40 @@ class JavaTypeCorrect {
               new InputStreamReader(process.getInputStream(), StandardCharsets.UTF_8));
       String line;
       // incorrect constraint type will be updated here. Other types will be updated by
-      // updateTypeToChange
-      String incorrectConstraintType = "";
-      String correctConstraintType = "";
+      // updateTypeToChange. The following pairs of two-line errors are recognized:
+      // incompatible constraint types: "equality constraints", "lower bounds"
+      // bad operand types for binary operator: "first type", "second type"
+      String[] firstConstraints = {"equality constraints: ", "first type: "};
+      String[] secondConstraints = {"lower bounds: ", "second type: "};
+      String firstConstraintType = "";
       while ((line = reader.readLine()) != null) {
         if (line.contains("error: incompatible types")) {
           updateTypeToChange(line, filePath);
         }
-        // the type error with constraint types will be in a pair of lines:
-        // equality constraint: correctType
-        // lower bounds: incorrectType
-        if (line.contains("equality constraints: ")) {
-          correctConstraintType = line.trim().replace("equality constraints: ", "");
+        // these type error with constraint types will be in a pair of lines
+        for (String firstConstraint : firstConstraints) {
+          if (line.contains(firstConstraint)) {
+            firstConstraintType = line.replace(firstConstraint, "").trim();
+          }
         }
-        if (line.contains("lower bounds: ")) {
-          incorrectConstraintType = line.trim().replace("lower bounds: ", "");
-          typeToChange.put(incorrectConstraintType, correctConstraintType);
+        for (String secondConstraint : secondConstraints) {
+          if (line.contains(secondConstraint)) {
+            String secondConstraintType = line.replace(secondConstraint, "").trim();
+            if (isSynthetic(firstConstraintType)) {
+              typeToChange.put(firstConstraintType, secondConstraintType);
+            } else if (isSynthetic(secondConstraintType)) {
+              typeToChange.put(secondConstraintType, firstConstraintType);
+            } else {
+              throw new RuntimeException(
+                  "JavaTypeCorrect found two incompatible types, but neither is "
+                      + "synthetic:\n"
+                      + "first constraint type: "
+                      + firstConstraintType
+                      + "\nsecond constraint type: "
+                      + secondConstraintType);
+            }
+            firstConstraintType = "";
+          }
         }
       }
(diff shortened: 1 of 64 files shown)


LoiNguyenCS (commented):
I am not very sure about this PR. Specimin should be able to use javac to update the synthetic types as needed. Specifically, we have a test case called `syntheticsupervariables` to check if those synthetic fields can end up with correct types.

kelloggm:
> I am not very sure about this PR. Specimin should be able to use javac to update the synthetic types as needed. Specifically, we have a test case called `syntheticsupervariables` to check if those synthetic fields can end up with correct types.

That test case assumes that the correct types will appear in the input (and this PR does not cause it to fail). The issue that this PR is designed to solve is cases like the new test, where the only information about the field is in the RHS of assignment(s) to it:
```
class Simple extends Foo {
    void bar() {
        x = 10;
    } }
```

If this is the only mention of x, then before this PR Specimin would generate the following for the superclass:
```
class Foo {
    public SyntheticTypeForX x;
}
```

The issue is that javac would then issue the following error:
> .../Simple.java: 7: error: incompatible types: int cannot be converted to SyntheticTypeForX

JavaTypeCorrect can't fix this problem, because the error doesn't occur in the same file in which we need to make a change (the error is in Simple.java, but the fix needs to be in the synthetic superclass' Foo.java file). The ideal fix here might be to change the type of x in Foo to `int`, but there's no reason to write logic to do that when converting to `Object` is much easier and will always typecheck.

LoiNguyenCS:
Professor,

I want to test this new change a bit more on my local computer. Specifically, I don't understand why the changes in this PR do not cause the `syntheticSuperVariables` test to fail. If every super field has object type, we should see some changes in `syntheticSuperVariables`.

LoiNguyenCS:
> JavaTypeCorrect can't fix this problem, because the error doesn't occur in the same file in which we need to make a change (the error is in Simple.java, but the fix needs to be in the synthetic superclass' Foo.java file).

Actually, JavaTypeCorrect only returns the map of incorrect types to correct types. It is the job of UnsolvedSymbolVisitor to look up the list of synthetic types it has created and change those types accordingly.

The failure of your test case could be attributed to JavaTypeCorrect currently operating in a single direction. For instance, if an error message states:

> error: incompatible types: X cannot be converted to Y

JavaTypeCorrect interprets X as the incorrect synthetic type and Y as the correct type. To address this, JavaTypeCorrect should consider both directions when updating its type map. This adjustment should be ok, since synthetic types generated by UnsolvedSymbolVisitor won't share names with primitive types.

kelloggm:
> To address this, JavaTypeCorrect should consider both directions when updating its type map

I considered this fix for this test case. However, it wasn't clear to me how JavaTypeCorrect would know in which direction a particular fix should be applied, in the general case.

> This adjustment should be ok, since synthetic types generated by UnsolvedSymbolVisitor won't share names with primitive types

We also have to consider user-written types, not just primitives (though this test case only includes a primitive): the same problem would occur if the RHS is e.g. a constructor invocation.

I'm hesitant to assume that users will never write a type with a name like "SyntheticTypeFor...`, which seemed to be the only way to make this work in JavaTypeCorrect to me, especially with this simpler fix available.

> Specifically, I don't understand why the changes in this PR do not cause the syntheticSuperVariables test to fail. If every super field has object type, we should see some changes in syntheticSuperVariables

JavaTypeCorrect does correct these types, because they are actually declared in the target class - the super class fields are adjusted to match. I agree it's a bit inconsistent that the types are corrected in that case, but not in the case of superclass fields without shadowing. `syntheticSuperVariables` is a highly-unrealistic test case, in my opinion, because it does include shadowing and explicit `super` references, both of which are rare in real code.

LoiNguyenCS:
> I considered this fix for this test case. However, it wasn't clear to me how JavaTypeCorrect would know in which direction a particular fix should be applied, in the general case.

The whole logic of `JavaTypeCorrect` has been relying on the hope that users will not create type names like "SyntheticTypeFor..." :-)

So we can simply add both directions to the map.

If we have time, we can make the synthetic type names more unique by adding a random number to those syntheic type names.

kelloggm:
> Given this scenario, it is better to use synthetic names for types. This minimizes the likelihood of conflicts arising from multiple synthetic fields sharing the same type name.

This gets at the core problem with the current technical approach here, I think: JavaTypeCorrect's logic operates at the type level, which relies on each thing it might correct having a unique type. It would be better if it operated at the declaration level: that is, if the changes that it makes were scoped properly to the declarations that the errors are related to.

> Consequently, UnsolvedSymbolVisitor will proceed to indiscriminately change every synthetic field with an Object type into boolean

I see your point. I'll try to write a test case that exposes this behavior as part of this PR, and then I'll adjust the technical strategy here to fix this error at the JavaTypeCorrect stage. However, I think we're taking on even more technical debt here: JavaTypeCorrect's overall strategy of relying on type names is fundamentally incorrect, and it will continue to cause us problems in the future.

LoiNguyenCS:
> JavaTypeCorrect's overall strategy of relying on type names is fundamentally incorrect, and it will continue to cause us problems in the future.

I think the main difficulty we have here is that, in the case of Specimin, the file where the type error occurs (i.e. a target file) is never the same as the file where we need to fix that error (i.e. a synthetic file). Unfortunately, javac only tells us the simple names of types within error messages.

kelloggm:
@LoiNguyenCS can you take another look? I've completely changed the implementation strategy.

LoiNguyenCS (approved):
LGTM

PR #186, opened 2024-03-12
Only take anno from jar

LoiNguyenCS:
Professor,

This PR makes sure that we only preserve an annotation if it is solvable by the jar files or from java.lang.

The reason we have to use separate visit methods instead of `findAll()` is that `findAll()` misses some annotations when applying to big codebase. I could not identify the reason behind, so there is no test cases included.

Note: Specimin by default only preserves annotations inside target methods. This PR simply enforces that.


kelloggm (commented):
We should discuss this PR in the context of the email that you sent me with the subject "Questions about annotations". Annotations on used members shouldn't be removed, either.

LoiNguyenCS:
Dear Professor,

I've updated the PR with a new idea.

I realized that annotations usually go with specific parts of the code, like fields or classes. So we simply let the `PrunerVisitor` do its job by removing any unused code along with the annotations. What we need to do is make sure that all the unresolved annotations are removed before that. One downside is that we'll keep all the imports for resolved annotations, even if they're not used. But the good thing is, we won't need to create a whole new visitor just for handling annotations.

kelloggm:
> I realized that annotations usually go with specific parts of the code, like fields or classes.

Of course! This makes total sense, and I should have thought of it myself. In particular, I think we only care about annotations in the following locations (of both target and used members):
* parameters
* receivers
* type arguments
* return types
* method declarations
* field declarations (in either type or declaration position)
* class declaration (in either type or declaration position)

src/main/java/org/checkerframework/specimin/UnsolvedAnnotationRemoverVisitor.java
```diff
@@ -30,6 +34,24 @@ public class UnsolvedAnnotationRemoverVisitor extends ModifierVisitor<Void> {
    */
   Map<String, String> classToFullClassName = new HashMap<>();
 
+  /** The set of annotations predefined by java.lang. */
+  static final Set<String> javaLangPredefinedAnnotations =
```
kelloggm, line 38: This should be combined/added to the new `JavaLangUtils` class in #188

kelloggm (approved):
Overall this looks good to me. I'll let you decide how you want to combine the java.lang utilities that we each wrote; I'm not picky, but I think that code should all be in one place. If you'd prefer to merge this now and fix it later (in another PR after you merge #188), that's fine with me.

LoiNguyenCS:
> If you'd prefer to merge this now and fix it later (in another PR after you merge https://github.com/kelloggm/specimin/pull/188), that's fine with me.

I would prefer this approach.

PR #187, opened 2024-03-12
Add support for least upper bounds of types in Specimin's output

kelloggm:
This PR also contains three important supporting changes, which should _maybe_ be PR'd separately (but you might want to review them separately in the sense that you could check each independently when looking over this PR):
* ~~convert fields that get fully-qualified names in their synthetic types to end their type names with "SyntheticType" instead of just "Type". This facilitates their easy detection in JavaTypeCorrect later~~ review #189 before reviewing this PR. That PR contains just this precursor change.
* ~~refactor the way that exceptions and throwables are handled to be a generic "extends" mechanism. This would have worked on its own, and I may PR it separately so that it can be reviewed without the rest of this, if I can cleanly separate it.~~ review #188 before reviewing this PR
* added explicit handling of "case 3" in SpeciminRunner, and force UnsolvedSymbolVisitor to run at least once more every time that JavaTypeCorrect actually makes a change. This was necessary to handle the fairly-complex `Node` example in the new `RealSuperLub` test: without it, the last round of JavaTypeCorrect correctly makes a change even though UnsolvedSymbolVisitor didn't find any issues (to an extends clause), making the output non-compilable if UnsolvedSymbolVisitor doesn't run one more time.

The core idea of the handling for upper bounds is to keep track of any time that JavaTypeCorrect would like to correct a type to multiple other types, and then if so instead use the synthetic type as a mutual supertype. It only works for "2 types at a time", which is why you see e.g. the extra layer in the "real" example (`SytheticTypeForChild` above `Node`); I tried to generalize it further and broke everything, so I decided we're better off with the simpler (but not all that simple...) version here, which appears to work. (But, there's a good chance I missed something, and if you can think of some test cases that this might not handle I'd welcome them!)

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -9,7 +9,6 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -237,18 +236,8 @@ class JavaTypeCorrect {
         return;
       }
     }
-    typeToChange.put(incorrectType, correctType);
 
-    // keep the two output maps in-sync
-    Set<String> mustFix = new HashSet<>(1);
-    for (Map.Entry<String, String> extendedTypesEntry : extendedTypes.entrySet()) {
-      if (incorrectType.equals(extendedTypesEntry.getValue())) {
-        mustFix.add(extendedTypesEntry.getKey());
-      }
-    }
-    for (String mustFixKey : mustFix) {
-      extendedTypes.put(mustFixKey, correctType);
-    }
+    typeToChange.put(incorrectType, correctType);
   }
 
   /**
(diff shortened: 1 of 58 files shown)


src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
```diff
@@ -191,24 +196,66 @@ private void updateTypeToChange(String errorMessage, String filePath) {
...
+      // TODO: what error message triggers this code? Do we have test cases for it?
+      String rhs = splitErrorMessage.get(5);
+      String lhs = splitErrorMessage.get(splitErrorMessage.size() - 1);
+      if (isSynthetic(lhs)) {
+        changeType(lhs, tryResolveFullyQualifiedType(rhs, filePath));
+      } else if (isSynthetic(rhs)) {
+        changeType(rhs, tryResolveFullyQualifiedType(lhs, filePath));
+      } else {
+        extendedTypes.put(rhs, lhs);
+      }
```
kelloggm, line 226: @LoiNguyenCS I'm not sure if the parallel structure here should be the same as the version above it, or if this code should be left alone, because I'm not sure what it does.

LoiNguyenCS: Professor,

The error message that triggers this code is `error: incompatible types: found <type1> required <type2>`, which is case 2 according to comments at the beginning of this method. I am not sure which input will lead to this kind of error messages. But I think it should be the same as the version above it.

LoiNguyenCS:
Professor, the CI is failing because we have two additional projects that Specimin fail to minimize.

kelloggm:
> Professor, the CI is failing because we have two additional projects that Specimin fail to minimize.

I'm investigating.

LoiNguyenCS (approved):
This PR looks great to me.

PR #189, opened 2024-03-12
Make it obvious that fields are synthetic

kelloggm:
This is a precursor to #187 (and was necessary to get it working). Review before #187.


LoiNguyenCS (approved):
LGTM

PR #192, opened 2024-03-13
Preserve arguments to enum constructors called in enum constants

kelloggm:
This fixes one of the compilation bugs in CF-6077 (i.e., the integration test target derived from Apache Cassandra); there are others (including, I think, the one that #191 is intended to solve). This PR only handles enum constants; I don't expect CF-6077 to compile with just this PR.

We already are very approximate with enum constants (see #81 for example), so this change just acknowledges that. The implementation is a bit hacky, but it's cleaner than trying to add another boolean that we'd need to check in each other visit method.


LoiNguyenCS (approved):
LGTM

PR #193, opened 2024-03-13
Preserve overrides that are required for compilability

kelloggm:
This prevents warnings like this one, in the integration test cf-3022, when compiling Specimin's output:
```
./com/google/common/collect/AbstractMapBasedMultimap.java:114: error: AbstractMapBasedMultimap.WrappedCollection is not abstract and does not override abstract method size() in AbstractCollection
    class WrappedCollection extends AbstractCollection<V> {
    ^
```

The key problem is making sure that we preserve methods that are necessary because an abstract class/interface doesn't actually define an implementation for them.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
@@ -9,8 +9,10 @@ import com.github.javaparser.ast.expr.SuperExpr;
 import com.github.javaparser.ast.stmt.BlockStmt;
 import com.github.javaparser.ast.visitor.ModifierVisitor;
 import com.github.javaparser.ast.visitor.Visitable;
+import com.github.javaparser.resolution.UnsolvedSymbolException;
 import com.github.javaparser.resolution.declarations.ResolvedMethodDeclaration;
 import com.github.javaparser.resolution.declarations.ResolvedParameterDeclaration;
+import java.util.HashSet;
 import java.util.Set;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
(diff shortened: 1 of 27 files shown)


kelloggm:
I am looking into the CI failures

src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
```diff
@@ -0,0 +1,151 @@
...
+  @Override
+  public Visitable visit(MethodDeclaration method, Void p) {
+    ResolvedMethodDeclaration overridden = getOverriddenMethod(method);
+    // two cases: the method is a solvable override, and we can show that
+    // it is abstract (before the ||), or we can't solve the override but there
+    // is an @Override annotation. This relies on the use of @Override when
+    // implementing required methods from interfaces in the target code,
+    // but unfortunately I think it's the best that we can do here. (@Override
+    // is technically optional, but it is widely used.)
+    if (isAbstract(overridden) || (overridden == null && isOverride(method))) {
```
LoiNguyenCS, line 82: I am not sure what will happen if there are unsolved symbols in these methods. Technically, the current version of `UnsolvedSymbolVisitor` only solves symbols belonged to or used by target methods. If we preserve these methods, we could have a non-compiled output due to the unsolved symbols that they introduce.

LoiNguyenCS (approved):
Professor, I have a question about the safety of this approach.

src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
kelloggm: > If we preserve these methods, we could have a non-compiled output due to the unsolved symbols that they introduce.

The only things preserved about these methods are their signatures, so the concern here is that the parameter or return types will be unsolved. I'm not sure if this can occur in practice: we'll only need to preserve a method like this when dealing with a library that 1) we cannot control, like java.util and 2) an abstract class or interface from that library is implemented by a used class. I think 1) especially is important: for most libraries, we would remove the interface if it wasn't required, or _remove the method from the interface_ if it wasn't used (because the interface would be a synthetic method under our control).

Regardless, this is an improvement on the current situation: at present, in this situation Specimin's output already doesn't compile (because the compiler complains about an unimplemented method).

kelloggm: As a follow up to my comment above, this problem does appear to be what's causing cf-6030 to fail now: `MustImplementMethodVisitor` is throwing an `UnsolvedSymbolException` when trying to get a return type. I'm going to make this more conservative, and only preserve these methods if their return type and parameter types are solvable, for now.

kelloggm:
Oof, this is still incorrect: CI is now failing because this PR made compilability worse. I'm looking into it, but may not finish it tonight.

LoiNguyenCS:
I think it happens because there are some @Override annotations for methods that we can not solve overriding. Removing all @Override annotations would be a good solution, I believe. I can make a quick PR to do that if you want me to.

kelloggm:
> I think it happens because there are some @Override annotations for methods that we can not solve overriding

I think that will just hide the problem. I fixed the specific issue with compilability in this PR (my first attempt didn't consider whether the overridden method was preserved in the super class), so I expect this PR to pass CI now.

LoiNguyenCS:
Professor, the CI is failing because of many unresolved `@Override` methods.

I guess the reason lies in `getOverriddenMethod`. Maybe this method can only solve inheritance for classes, because it relies on a synthetic super call? This approach should not work for method signatures from interfaces.

kelloggm:
@LoiNguyenCS this is actually ready to be reviewed again. I've given in and taken your suggestion to never output `@Override`.

LoiNguyenCS:
This PR looks great. Thank you very much.

PR #194, opened 2024-03-13
Handle cases when the target class extends an inner class.

LoiNguyenCS:
Professor,

This PR uses the update method from `TargetMethodFinderVisitor` to update the list of extended classes more comprehensively. Please take a look. Thank you.


LoiNguyenCS:
Professor, please help me review #195 instead. Thank you.

PR #195, opened 2024-03-14
Elaborate inheritance visitor

LoiNguyenCS:
This PR updates the inheritance visitor to take inner class and type parameters into consideration. Please take a look. Thank you.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/InheritancePreserveVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/InheritancePreserveVisitor.java
@@ -2,6 +2,7 @@ package org.checkerframework.specimin;
 
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
+import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.visitor.ModifierVisitor;
 import com.github.javaparser.ast.visitor.Visitable;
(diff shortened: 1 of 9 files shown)


src/main/java/org/checkerframework/specimin/InheritancePreserveVisitor.java
```diff
@@ -49,17 +50,47 @@ public void emptyAddedClasses() {
...
           continue;
         }
-        addedClasses.add(qualifiedName);
       }
     }
     return super.visit(decl, p);
   }
+
+  /**
+   * (This method is copied from TargetMethodFinderVisitor).
```
kelloggm, line 71: Rather than copying, we should create a shared static utility method.

kelloggm: (for now, it could be in either visitor, though at some point we'll want to consolidate these kind of utility methods in one place)

kelloggm (changes requested):
LGTM other than the copying.

LoiNguyenCS:
I am trying to address an infinite loop problem.

LoiNguyenCS:
Professor,

I need to do some refactoring to make the update method static and publicly accessible. Please help me review again. Thank you.

LoiNguyenCS:
I think the new updates in main have caused one of the test case to fail. I am resolving it.

LoiNguyenCS:
Professor @kelloggm, this PR is now ready for review.

kelloggm (approved):
LGTM

PR #196, opened 2024-03-14
Consider type variable when updating used classes

LoiNguyenCS:
Professor,

This PR handles the bug where a type variable is ignored when updating used classes, leading to uncompiled output. Please take a look. Thank you.


kelloggm (approved):
LGTM

PR #198, opened 2024-03-14
Fix how static imports are preserved

kelloggm:
I'm hopeful that this is the last thing necessary for cf-6388 to pass.


LoiNguyenCS (approved):
LGTM.

PR #201, opened 2024-03-15
Remove unused enum constants

LoiNguyenCS:
Professor,

This PR makes sure that Specimin will not updates the list of used classes based on unused enums. It also makes sure that only used enums are preserved in the final output. 


src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -248,6 +258,22 @@ public Visitable visit(ConstructorDeclaration method, Void p) {
...
+    if (method.getParentNode().isEmpty()) {
+      return result;
+    }
+    if (method.getParentNode().get() instanceof EnumDeclaration) {
+      EnumDeclaration parentNode = (EnumDeclaration) method.getParentNode().get();
+      if (parentNode.getFullyQualifiedName().isEmpty()) {
+        return result;
+      }
+      // used enums needs to have compilable constructors.
+      if (usedEnum.contains(parentNode.getFullyQualifiedName().get())) {
```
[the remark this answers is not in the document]
kelloggm: In this case, though, you know that the name of interest already belongs to an enum, so there's no danger of getting it mixed up with a class.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -76,6 +77,9 @@ public class TargetMethodFinderVisitor extends ModifierVisitor<Void> {
    */
   private Set<String> usedClass = new HashSet<>();
 
+  /** Enums that were actually used by the targets. These classes will be included in the input. */
+  private Set<String> usedEnum = new HashSet<>();
```
kelloggm, line 81: Why do we need a separate set for used enums? I think it would work correctly if they were stored in the same set as used classes: that is, if they were added to `usedClass` instead of kept as a separate field.

(See my other comment below, as well, for why I think this is safe.)

kelloggm (approved):
Overall this is a great PR! I have one question about a possible refactoring to simplify it a bit, but it could also be merged as-is if I'm wrong about whether the simplification that I propose is safe.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
LoiNguyenCS: Professor, it's mostly because JavaParser does not consider an enum a class. So I thought we should also separate them.

kelloggm: I disagree with JavaParser on this: under the hood, enums are compiled to classes, and they share the same namespace. Let's simplify our own code and just have one set of used class-like structures.

On a related note, we should settle on a term for "class-like structures" and use it throughout the code. `usedClass` isn't a great name anymore for that field - it's more like "used types" or "used concrete type declarations". Within the checker framework, we'd usually refer to such a thing as a `TypeElement`, following javac's internal convention.

LoiNguyenCS: I'm a bit confused. If we consider an enum to be a class, then why `usedClass` is not a good name anymore?

kelloggm: Because `usedClass` also contains interfaces (which are not classes, but are types) and enums (which are a special kind of class-like thing that javac will compile into a class). But this is a minor detail and shouldn't block this PR from being merged once we've applied the suggested refactoring.

LoiNguyenCS:
Professor, I have refactored Specimin to use a new field name. Please take a look. Thank you.

kelloggm (approved):
Thanks, LGTM

PR #202, opened 2024-03-15
Preserve used method references

kelloggm:
I noticed this omission while looking into cf-3850, but I don't think it's the proximate cause of that crash.


LoiNguyenCS (approved):
LGTM

PR #203, opened 2024-03-17
Handle unsolved parameters with array types.

LoiNguyenCS:
Professor,

I figured out this issue while trying to fix a crashing project. Please take a look. Thank you.


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1895,6 +1895,14 @@ public List<String> getArgumentTypesFromMethodCall(
         parametersList.add(((ResolvedReferenceType) type).getQualifiedName());
       } else if (type.isPrimitive()) {
         parametersList.add(type.describe());
+      } else if (type.isArray()) {
+        // the type without the [].
+        ResolvedType componentType = type.asArrayType().getComponentType();
```
kelloggm, line 1900: What if the component type is also an array? Can you add a test case with a 2d array to check that this will work?

kelloggm (changes requested):

LoiNguyenCS:
I have updated this PR with the new test case and codes. Please review again. Thank you.

kelloggm (approved):

PR #204, opened 2024-03-17
Handle stuck method calls.

LoiNguyenCS:
Professor,

This PR was supposed to fix `cf-3021`. While it appears to fix `cf-3850`, it actually does not. The solution proposed here only holds true if `UnsolvedSymbolVisitor` successfully fix unresolved symbols, which is not the case for `cf-3850`.

This solution is not very elegant. But since it does not preserve any additional unresolved symbols in the final output, it will not introduce any adverse effect on the behavior of Specimin. 


kelloggm (approved):
This is a nice solution to this problem! LGTM.

PR #205, opened 2024-03-17
Preserve wildcard imports

LoiNguyenCS:
This PR fixes some compilation errors in `cf-6030`. The problem occurs because not all classes have access to every class in the list of used classes. So, it's risky to transform a wildcard import based on that list. Unfortunately, I couldn't find a good way to write a test case to show this bug.


kelloggm (approved):
Disappointing that we need to remove this, but this is a bit of a "vanity feature" anyway (vanity feature = it's nice to have, but not critical). So, it's okay to remove it.

PR #208, opened 2024-03-19
Preserve methods that match unsolvable methods used in lambda expression bodies

kelloggm:
This doesn't appear to actually fix cf-3850, despite the fact that the test is nearly identical to cf-3850's target method. In particular, running cf-3850 locally still fails to solve the call to `castNonNull`, which I think is unrelated to the problem solved by this PR. I'll look into that separately.


LoiNguyenCS (approved):
LGTM

PR #209, opened 2024-03-19
Handle return types mismatching

LoiNguyenCS:
Professor,

This PR handles the mismatching in return types that we discussed yesterday. Please take a look. Thank you.


kelloggm (approved):
LGTM, thanks

PR #210, opened 2024-03-19
Type var consistency

LoiNguyenCS:
Professor,

This PR introduces a new field in `UnsolvedClassOrInterfaceType` to make sure that we don't have compilation error due to mismatches in type variables. 

In the provided test case, if we have `SimpleParent<T, T1>`, there arises a potential issue when the javac updates the synthetic type of `get()` to `V`. This update would result in a type variable mismatch within the `SimpleParent` class.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedClassOrInterface.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedClassOrInterface.java
@@ -39,8 +39,8 @@ public class UnsolvedClassOrInterface {
   /** This field records the number of type variables for this class */
   private int numberOfTypeVariables = 0;
 
-  /** This field records the name of type variables for this class as it is used */
-  private Set<String> preferedTypeVariables = new HashSet<>();
+  /** This field records the name of type variables that we prefer this class to have. */
+  private Set<String> preferredTypeVariables = new HashSet<>();
 
   /** The field records the extends/implements clauses, if one exists. */
   private @Nullable String extendsClause;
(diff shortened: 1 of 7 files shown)


src/main/java/org/checkerframework/specimin/UnsolvedClassOrInterface.java
```diff
@@ -354,9 +366,15 @@ public String getTypeVariablesAsStringWithoutBrackets() {
    * @param result a string builder. Will be side-effected.
    */
   private void getTypeVariablesImpl(StringBuilder result) {
-    for (int i = 0; i < numberOfTypeVariables; i++) {
-      String typeExpression = "T" + ((i > 0) ? i : "");
-      result.append(typeExpression).append(", ");
+    if (preferredTypeVariables.size() == 0) {
```
kelloggm, line 369: Are we certain it's never possible for `preferredTypeVariables` to be a subset of the actual type variables of the class? For example, what if `preferredTypeVariables` has size 1, but `numberOfTypeVariables` is 2? Right now, the code will only add the elements of `preferredTypeVariables`.

At a minimum, such a situation here should raise an exception, so that we know if it occurs in any of our large integration tests when this PR goes through CI.

kelloggm (changes requested):
Overall this is a nice PR, but I have one major concern.

LoiNguyenCS:
Professor, I have upgraded the codes to throw exception.

kelloggm (approved):

PR #211, opened 2024-03-19
Make CF-3850 compilable

kelloggm:
This PR contains three related fixes, all of which are necessary to get CF-3850 (and therefore the new test case in this PR) to compile:
* we need to add something to the list of arguments for a synthetic method as a placeholder if `null` is passed. This PR just adds "Object" there.
* the strategy for handling constraints in JavaTypeCorrect has changed to handle constraints with multiple types in them (they're comma-separated). This is necessary to trigger the appropriate type correction so that the return type of `castNonNull` is compatible with the lambda (there are two lower bounds).
* `updateTypeForSyntheticClasses` has been changed to attempt to use a fully-qualified name when possible, because otherwise there is an unsolved symbol once we correctly mark the return type of `castNonNull` as `SqlParserPos` (a simple name from javac's output) in the test case.

This PR is currently a draft because it passes the test cases locally but doesn't actually fix CF-3850's last remaining error. I'm investigating why it fails now.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaLangUtils.java
+++ b/src/main/java/org/checkerframework/specimin/JavaLangUtils.java
@@ -7,13 +7,25 @@ import java.util.Set;
 public final class JavaLangUtils {
 
   /**
-   * Checks if the given simple name is a member of the java.lang package.
+   * Checks if the given simple name is a member of the java.lang package. This method returns false
+   * for primitive names (e.g., "int").
    *
    * @param simpleName a simple name
    * @return true if this name is defined by java.lang
    */
   public static boolean isJavaLangName(String simpleName) {
     return javaLangClassesAndInterfaces.contains(simpleName);
+  }
+
+  /**
+   * Checks if the given simple name is a member of the java.lang package. This method also returns
+   * true for primitive types (like "int").
+   *
+   * @param simpleName a simple name
+   * @return true if the name is defined by java.lang or is a primitive
+   */
+  public static boolean isJavaLangOrPrimitiveName(String simpleName) {
+    return primitives.contains(simpleName) || javaLangClassesAndInterfaces.contains(simpleName);
   }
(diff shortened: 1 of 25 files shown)


LoiNguyenCS:
> Fundamentally, I think the core problem is that our strategy for lubbing only works for types in positions where the correct thing to do with two types is LUB them - that is, locations like fields or parameters. For return types, we actually need GLBs, not LUBs - in other words, the right thing to do is to add a generic type that's unconstrained.

Professor, can we rely on the fact that synthetic return types always end with `ReturnType` in their names?

LoiNguyenCS:
Ah sorry, there are two cases where the synthetic return types do not end with `ReturnType`, but those cases should not matter in this issue. The first case is a constructor method. The second case is for a super call. For example, if there is `super.get()`, Specimin will look up to see if there is a `get()` method in the current class and set the return type of that method in the parent class accordingly. However, I don't think we have to do any type correcting for these two cases.

kelloggm:
> Professor, can we rely on the fact that synthetic return types always end with ReturnType in their names?

I've implemented this. I can pass the test case in this PR using this strategy (nice!), but CF-3850 seems to cause specimin to crash in my local testing. I'll keep investigating and this PR can wait, for now.

LoiNguyenCS:
> but CF-3850 seems to cause specimin to crash in my local testing.

Good news is that it is shown as passed in this CI>

kelloggm:
> Good news is that it is shown as passed in this CI

It does! In fact, the problem was that my local copy had an extra file that was created by specimin itself during a previous run (that crashed). After cleaning that git repository it works. So, I've marked this PR as ready for review.

It shouldn't be reviewed until after #213 is merged, though, since it also includes those changes.

LoiNguyenCS (approved):
This is a great PR. Thank you.

PR #213, opened 2024-03-20
Correctly preserve methods with a null argument

kelloggm:
This PR also includes a significant refactoring to unite the code that's used for inferring parameter types from arguments. Previously, we had three copies of it (only one of which has been modified by @LoiNguyenCS's recent PRs) - one for method call expressions, another for object creation expressions, and a third for explicit constructor invocations. Now all these locations call into the same routine (derived from the method call code), so fixes in one location will apply to all three.


kelloggm:
Note: I did not intentionally fix cf-691, but apparently the changes in this PR do that, too :)

LoiNguyenCS:
This PR is great. Is it ready for review?

kelloggm:
> This PR is great. Is it ready for review?

Yes, definitely! Sorry, I created it as a draft PR by accident.

LoiNguyenCS (approved):
LGTM

PR #214, opened 2024-03-20
Use parameters' types instead of names

LoiNguyenCS:


kelloggm (approved):
LGTM. It wouldn't be correct to use `getArgumentTypesImpl` anyway because that method handles _arguments_ rather than _parameters_: at a method declaration, we should be able to find all of the parameter types directly by just inspecting the declaration, so there's no need to look up the types of each actual argument and infer the parameter types from there.

In other words, parameters are LHSes and arguments are RHSes in the pseudo-assignment that is a method call.

PR #215, opened 2024-03-20
Use Error instead

LoiNguyenCS:


src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
```diff
@@ -220,7 +220,10 @@ private void updateTypeToChange(String errorMessage, String filePath) {
       String rhs = splitErrorMessage.get(4);
       String lhs = splitErrorMessage.get(splitErrorMessage.size() - 1);
       if (lhs.equals("Throwable")) {
-        extendedTypes.put(rhs, "Throwable");
+        // Although not technically precise, using "Error" is safe in all cases. Extending Throwable
+        // alone might result in compilation errors if there's no corresponding catch in the source
+        // codes.
+        extendedTypes.put(rhs, "Error");
```
kelloggm, line 226: This change does not seem correct to me, since `Throwable` is the top type in the exception hierarchy (that is, `Error` is a subtype of `Throwable`. This change will cause issues in other code that expects a `Throwable` but not an `Error` (e.g., where an `Exception` is expected instead by the `catch`).

I think we need to re-evaluate how to fix this problem, since this specific fix is too coarse. Do you have a test case demonstrating the underlying problem in cf-689?

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
LoiNguyenCS: Professor, I have added a test case to illustrate how we will get a compilation problem if the synthetic exception only extend Throwable.

Here is the error that we will get if we try to compile the output where Throwable is used:

> @LoiNguyenCS ➜ /workspaces/specimin/tempDir (throw-runtime-exception) $ javac com/example/*.java
> com/example/Simple.java:9: error: unreported exception MyException; must be caught or declared to be thrown
>             throw new MyException("A blank String is no good");
>             ^
> 1 error

kelloggm: I see the problem. javac assumes that a `Throwable` is a checked exception; `Error`s are exempt from that requirement. I see your point that it would be better to avoid creating a checked exception (at least in the example in your new test case, where there is no `throws` clause). Let me propose two other scenarios, though, that I think will require different handling:

### Scenario A (with a throws clause):
```
package com.example;

import org.testing.MyException;

class Simple {
    // Target method.
    void bar(String input) throws MyException {
        if (input.isBlank()) {
            throw new MyException("A blank String is no good");
        }
    }
}
```

In this case, I'd argue that `Throwable` (or maybe `Exception`? The actual `MyException` must extend `Exception` and not `Error`, because of the throws clause) is the right class to extend. `Error` is certainly wrong, because the exception is declared as checked, but exceptions that extend `Error` are not checked.

### Scenario B (catch clauses):
```
package com.example;

import org.testing.MyException;

class Simple {
    // Target method.
    void bar(String input) {
        try {
              foo();
         catch(Exception me) {
              // doesn't matter
         }
    }

    void foo() throws MyException { }

}
```

In this case, whether or not the catch clause is a feasible path depends on whether the class that we construct for `MyException` extends `Exception` or `Error`. If it extends `Error`, then the path is infeasible (since `Error` is not a subtype of `Exception`). Whether a catch clause is feasible is within the realm of what a modular analysis is allowed to reason about (e.g., the Resource Leak Checker does this kind of analysis internally, to some extent); more broadly, this is a question of whether exceptions are also subject to the same rules about the type hierarchy that other classes are, and I'd say that they are.

LoiNguyenCS: What if instead of `Error`, we use `RuntimeException`? I think that will fit nicely to both scenarios. And `RuntimeException` does not require a corresponding check, too. 

Here is what I have read from the documentation of Oracle:

> Runtime exceptions can occur anywhere in a program, and in a typical one they can be very numerous. Having to add runtime exceptions in every method declaration would reduce a program's clarity. Thus, the compiler does not require that you catch or specify runtime exceptions (although you can).

kelloggm: I think this would handle scenario B, but not scenario A: fundamentally, I think we need a smarter approach there to make sure that we preserved the checked-ness of the exceptions.

LoiNguyenCS: So I believe that we are not concerning about compilation here but rather the behavior of a program analysis tool regarding the correctness of check-ness in the minimized output?

kelloggm: We need to be concerned with both, but yes that's my concern. Checked-ness of exceptions is something that a modular analysis tool is allowed to reason about, so we need to preserve it.

LoiNguyenCS:
Professor,

I've implemented some updates to Specimin, adopting a new approach. The core idea is to let `UnsolvedSymbolVisitor` handle all checked exceptions prior to reaching `JavaTypeCorrect`. This ensures that any remaining exceptions by the time we reach `JavaTypeCorrect` are indeed unchecked. Additionally, this approach addresses the imprecision noted in `ThrowableImprecisionTest`, which is a nice improvement.

As I understand [Oracle's tutorial](https://docs.oracle.com/javase/tutorial/essential/exceptions/declaring.html), method declarations are the sole locations where exceptions should be specified. So I've written the code accordingly.

Please let me know your opinions. Thank you.

Note: This approach can't handle cases where type inference is required, such as this one:
```java
try {
 if (x=1) {
   throw new NotTwoException();
 }
}
catch (Exception e) {

} 
```
`NotTwoException` will still be considered a `RuntimeException` by this approach, since it is neither in a catch clause nor specified in a method's declaration. However, I think this approach is still an improvement compared to the old approach.

kelloggm (approved):
This implementation is much improved, thanks!

> Note: This approach can't handle cases where type inference is required, such as this one:

Can you add this example as a test case (with the fact that the expected output is wrong noted)? I'd prefer that we write down that this behavior isn't ideal, even if we're not going to fix it ASAP. Once you do, I don't need to review again.

PR #216, opened 2024-03-21
Remove completely unresolvable interfaces

LoiNguyenCS:
Professor,

This PR removes interfaces that cannot be resolved for successful compilation  These are interfaces from the java packages. This issue only happens if we isolate the input file from its codebase. Please take a look. Thank you.

### DIFF
--- a/src/test/resources/unsolvableinterface/input/com/example/Foo.java
+++ b/src/test/resources/unsolvableinterface/input/com/example/Foo.java
@@ -1,0 +1,13 @@
+package com.example;
+
+import java.util.List;
+
+// Suppose this is an isolated Java program from a real life project, and all of the implementation
+// for methods from List are in Baz, then Specimin should simply remove List<E> from the class
+// declaration.
+class Foo<E> extends Baz<E> implements List<E> {
+    public void bar() {
+        System.out.println("Foo is doing something!");
+    }
+
+}
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -53,7 +53,7 @@ class JavaTypeCorrect {
   private Map<String, String> extendedTypes = new HashMap<>();
 
   /**
-   * /** This map associates the name of a class with the name of the unresolved interface due to
+   * This map associates the name of a class with the name of the unresolved interface due to
    * missing method implementations.
    */
   private Map<String, String> classAndUnresolvedInterface = new HashMap<>();
@@ -166,7 +166,7 @@ class JavaTypeCorrect {
         // source codes to begin with. This usually happens when a file is isolated from its
         // package, and its parent is supposed to override some of the methods in the given
         // interface. For these cases, if the interface is not from Java language, we will modify
-        // the codes of the interface. Otherwise, we will remove that interface completely..
+        // the codes of the interface. Otherwise, we will remove that interface completely.
         if (line.contains("not abstract and does not override abstract method")) {
           updateClassAndUnresolvedInterface(line);
         }
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -167,6 +167,10 @@ class JavaTypeCorrect {
         // package, and its parent is supposed to override some of the methods in the given
         // interface. For these cases, if the interface is not from Java language, we will modify
         // the codes of the interface. Otherwise, we will remove that interface completely.
+
+        // TODO: Update Specimin to generate a synthetic version for the missing parent class with
+        // synthetic method implementations, particularly if the targeted method invokes a method
+        // from the parent class that implements a method from a Java language interface.
         if (line.contains("not abstract and does not override abstract method")) {
           updateClassAndUnresolvedInterface(line);
         }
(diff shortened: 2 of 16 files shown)


src/test/resources/unsolvableinterface/input/com/example/Foo.java
```diff
@@ -0,0 +1,13 @@
+package com.example;
+
+import java.util.List;
+
+// Suppose this is an isolated Java program from a real life project, and all of the implementation
+// for methods from List are in Baz, then Specimin should simply remove List<E> from the class
+// declaration.
```
kelloggm, line 7: I am unconvinced that this is the right thing to do in all situations. In particular, I'm concerned about a similar case where the target method uses some part of the interface - e.g., what if `bar()` in this example includes a call to `this.add()` or something like that.

The natural thing that I would expect Specimin to do in that situation is to create the required stubs for List's various methods in the super class (i.e., `Baz`) and keep the interface. I'm concerned that this change will cause that not to happen properly. Can you add a test case to demonstrate that that doesn't happen?

kelloggm (changes requested):

src/test/resources/unsolvableinterface/input/com/example/Foo.java
LoiNguyenCS: > Can you add a test case to demonstrate that that doesn't happen?

I think it depends on whether `Baz.java` is available in the input. I will add the test cases.

LoiNguyenCS: I have added two tests to show how Specimin will behave depends on whether `Baz.java` is available in the codebase. Please let me know if the expected results for these tests are in accord with your expectations.

LoiNguyenCS: Sorry for the names of these two tests. My English got stuck.

kelloggm: I don't think the current behavior is correct. In particular, I don't think it should matter whether `Baz.java` is present in the original repository: the output of the `unsolvableinterfaceprettyhard` test is correct for both cases.

The reason is that the signature of `add` is known, even in the other case: it's the one from `List`. To retain not just compilability but the same behavior from analyses, we need to correctly have that method call be defined by the interface, not just in an arbitrary superclass.

LoiNguyenCS: I believe you are suggesting that if `Foo` extends `Baz` and implements `List`, and there is missing method implementations in `Foo` and no `Baz.java`, then we should be able to create a synthetic version for `Baz.java` that matches `List`?

kelloggm: > there is missing method implementations in Foo and no Baz.java, then we should be able to create a synthetic version for Baz.java that matches List

Yes, if and only if the missing method is compatible with something in List. So for example if the target method calls "bananas()", it's okay to eliminate List and put `bananas()` in `Baz`. But if the target method calls `add(E)`, then we need to preserve List in Foo's implemented interfaces and then make Baz actually implement List, because `List#add`'s signature (which must be realized in Baz or some of its superclasses) was used.

LoiNguyenCS: I see, thank you for the clarification. I will try to update the codes to work that way.

LoiNguyenCS: Professor,

I've run into a problem with this task. JavaParser isn't making it easy to find missing methods, so we have to use javac instead. But the issue with javac is that it only tells us about one missing method at a time. So, if there are 10 missing methods, we'd have to run javac 10 times to find them all.

LoiNguyenCS: One potential solution I could think of is to raise an exception or issue a warning, telling users to input the entire codebase into Specimin instead of an isolated file.

kelloggm: > JavaParser isn't making it easy to find missing methods, so we have to use javac instead

We shouldn't use javac to do this: it can be done at the JavaParser level, or we shouldn't try to do it at all.

Have you looked at the code in `MustImplementMethodsVisitor.java#overridesAnInterfaceMethodImpl`? Or considered just calling that method directly? I wrote that code specifically to solve this problem.

LoiNguyenCS: > Have you looked at the code in MustImplementMethodsVisitor.java#overridesAnInterfaceMethodImpl? Or considered just calling that method directly? I wrote that code specifically to solve this problem.

Thank you for the suggestion. I haven't considered that method. I will try using it and let you know.

LoiNguyenCS: Professor,

This issue is still difficult to handle.
 
`MustImplementMethodsVisitor.java#overridesAnInterfaceMethodImpl`  identifies **existing** methods that must be retained for compilation, our current issue revolves around determining the necessity of preserving **non-existing** methods. Addressing this requires traversing all the inheritance trees and tracking implemented methods. This task turns out to be quite complicated to accomplish.

LoiNguyenCS:
Professor,

Following our discussion during the meeting, I have added a comment so that future maintainers are aware of the current limitation of this approach. Please take a look. Thank you.

kelloggm (approved):
LGTM

PR #218, opened 2024-03-26
Avoid unsolved symbols in functional interface methods

LoiNguyenCS:
Professor,

This PR makes sure that Specimin empties out the content of all non-target methods, even the methods inside a functional interface. Please take a look. Thank you.


kelloggm (approved):
LGTM

PR #219, opened 2024-03-27
Remove FunctionalInterface anotations

LoiNguyenCS:
Professor,

This is a simple PR to remove all FunctionalInterface annotations. Please take a look. Thank you.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
@@ -34,6 +34,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Optional;
 import java.util.Set;
+import javax.annotation.Nullable;
 import org.checkerframework.checker.signature.qual.FullyQualifiedName;
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/java/org/checkerframework/specimin/PrunerVisitor.java
```diff
@@ -34,6 +34,7 @@
 import java.util.Iterator;
 import java.util.Optional;
 import java.util.Set;
+import javax.annotation.Nullable;
```
kelloggm, line 37: Please use `org.checkerframework.checker.nullness.qual`'s `@Nullable` annotation instead of the deprecated `javax.annotation.Nullable`.

kelloggm (changes requested):
This PR should also remove the code in `UnsolvedClassOrInterface` around line 310 that adds `@FunctionalInterface` annotations.

LoiNguyenCS:
Professor,

> This PR should also remove the code in UnsolvedClassOrInterface around line 310 that adds @FunctionalInterface annotations.

Since `PrunerVisitor` still relies on the @FunctionalInterface annotation to know whether to preserve an abstract method declaration that is not used by any target method, I think we should keep the codes that add @FunctionalInterface annotations in `UnsolvedClassOrInterface`.

kelloggm (approved):

PR #220, opened 2024-03-28
Preserve constructors to solve compilation errors

LoiNguyenCS:
Professor,

This PR fixes the compilation error in cf-3022.

I think my comments and Javadoc are not clear. So I would greatly appreciate your suggestions.

Given a class with a constructor utilized by target methods (let's call it constructor _A_), if constructors from the super-classes of that class are also used by target methods, we need to ensure the preservation of the `super(...)` call within _A_. However, implementing this preservation strategy can be quite challenging. Therefore, I've opted to maintain the entirety of _A_'s body.

To illustrate this point, consider the following scenario with the test case `constructorrequirement`: If we were to empty out the constructor of `BazChild`, we would encounter a compilation error similar to the one below:

```
@LoiNguyenCS ➜ /workspaces/specimin/tempDirOutput (fix-3022) $ javac com/example/*.java
com/example/Foo.java:5: error: constructor Baz in class Baz cannot be applied to given types;
    BazChild(String args) {
                          ^
  required: String
  found: no arguments
  reason: actual and formal argument lists differ in length
1 error
```

I would greatly appreciate your feedback on this approach. Please let me know your thoughts. Thank you.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
@@ -40,6 +40,7 @@ import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.checkerframework.checker.signature.qual.FullyQualifiedName;
 
 /**
@@ -187,10 +188,17 @@ public class PrunerVisitor extends ModifierVisitor<Void> {
   @Override
   public Visitable visit(ClassOrInterfaceDeclaration decl, Void p) {
     boolean oldInsideFunctionalInterface = insideFunctionalInterface;
+    @Nullable AnnotationExpr functionInterfaceAnnotationExpr = null;
     for (AnnotationExpr anno : decl.getAnnotations()) {
       if (anno.getNameAsString().equals("FunctionalInterface")) {
         insideFunctionalInterface = true;
+        functionInterfaceAnnotationExpr = anno;
       }
+    }
+    if (functionInterfaceAnnotationExpr != null) {
+      // @FunctionalInterface is optional, so we will remove it to avoid possible compilation
+      // errors.
+      functionInterfaceAnnotationExpr.remove();
     }
     decl = minimizeTypeParameters(decl);
     if (!classesUsedByTargetMethods.contains(decl.resolve().getQualifiedName())) {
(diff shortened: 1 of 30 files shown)


kelloggm (changes requested):
I've added two new tests. One of these tests passes (`constructorchaining`); it comes from an attempt by me to fix this same bug a few weeks ago. The other test (`constructorchaining2`) fails. In particular, it fails because the rest of preserved constructors should not be preserved - the `super` or `this` call must be, but the rest of the content can be deleted or replaced with `throw new Error()`.

LoiNguyenCS:
> In particular, it fails because the rest of preserved constructors should not be preserved - the super or this call must be, but the rest of the content can be deleted or replaced with throw new Error().

As explained above, just preserving the super() call is hard, since there could be local variables inside that super() which are defined somewhere else in the same constructor. I think we can somewhat resolve this issue by removing every statement that appears after the super() call.

kelloggm:
> As explained above, just preserving the super() call is hard, since there could be local variables inside that super() call, which are defined somewhere in the same constructor. I think we can somewhat resolve this issue by removing every statement that appears after the super() call.

That's not the case: the grammar [requires](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.8.7) the `super`/`this` call to be the first statement in a constructor, if it exists.

LoiNguyenCS:
Professor,

Thanks to your insight, I have been able to simplify the codes. Please review again. Thank you.

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
```diff
@@ -303,7 +309,23 @@ public Visitable visit(ConstructorDeclaration constructorDecl, Void p) {
...
+      if (!needToPreserveSuperOrThisCall(constructorDecl.resolve())) {
+        constructorDecl.setBody(StaticJavaParser.parseBlock("{ throw new Error(); }"));
+      } else {
+        NodeList<Statement> bodyStatement = constructorDecl.getBody().getStatements();
+        if (bodyStatement.size() == 0) {
+          return constructorDecl;
+        }
+        Statement firstStatement = bodyStatement.get(0);
+        if (firstStatement.isExplicitConstructorInvocationStmt()) {
+          BlockStmt minmized = new BlockStmt();
```
kelloggm, line 321: typo: "minimized"

kelloggm (approved):
This implementation is much cleaner, thanks!

LoiNguyenCS:
I'm not sure why the CI is failing. Maybe something is wrong with the relevant Python script?

kelloggm:
https://github.com/tahiat/specimin-evaluation/pull/25 was recently merged. I suspect that's the cause of the failure. @tahiat can you take a look at this failure and, if the cause is in fact https://github.com/tahiat/specimin-evaluation/pull/25, roll back ore revert that PR? Once you've done so, we can rerun CI here.

PR #223, opened 2024-04-02
Enforce Specimin to not create any synthetic classes for packages from Java language

LoiNguyenCS:
Professor,

Some methods from the classes of the Java language might have generic types, such as `requireNonNull(T)`. These methods will confuse Specimin, making Specimin assumes these methods as unsolved and create corresponding synthetic classes. The solution is simple: we make sure that we don't create any synthetic files for the classes from the Java language.


kelloggm (approved):
LGTM

PR #227, opened 2024-04-05
Avoid crash when encountering lambda param

kelloggm:
Locally, this fixed the proximate cause of the cf-3032 crash. However, there is another crash hiding behind this one, it seems:

```
Exception in thread "main" java.lang.RuntimeException: Error calculating the type of parameter BillingEvent::getInvoiceGroupingKey of method call MapElements.into(TypeDescriptor.of(InvoiceGroupingKey.class)).via(BillingEvent::getInvoiceGroupingKey)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveMethodAsUsage(JavaParserFacade.java:751)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodReferenceExprContext.inferArgumentTypes(MethodReferenceExprContext.java:108)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodReferenceExprContext.solveMethod(MethodReferenceExprContext.java:72)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:337)
cf-3032 - FAIL
```

Still, this is progress towards handling this example, and these changes solve the test cases that I wrote.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
@@ -372,7 +372,11 @@ public class TargetMethodFinderVisitor extends ModifierVisitor<Void> {
         if (para.getParentNode().isPresent() && para.getParentNode().get() instanceof CatchClause) {
           paramType = para.getType().resolve();
         } else {
-          paramType = para.resolve().getType();
+          try {
+            paramType = para.resolve().getType();
+          } catch (UnsupportedOperationException e) {
+            throw new RuntimeException("cannot solve: " + para, e);
+          }
         }
 
         if (paramType.isReferenceType()) {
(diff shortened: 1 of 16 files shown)


src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -359,20 +360,26 @@ public Visitable visit(MethodDeclaration method, Void p) {
...
         // Bug report: https://github.com/javaparser/javaparser/issues/4240
         ResolvedType paramType;
         if (para.getParentNode().isPresent() && para.getParentNode().get() instanceof CatchClause) {
           paramType = para.getType().resolve();
         } else {
-          paramType = para.resolve().getType();
+          try {
+            paramType = para.resolve().getType();
+          } catch (UnsupportedOperationException e) {
+            throw new RuntimeException("cannot solve: " + para, e);
```
kelloggm, line 381: I added this as a debugging aid, but I don't see a reason to remove it - it's useful to know what causes crashes

LoiNguyenCS (approved):
LGTM.

Professor, are you working on cf-3032 at the moment?

Issue #221, opened 2024-03-30
specimin crashes for cf-6030 on other target method


tahiat:
Since the minimized program of cf-6030 does not reproduce the [crash stack](https://github.com/typetools/checker-framework/issues/6030), I tried changing the target method of `DataResolver.java`. I am providing the crash stacks for each targets.

Duplicates exists and for details please [refer](https://docs.google.com/document/d/1zHqZ94rkXwddUflTsNxwv8N3oNPfHKyK2Sy55kcxkrE/edit?usp=sharing)  

`--targetMethod "org.apache.cassandra.service.reads.DataResolver#resolve(Runnable)" ` 

> > Task :run FAILED
> Exception in thread "main" java.lang.RuntimeException: encountered a lambda when the package name was unknown
>         at org.checkerframework.specimin.UnsolvedSymbolVisitor.getArgumentTypesImpl(UnsolvedSymbolVisitor.java:1986)
>         at org.checkerframework.specimin.UnsolvedSymbolVisitor.getArgumentTypesFromMethodCall(UnsolvedSymbolVisitor.java:1927)
>         at org.checkerframework.specimin.UnsolvedSymbolVisitor.declaredInCurrentClass(UnsolvedSymbolVisitor.java:1147)
>         at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:901)
>         at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:83)
>         at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:116)
>         at com.github.javaparser.ast.visitor.ModifierVisitor.lambda$visit$79(ModifierVisitor.java:852)
>         at java.base/java.util.Optional.map(Optional.java:260)
>         at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:852)
>         at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
>         at com.github.javaparser.ast.stmt.ReturnStmt.accept(ReturnStmt.java:78)
>         at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
>         at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
>         at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
[...]

PR #228, opened 2024-04-05
Address the limitation of JavaParser

LoiNguyenCS:
Professor,

This PR fixes one of the crashes in #221.

I can not figure out the specific case where `JavaParser` will fail if there are generic types, hence the word "sometimes" in the explanatory comments. It seems to be that we need to have a generic type with generic bound or something similar for `JavaParser` to fail.

Right now this limitation of `JavaParser` only affect the jar mode, so it is not too serious.  

"


kelloggm (approved):
I'd prefer a test case, but given the proximity of the deadline I'm going to allow this through without one for now.

PR #230, opened 2024-04-09
Handle Override annotation

LoiNguyenCS:
Professor,

This PR handles the missing overridden method bug related to cf-577. Please take a look. Thank you.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -1449,6 +1449,7 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
       boolean updatingInterface) {
     String methodName = "";
     List<String> listOfParameters = new ArrayList<>();
+    String accessModifer = "public";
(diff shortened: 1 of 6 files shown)


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1445,10 +1445,11 @@ private void updateSyntheticClassesForTypeVar(ClassOrInterfaceType type) {
   public void updateUnsolvedClassOrInterfaceWithMethod(
       Node method,
       @ClassGetSimpleName String className,
-      @ClassGetSimpleName String desiredReturnType,
+      String desiredReturnType,
```
LoiNguyenCS, line 1448: It seems there's a `@ClassGetSimpleName` annotation present here, but I'm unsure why. We generally prefer fully-qualified names for return types, just as we do for parameter types.

kelloggm: I agree that this is a mistake, so removing it is correct.

kelloggm (approved):
This change is great, thank you!

PR #235, opened 2024-04-09
Fix bug in TargetMethodFinderVisitor's handling of annotations

kelloggm:
This was preventing me from testing a hypothesis related to cf-691. Sadly, my hypothesis was wrong (but this code allowed me to prove it!), so I'm still working on that.


LoiNguyenCS (approved):
LGTM

PR #237, opened 2024-04-09
Don't create raw types

kelloggm:
Previously, the synthetic method for `containsAll` in `Baz` had the parameter type `com.example.Baz` (without the `<?>`, meaning it is a raw type). There are two problems with that:
* raw types trigger a warning from javac by default
* in this case, it causes the output not to compile (because the implicit override is not correct - the types aren't compatible)

Based on an example I found while investigating cf-691.


LoiNguyenCS (approved):
LGTM. Thank you.

PR #240, opened 2024-04-09
Look at all classes in each file, not just top-level declarations, when constructing the existingClassesToFilePath map

kelloggm:


LoiNguyenCS (approved):
LGTM. Thank you.

PR #241, opened 2024-04-10
Jar mode execution ci pipeline and status json add

tahiat:


tahiat:
depends on https://github.com/tahiat/specimin-evaluation/pull/37

src/main/resources/jar_mode_specimin_execution_status.json
```diff
@@ -0,0 +1,23 @@
+{
+  "cf-1291": "FAIL",
```
kelloggm, line 2: Why are these all FAIL? Is it because specimin-evaluation#37 hasn't yet been merged, so you don't know? Or are they actually all failing in local testing (if so, that's a crisis).

tahiat: Professor, I only configured jar mode for cf-6282, other project is not yet configured for jar mode.
in jar mode, specimin producing minimized program for cf-6282. However compilation is failing  for the minimized program.

tahiat: The printed table will show the reason of failure

LoiNguyenCS: I'm curious. Did you use the jar files when compiling these projects? In essence, whichever jar files given to Specimin should also be used when compiling the minimized programs.

tahiat: you are right Loi, I did not use the jars when compiling these minimized projects. I need to modify the gradle script to use the jars. Now configuring that.

src/main/resources/jar_mode_min_program_compilation_status.json
```diff
@@ -0,0 +1,22 @@
+{
+  "cf-1291": "FAIL",
```
kelloggm, line 2: Are these accurate? If so, that's very concerning.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -76,6 +76,8 @@ public static void main(String... args) throws IOException {
       jarFiles = getJarFiles(jarDirectory);
     }
 
+    jarFiles.stream().forEach(System.out::println);
```
kelloggm, line 79: This has to be removed before merging.

kelloggm (changes requested):

src/main/resources/jar_mode_min_program_compilation_status.json
tahiat: Projects where I could not write the script to pull dependencies, specimin was not run on them and results are showing as  FAIL.
specimin was not executed on cf-1291 in jar mode. Thats why status is FAIL

PR #242, opened 2024-04-10
Use decompiler

LoiNguyenCS:
Professor,

This is the PR to invoke a decompiler. I have decided to use [VineFlower](https://vineflower.org). It is built based on [FernFlower](https://github.com/fesh0r/fernflower). However, VineFlower supports Gradle, and it seems to be [cooler](https://www.reddit.com/r/java/comments/ue8u59/new_open_source_java_decompiler/) according to its creator. :-)

Sorry for adding the try-catch statements when parsing the file. As explained in the comments, it is actually difficult to fix a bug that can't be reproduced locally. But since those bugs do not happen locally., I think we don't have to worry about them.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
+++ b/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
@@ -321,6 +321,9 @@ public class SpeciminRunner {
         cu.accept(inheritancePreserve, null);
       }
       for (String targetFile : inheritancePreserve.getAddedClasses()) {
+        if (targetFile.startsWith("java.")) {
+          continue;
+        }
         String directoryOfFile = targetFile.replace(".", "/") + ".java";
         File thisFile = new File(root + directoryOfFile);
         // classes from JDK are automatically on the classpath, so UnsolvedSymbolVisitor will not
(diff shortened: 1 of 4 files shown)


src/test/java/org/checkerframework/specimin/JarFileTest.java
```diff
@@ -8,7 +8,7 @@ public class JarFileTest {
   @Test
   public void runTest() throws IOException {
     SpeciminTestExecutor.runTest(
-        "nodependenciesreturnssame",
+        "jarfile",
```
LoiNguyenCS, line 11: I figured out this mistake while trying to run the test.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -185,7 +199,14 @@ public static void performMinimization(
...
       }
       for (String targetFile : addMissingClass.getAddedTargetFiles()) {
-        parsedTargetFiles.put(targetFile, parseJavaFile(root, targetFile));
+        try {
+          parsedTargetFiles.put(targetFile, parseJavaFile(root, targetFile));
+        } catch (ParseProblemException e) {
+          // These parsing codes cause crashes in the CI. Those crashes can't be reproduced locally.
+          // Not sure if something is wrong with VineFlower or Specimin CI. Hence we keep these
+          // lines as tech debt.
+          continue;
```
kelloggm, line 208: This does make me very nervous, but I agree it's okay to keep it for now. Please make this comment a TODO, and add another TODO at each of the other places that you've added such a catch below.

kelloggm (approved):
LGTM other than the need for a few more TODO comments. Can be merged without another review after those are added.

PR #243, opened 2024-04-11
Unsolved return types could be interfaces

kelloggm:
Unfortunately I had to make `isAnInterface` non-final, but I couldn't figure out a way to check whether there is such an `implements` clause at the point where I wrote the new comment around line 1500 (which is the ideal place to fix this).

I also made one small refactoring to get rid of an antipattern.


LoiNguyenCS (approved):
LGTM

PR #245, opened 2024-04-11
Don't crash when encountering a field inside an enum

kelloggm:
...also required a corresponding change in target method finder visitor to allow us to target methods of enums properly. Why enums aren't a regular class I have no idea.

~~May~~ does fix the crash in jdk-8288590 ~~(I haven't checked yet, that's what CI is for)~~

Also makes jdk-8288590 compile (yay!)


LoiNguyenCS (approved):
LGTM. Thanks.

PR #247, opened 2024-04-11
Allow target fields

LoiNguyenCS:
Professor,

This is the PR to allow target fields.

I have decided to add -`-targetFields` as an additional argument to run Specimin. Another approach would be to change` --targetMethod` to `--targetMembers`, but that would require refactoring in too many places: test files, field names in the Visitors, comments related to target methods, scripts in specimin-evaluation, and so on.


src/test/java/org/checkerframework/specimin/SpeciminTestExecutor.java
```diff
@@ -69,9 +70,14 @@ public static void runTest(
       speciminArgs.add("--targetFile");
       speciminArgs.add(targetFile);
     }
-    for (String targetMethod : targetMethods) {
-      speciminArgs.add("--targetMethod");
-      speciminArgs.add(targetMethod);
+    for (String targetMember : targetMembers) {
```
LoiNguyenCS, line 73: This is the most elegant approach I can come up with that does not require modifying all the existing 148 test files.

tahiat:
Please, merge the latest main to your branch

src/test/java/org/checkerframework/specimin/SpeciminTestExecutor.java
kelloggm: This seems fine to me for now. We can come back to this after the deadline if you don't think it's elegant enough.

kelloggm (approved):
LGTM

kelloggm:
Waiting to merge this until I change the evaluation script so that cf-691 targets the correct field, and we can update the preservation status on this branch.

kelloggm:
I am working on fixing the fact that with this change the output of CF 691 is not correct. Since this branch just reveals the problem (rather than causing it), that fix can go in a separate branch.

PR #248, opened 2024-04-12
Fix a bunch more problems related to enums

kelloggm:
Includes the changes in #245, so don't review until that one is merged.

This should make cf-3025 compile (fingers crossed, seems to be working locally). It doesn't look like we're correctly preserving the crash in javac, though. I'll investigate that later.

I took on quite a bit of technical debt on purpose in this PR that I will fix next week.


LoiNguyenCS (approved):
LGTM

PR #249, opened 2024-04-12
Skip processing explicit constructor invocations until all arguments can be solved

kelloggm:
This adopts the same strategy we use for "regular" method calls.

Fixes one crash on CF-691. However, another is revealed.


LoiNguyenCS (approved):
LGTM.

PR #251, opened 2024-04-12
Do something sensible with unsolved inner classes

kelloggm:
This definitely an incomplete solution, but it avoids the crash in cf-691 (and should handle most cases in the wild, IMO).

Note that the changes around line 738 of `UnsolvedSymbolVisitor` fix one crash, but that fix causes another crash (related to mismatched type variables) that the inner class logic fixes.


kelloggm:
Because @LoiNguyenCS  is unavailable, I'm going to merge this PR without review. @LoiNguyenCS you should take a look at this when you have a chance (maybe next week).

PR #252, opened 2024-04-12
Preserve the bounds of the type parameters of used classes

kelloggm:
This should be part of the modularity model. The previous test case that showed we didn't do this was wrong, and I should not have approved it.

Should fix JDK-8319461.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/InheritancePreserveVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/InheritancePreserveVisitor.java
@@ -57,6 +57,9 @@ public class InheritancePreserveVisitor extends ModifierVisitor<Void> {
     addedClasses = new HashSet<>();
   }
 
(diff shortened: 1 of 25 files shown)


kelloggm:
If this passes CI, I will override approval requirements and merge it without review from @LoiNguyenCS. He should review it later when he gets a chance.

LoiNguyenCS (approved):
LGTM. 

I was aware of the fact that Specimin did not preserve a type bound if that type was not used by the target methods. I just assumed that removing the unused type bound should not affect an analysis tool.

kelloggm:
> I just assumed that removing the unused type bound should not affect an analysis tool.

No worries, I should have noticed earlier. This is why we weren't preserving one of the JDK bugs, though, so I wanted to fix it.

kelloggm:
@LoiNguyenCS next week you should look through the other changes I made in this PR to keep CF-6388 compiling. They're pretty extensive; we had the modularity model for interfaces wrong, too.

PR #260, opened 2024-04-22
Treat static methods and static fields the same way wrt our simple name heuristics

kelloggm:


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -2701,10 +2702,30 @@ public boolean unsolvedAndCalledByASimpleClassName(MethodCallExpr method) {
       if (callerExpression.isEmpty()) {
         return false;
       }
-      return classAndPackageMap.containsKey(callerExpression.get().toString());
+      String callerExpressionString = callerExpression.get().toString();
+      return classAndPackageMap.containsKey(callerExpressionString)
```
LoiNguyenCS, line 2706: It seems we can safely eliminate the check for `classAndPackageMap.containsKey(callerExpressionString)`. Since we're only dealing recognizing simple class names, verifying its presence in `classAndPackageMap` doesn't offer significant value. It appears I overlooked this while drafting the code, likely due to not enough coffee intake.

LoiNguyenCS (changes requested):

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: > It seems we can safely eliminate the check for classAndPackageMap.containsKey(callerExpressionString).

While technically that's probably true, it's both cheaper and more reliable to check there than to do the "string guessing" that my new logic does. Personally, I'd prefer to keep both, even though we know that the map is a bit redundant most of the time.

Another advantage of the map is that it will catch imported classes that don't follow convention, too, so it's able to cover more than the heuristic. I think we definitely need both.

> likely due to not enough coffee intake

lol, easy mistake to make :)

LoiNguyenCS (approved):

Issue #164, opened 2024-03-01
UnsolvedSymbolVisitor is stuck at one or more exception


jonathan-m-phillips:
`--args=--outputDirectory "/temp/dir" --root "/my/root/bcel-util/src/main/java" --targetFile "org/plumelib/bcelutil/StackVer.java" --targetMethod "org.plumelib.bcelutil.StackVer#do_stack_ver(MethodGen)"`

`--args=--outputDirectory "temp/dir" --root "/root/bcel-util/src/main/java" --targetFile "org/plumelib/bcelutil/StackTypes.java" --targetMethod "org.plumelib.bcelutil.StackTypes#set(int, Frame)"`

`--args=--outputDirectory "tempdir" --root "/root/bcel-util/src/main/java" --targetFile "org/plumelib/bcelutil/StackTypes.java" --targetMethod "org.plumelib.bcelutil.StackTypes#toString(OperandStack)"`

`--args=--outputDirectory "tempdir" --root "/root/bcel-util/src/main/java" --targetFile "org/plumelib/bcelutil/StackTypes.java" --targetMethod "org.plumelib.bcelutil.StackTypes#toString(LocalVariables)"`

`--args=--outputDirectory "/tempdir" --root "/root/bcel-util/src/main/java" --targetFile "org/plumelib/bcelutil/SimpleLog.java" --targetMethod "org.plumelib.bcelutil.SimpleLog#log(String, Object)"`

`--args=--outputDirectory "/tempdir" --root "/root/bcel-util/src/main/java" --targetFile "org/plumelib/bcelutil/BcelUtil.java" --targetMethod "org.plumelib.bcelutil.BcelUtil#instructionListToString(InstructionList, ConstantPoolGen)"`

`--args=--outputDirectory "/temodir" --root "/root/bcel-util/src/main/java" --targetFile "org/plumelib/bcelutil/BcelUtil.java" --targetMethod "org.plumelib.bcelutil.BcelUtil#localVariablesToString(MethodGen)"`

`--args=--outputDirectory "/tempdir" --root "/root/bcel-util/src/main/java" --targetFile "org/plumelib/bcelutil/BcelUtil.java" --targetMethod "org.plumelib.bcelutil.BcelUtil#attributeNameToString(Attribute, ConstantPoolGen)"`

[...]

PR #261, opened 2024-04-22
Also ensure that the arguments to constructor invocations are solved before deciding that a new expression is solvable

kelloggm:
Fixes a bug exposed by one of the test cases from #164. I can no longer reproduce the actual failure there, so I think we can close it.


LoiNguyenCS (approved):
LGTM

PR #264, opened 2024-04-25
add the --targetField argument to the README

kelloggm:


README.md
```diff
@@ -25,6 +25,7 @@ The available options are (required options in **bold**, repeatable options in *
 * **--root**: specifies the root directory of the target project.
 * ***--targetFile***: a source file in which to search for target methods
 * ***--targetMethod***: a target method that must be preserved, and whose dependencies should be stubbed out. Use the format `class.fully.qualified.Name#methodName(Param1Type, Param2Type, ...)`. Note: If a target method has a receiver parameter, i.e., (.. this), exclude that parameter from the signature. Check this [documentation](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-ReceiverParameter) for more info.
+* ***--targetField***: a target field that must be preserved (including its initializer). Uses the same format as `--targetMethod`, but without the parameter list. At least one of `--targetMethod` or `--targetField` is required, but they can otherwise be freely combined.
```
LoiNguyenCS, line 28: > At least one of `--targetMethod` or `--targetField` is required

This line is not precise. We can run Specimin without any target method or field. Specimin won't return any output in that case, but it won't raise any exception either.

LoiNguyenCS (changes requested):
I just have a minor comment regarding the necessity of `--targetMethod` and `--targetField`.

LoiNguyenCS (approved):
LGTM

PR #266, opened 2024-04-25
Update JavaParser

LoiNguyenCS:
Professor,

This is the PR to use the newest version of JavaParser.

The type variables collision test is failing. Here is the input file:

```java
package org;

// This import isn't used.
import com.T;

public class Foo<T> {
    T useT(T t) {
        return t;
    }
}
```
It turns out that this new version of JavaParser confuses the `T` in `useT` to be `com.T`. I don't think this is a serious bug, since we will only end up with an extra file in the final output of Specimin. I am thinking of removing the type variables collision test since the new version of JavaParser can not discern those collisions properly anymore. Let me know what you think.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
@@ -64,15 +64,8 @@ public class GetTypesFullNameVisitor extends ModifierVisitor<Void> {
   public Visitable visit(ClassOrInterfaceType type, Void p) {
     String typeFullName;
     try {
-      /*
-       * In JavaParser, ClassOrInterfaceType is a subtype of ReferenceType
-       * (check an example here: https://github.com/javaparser/javaparser/blob/9c133d19d5b85b3b758f05762fb4d7c9875ef681/javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java#L258).
-       * However, the resolve() method in ClassOrInterfaceType only returns a ResolvedType instead of a specific ResolvedReferenceType.
-       * This appears to be an inaccuracy within JavaParser's type hierarchy.
-       */
-
-      // since type is a ClassOrInterfaceType instance, it is safe to cast it to a ReferenceType.
-      typeFullName = type.resolve().asReferenceType().getQualifiedName();
+      typeFullName =
+          JavaParserUtil.classOrInterfaceTypeToResolvedReferenceType(type).getQualifiedName();
     } catch (UnsolvedSymbolException | UnsupportedOperationException e) {
       return super.visit(type, p);
     }
--- a/src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
@@ -64,7 +64,8 @@ public class GetTypesFullNameVisitor extends ModifierVisitor<Void> {
   public Visitable visit(ClassOrInterfaceType type, Void p) {
     String typeFullName;
     try {
-      typeFullName = type.resolve().getQualifiedName();
+      // since type is a ClassOrInterfaceType instance, it is safe to cast it to a ReferenceType.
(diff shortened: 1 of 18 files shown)


kelloggm:
> I don't think this is a serious bug, since we will only end up with an extra file in the final output of Specimin.

I agree that this is acceptable imprecision.

> I  am thinking of removing the type variables collision test

Rather than removing the test, I would prefer that we keep the test but change the expected output to the new behavior. We should add a comment indicating that the `com.T` usage isn't accurate, but that it's a limitation of JavaParser.

src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
```diff
@@ -192,7 +192,9 @@ private boolean overridesAnInterfaceMethodImpl(
     for (ClassOrInterfaceType implementedType : implementedTypes) {
       ResolvedReferenceType resolvedInterface;
       try {
-        resolvedInterface = implementedType.resolve();
+        // since implementedType is a ClassOrInterfaceType instance, it is safe to cast it to a
+        // ReferenceType.
```
kelloggm, line 196: If this is the case, why doesn't `ClassOrInterfaceType#resolve()` return `ReferenceType`?

kelloggm: Is this imprecision in JavaParser's type hierarchy, or is there something else that might be returned here? A link to JavaParser (either the code or the docs) is probably the best evidence you could add here to convince me that this is safe.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
kelloggm: The Javadoc you linked is for an old version (3.2.5), not the latest version.

However, I agree with your explanation after looking into myself. In particular, `ClassOrInterfaceType`'s implementation makes the same assumption, e.g.: https://github.com/javaparser/javaparser/blob/9c133d19d5b85b3b758f05762fb4d7c9875ef681/javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java#L258

I think this is safe, and that the return type of the `resolve()` method could be more specific.

LoiNguyenCS:
Professor,

I have added the explanations for the safety of casting ClassOrInterfaceType to ReferenceType and updated the test output for typevar-collision.

src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
```diff
@@ -64,7 +64,15 @@ public Visitable visit(EnumDeclaration decl, Void p) {
   public Visitable visit(ClassOrInterfaceType type, Void p) {
     String typeFullName;
     try {
-      typeFullName = type.resolve().getQualifiedName();
+      /*
+       * In JavaParser, ClassOrInterfaceType is a subtype of ReferenceType
+       * (check an example here: https://github.com/javaparser/javaparser/blob/9c133d19d5b85b3b758f05762fb4d7c9875ef681/javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java#L258).
+       * However, the resolve() method in ClassOrInterfaceType only returns a ResolvedType instead of a specific ResolvedReferenceType.
+       * This appears to be an inaccuracy within JavaParser's type hierarchy.
+       */
```
kelloggm, line 72: I think this explanation should only appear once in the code. What I suggest you do is add a utility method of type `ClassOrInterfaceType -> ResolvedReferenceType` that has the comment, and then call at each of these locations.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/GetTypesFullNameVisitor.java
LoiNguyenCS: I have added a utility method. Please review again. Thank you.

kelloggm (approved):
LGTM, thanks!

PR #262, opened 2024-04-22
Fix handling of boolean expressions

kelloggm:
TODO: also look into unary expressions ~~and the `==` and `!=` operators~~ later.

~~TODO: tests for other operators?~~

PR #267, opened 2024-04-26
Test for unary expressions

kelloggm:
Based on the bug that was fixed by #262, I was concerned that our handling of unary expressions might also be incorrect. I checked for a test case, and found that we didn't have one (because we didn't have a test containing the string `+=`), so I wrote the one in this PR. To my pleasant surprise, the test passes - I think our usual type correct algorithm actually handles this case just fine, since there is no special error message for mistakes in unary operator types (as there is for binary operators).


LoiNguyenCS (approved):
LGTM

PR #271, opened 2024-04-26
Don't lose information in synthetic types when correcting their type via JavaTypeCorrect

kelloggm:
Also incidentally cleaned up some ISSTA tech debt that happened to get in the way.


LoiNguyenCS (approved):
LGTM

Issue #272, opened 2024-04-27
Potential infinite loop when hitting a specific method


jonathan-m-phillips:
Running ASHE on dryrun mode, we haven gotten hung up on the same method multiple times for hours. This makes me think there is an infinite loop when minimizing a specific method.

`--outputDirectory "/my/temp/dir" --root "/my/root/checker-framework/framework/src/main/java" --targetFile "org/checkerframework/framework/util/typeinference8/types/AbstractType.java" --targetMethod "org.checkerframework.framework.util.typeinference8.types.AbstractType#getFunctionTypeParameterTypes()"`

java.lang.RuntimeException: Unfound parent for this class: JavaExpression
Caused by: UnsolvedSymbolException{context='mapList', name='CollectionsPlume.mapList(JavaExpression::fromNode, invocationNode.getArguments())', cause='null'}

That is the most consistent error, but there were two other times it appeared we hit an infinite loop at different methods:
`--outputDirectory "/my/temp/dir" --root "/my/root/checker-framework/framework/src/main/java" --targetFile "org/checkerframework/framework/util/Contract.java" --targetMethod "org.checkerframework.framework.util.Contract#viewpointAdaptDependentTypeAnnotation(GenericAnnotatedTypeFactory<?,?,?,?>, StringToJavaExpression, Tree)"`

`--outputDirectory "/my/temp/dir" --root "/my/root/checker-framework/dataflow/src/main/java" --targetFile "org/plumelib/util/ArrayMap.java" --targetMethod "org.plumelib.util.ArrayMap#newArrayMapOrLinkedHashMap(int)"`

jonathan-m-phillips:
Another targeted method I found with an infinite loop:
`--outputDirectory "/temp/dir" --root "/my/root/randoop/src/main/java" --targetFile "randoop/output/PrimitiveAndWrappedTypeVarNameCollector.java" --targetMethod "randoop.output.PrimitiveAndWrappedTypeVarNameCollector#visit(VariableDeclarationExpr, Set<String>)"`

If you would like me to create a new issue for this, please let me know.

kelloggm:
> If you would like me to create a new issue for this, please let me know.

Please do. This issue should have been closed when I merged #275.

PR #273, opened 2024-04-29
Delete all synthetic files, even in the event of a crash, SIGINT, or SIGTERM

kelloggm:
While debugging the infinite loop in #272, I got frustrated by the number of temporary files that I was cleaning up manually, and decided to take a detour to fix this long-standing annoyance in debugging Specimin.

I tested this change by sending SIGINT (by typing CTRL-C) to an infinite-looping Specimin process targeting the Checker Framework (i.e., my reproduction script for #272), and confirming that no temp files are left in the input directory.


LoiNguyenCS (approved):
Thank you. This is very helpful. I also had to manually delete a lot of temp files while debugging Specimin.

PR #275, opened 2024-04-29
Fix infinite loop when extending a class fails

kelloggm:
Review and merge after #274 (includes the changes in that PR). Fixes the infinite loop in #272.

I was not able to find a minimal test case for this bug, unfortunately. I've confirmed that Specimin now terminates on the following input, which #272 reported caused an infinite loop and which I could repro locally:
```
--outputDirectory "/my/temp/dir" --root "/my/root/checker-framework/framework/src/main/java" --targetFile "org/checkerframework/framework/util/Contract.java" --targetMethod "org.checkerframework.framework.util.Contract#viewpointAdaptDependentTypeAnnotation(GenericAnnotatedTypeFactory<?,?,?,?>, StringToJavaExpression, Tree)"
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,12 +58,12 @@ task expectedTestOutputsMustCompile(type: Exec) {
 
 tasks.compileJava {
     // uncomment for testing
-    options.errorprone.enabled = false
+    // options.errorprone.enabled = false
 }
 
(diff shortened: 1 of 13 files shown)


LoiNguyenCS (approved):
LGTM

PR #276, opened 2024-04-29
Make type correction and inner classes compose correctly

kelloggm:
~~This PR cannot be reviewed until both #274 and #275 have been merged, because it incorporates both of those.~~


kelloggm:
@LoiNguyenCS I think this PR is ready for review (as long as it passes CI).

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -934,7 +934,11 @@ public Visitable visit(MethodDeclaration node, Void arg) {
     } else if (insideTargetMember) {
       return processMethodDeclaration(node);
     } else {
-      return super.visit(node, arg);
+      // Do not call super.visit(): this method is definitely unused by the targets, and so
```
LoiNguyenCS, line 937: Wow, this is quite a subtle bug. Thanks for pointing it out.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1347,6 +1353,19 @@ private void solveSymbolsForClassOrInterfaceType(
         }
         if (isTypeVar(typeArgStandardForm)) {
           preferredTypeVariables.add(typeArgStandardForm);
+        } else if (typeArgument.isClassOrInterfaceType()) {
```
LoiNguyenCS, line 1356: Is this code written due to some observed bugs in Specimin? I would expect the visit method for `ClassOrInterfaceType` to visit every `ClassOrInterfaceType` instance. These lines seem to assume that the visit method for `ClassOrInterfaceType` skip `ClassOrInterfaceType` instances if those instances appear as type variable.

LoiNguyenCS (approved):
LGTM. I just have a minor question.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: I'm not sure that this fix is really the correct one. Would it be better to move all of the data structure maintenance from `processMethodDeclaration` to this visit method, instead?

I'm a bit concerned that not calling `super.visit()` here will lead to some other subtle bug in the future, but I can't articulate exactly why it makes me nervous.

LoiNguyenCS: It seems fine to me, since we are not using any local variable in those methods, and the data structure maintenance in `processMethodDeclaration` is simply for updating the list of local variables.

PR #281, opened 2024-05-07
Re-organizing and documenting UnsolvedSymbolVisitor

LoiNguyenCS:
Professor,

I've made some updates so that `UnsolvedSymbolVisitor` can be more accessible for future programmers.

Initially, I attempted to write perfect specifications so that the `UnsolvedSymbolVisitor`'s functionality could be understood without the need to read the code at all. However, I found this to be more challenging than anticipated—Leslie Lamport makes it sound deceptively simple! As an alternative approach, I've added comments to clarify potentially confusing sections of the code.

Additionally, I've removed the portion of the code responsible for updating parameter types and return types of methods. That functionality is already handled within the visit method for `ClassOrInterfaceType`. (Note: The code for updating parameter types had a useful feature that checked if the parameter was within a catch clause and set the synthetic type to extend Exception. To replicate this, I've introduced a local variable, `insideACatchClause`.)

Please review the changes and let me know if there are any further adjustments needed.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -711,7 +711,7 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
   // These are the codes from JavaParser, so we optimistically assume that these lines are safe.
   public Visitable visit(CatchClause node, Void arg) {
     /*
-     * This method is a copy from the visit(CatchClause, Void) method of JavaParser. We adds lines to update the local variables and isInsideCatchBlockParameter.
+     * This method is a copy from the visit(CatchClause, Void) method of JavaParser. We add lines to update the set of local variables and isInsideCatchBlockParameter.
      */
     HashSet<String> currentLocalVariables = new HashSet<>();
     currentLocalVariables.add(node.getParameter().getNameAsString());
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -1650,27 +1650,12 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
     Node parentNode = node.getParentNode().get();
     Type nodeType = node.getType();
 
-    // This scope logic must happen here, because later in this method there is a check for
-    // whether the return type is a type variable, which must succeed if the type variable
-    // was declared for this scope.
     addTypeVariableScope(node.getTypeParameters());
 
     // since this is a return type of a method, it is a dot-separated identifier
     @SuppressWarnings("signature")
     @DotSeparatedIdentifiers String nodeTypeAsString = nodeType.asString();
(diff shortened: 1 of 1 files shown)


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -245,6 +246,9 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
   /** The qualified name of the current class. */
   private String currentClassQualifiedName = "";
 
+  /** Check if the visitor is inside a catch clause. */
```
kelloggm, line 249: "Check if" here sounds as if this is an operation or procedure, but this field stores state. I would phrase this differently - "true if the visitor is inside a catch clause" would work.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -74,17 +73,19 @@
...
 
 /**
  * The visitor for the preliminary phase of Specimin. This visitor goes through the input files,
- * notices all the methods belonging to classes not in the source codes, and creates synthetic
- * versions of those classes and methods. This preliminary step helps to prevent
+ * notices all the methods, fields, and types belonging to classes not in the source codes, and
+ * creates synthetic versions of those symbols. This preliminary step helps to prevent
  * UnsolvedSymbolException errors for the next phases.
+ *
+ * <p>Note: To comprehend this visitor quickly, it is recommended for future programmers to start by
```
kelloggm, line 87: "for future programmers" isn't necessary here and can be cut

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -691,7 +705,11 @@ public Visitable visit(CatchClause node, Void p) {
     HashSet<String> currentLocalVariables = new HashSet<>();
     currentLocalVariables.add(node.getParameter().getNameAsString());
     localVariables.addFirst(currentLocalVariables);
+    // Since there can not be a catch clause inside a catch clause, we don't need to use a temporary
+    // variable like in the case of insideTargetMethod.
```
kelloggm, line 709: Do you have evidence for this claim?

I don't see anything in JLS 14.20 that suggests this, and a google search for "nested try-catch java" turns up results that suggest that it is not only possible but somewhat common (even if it's ugly), e.g.:
* https://softwareengineering.stackexchange.com/questions/118788/is-using-nested-try-catch-blocks-an-anti-pattern
* https://softwareengineering.stackexchange.com/questions/237444/alternatives-to-nested-try-catches-for-fallbacks
* etc

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1191,13 +1186,17 @@ public Visitable visit(ObjectCreationExpr newExpr, Void p) {
       return result;
     }
     gotException();
+    /*
+     * For an unresolved object creation, the symbols are resolved first before the expression itself is resolved.
```
kelloggm, line 1190: By "the symbols" here, I think you mean "symbols for the arguments"?

kelloggm (changes requested):

PR #285, opened 2024-05-13
Use instanceof expressions to avoid infinite loop in Randoop

### LINKED ISSUE #279: Potential infinite loop targeting a method in Randoop
Running ASHE on dryrun mode, I believe I have hit a potential loop:
`--outputDirectory "/temp/dir" --root "/my/root/randoop/src/main/java" --targetFile "randoop/output/PrimitiveAndWrappedTypeVarNameCollector.java" --targetMethod "randoop.output.PrimitiveAndWrappedTypeVarNameCollector#visit(VariableDeclarationExpr, Set<String>)"`

When running the project in the background, it gets hung up on this method and does not move forward.

kelloggm:
Fixes #279

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
+++ b/src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
@@ -344,19 +344,13 @@ class JavaTypeCorrect {
         // in the target), so make the rhs a subtype of the lhs.
         // TODO: we must check here that there is no entry for the rhs already.
         // However, it's not clear what the right behavior is when there is
-        // an existing entry. I've set this up to crash to avoid thrashing
+        // an existing entry. I've set this up to do nothing to avoid thrashing
         // behavior like that seen in https://github.com/njit-jerse/specimin/issues/279.
-        if (extendedTypes.containsKey(rhs)) {
-          throw new RuntimeException(
-              "Trying to extend a class that already extends "
-                  + "another class. Class to be extended: "
-                  + rhs
-                  + "\nclass it currently extends: "
-                  + extendedTypes.get(rhs)
-                  + "\nclass it also should extend: "
-                  + lhs);
+        // However, this does sometimes occur, including in some of our test targets,
+        // so we have to not crash.
+        if (!extendedTypes.containsKey(rhs)) {
+          extendedTypes.put(rhs, lhs);
         }
-        extendedTypes.put(rhs, lhs);
       }
     }
     /*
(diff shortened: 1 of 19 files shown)


src/main/java/org/checkerframework/specimin/JavaTypeCorrect.java
```diff
@@ -341,8 +341,16 @@ private void updateTypeToChange(String errorMessage, String filePath) {
...
+        // in the target), so make the rhs a subtype of the lhs.
+        // TODO: we must check here that there is no entry for the rhs already.
+        // However, it's not clear what the right behavior is when there is
+        // an existing entry. I've set this up to do nothing to avoid thrashing
+        // behavior like that seen in https://github.com/njit-jerse/specimin/issues/279.
+        // However, this does sometimes occur, including in some of our test targets,
+        // so we have to not crash.
+        if (!extendedTypes.containsKey(rhs)) {
+          extendedTypes.put(rhs, lhs);
+        }
```
jonathan-m-phillips, line 353: Do we want to add an `else` here just in case some unknown condition occurs?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -743,6 +744,48 @@ public Visitable visit(BlockStmt node, Void p) {
...
+      return super.visit(node, p);
+    }
+
+    if (referenceTypeFQN.equals(relationalExprFQN)) {
+      // A type can't extend itself.
+      return super.visit(node, p);
+    }
+
+    for (UnsolvedClassOrInterface syntheticClass : missingClass) {
+      if (syntheticClass.getQualifiedClassName().equals(referenceTypeFQN)) {
```
jonathan-m-phillips, line 779: Are we sure we cannot get a null pointer from syntheticClass.getQualifiedClassName()?
If so, it makes sense to do this:
```
    String qualifiedClassName = syntheticClass.getQualifiedClassName();
    if (qualifiedClassName != null && qualifiedClassName.equals(referenceTypeFQN)) {
```

Otherwise, this is fine.

jonathan-m-phillips (commented):
Waiting for responses to comments before approving.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: > Are we sure we cannot get a null pointer from syntheticClass.getQualifiedClassName()?

Yes, we're sure. Specimin uses the Checker Framework's [Nullness Checker](https://checkerframework.org/manual/#nullness-checker), so the type system would warn us if `getQualifiedName()` could return an `@Nullable String`.

jonathan-m-phillips (approved):
LGTM

PR #287, opened 2024-05-13
Add an infinite loop defense and extend the fix in #285 to also handle Java 14 pattern variables

kelloggm:
The change to the infinite loop defense in `SpeciminRunner` would have prevented the infinite loop bug in #279 entirely: instead of an infinite loop, the manifestation of the bug would have been non-compilable output from Specimin, which is much less bad. I've tested that with both the new test case in this PR and the test case in #285.


jonathan-m-phillips (approved):
LGTM

PR #293, opened 2024-05-29
Preserve methods when argument and parameter types are different

theron-wang:
When running on a method which references another method, passing in an argument which has a different type than the parameter type, the output file does not preserve the used method, resulting in unintended behavior. For example, when given this input:

```java
import java.util.ArrayList;
import java.util.List;

class Foo {
    void bar() {
        foo(new ArrayList<?>());
    }
    void foo(List<?> baz) {
    }
}
```

The output only contains the `bar()` method, removing the `foo(List<?>)` method entirely:

```java
import java.util.ArrayList;

class Foo {

    void bar() {
        foo(new ArrayList<>());
    }
}
```

This PR aims to fix this issue by checking the parameters of each method used within the target method with the imports, so they are not removed prematurely and eventually cause an `UnsolvedSymbolException`, causing the removal of the method.

I wasn't able to make sure that this didn't cause any unintended changes as I have a Windows computer (the tester does not work on Windows), so if someone could verify that, that would be great.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
@@ -172,6 +172,39 @@ public class PrunerVisitor extends ModifierVisitor<Void> {
       return super.visit(decl, p);
     }
 
+    // Check to see if import is used in a separate method used by the target method(s)
+    for (String member : membersToEmpty) {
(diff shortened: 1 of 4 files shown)


kelloggm (approved):

theron-wang:
Got it, thanks!

Issue #148, opened 2024-02-26
java.lang.IllegalStateException: Analysis.BeforeOrAfter is not NameExpr, it is FieldAccessExpr


jonathan-m-phillips:
The following exception was found when dry running ASHE Automation.

`--args=--outputDirectory "my/Temp/Output" --root "path/to/checker-framework/dataflow/src/main/java" --targetFile "org/checkerframework/dataflow/analysis/AnalysisResult.java" --targetMethod "org.checkerframework.dataflow.analysis.AnalysisResult#getStoreAfter(Block)"`

```
Exception in thread "main" java.lang.IllegalStateException: Analysis.BeforeOrAfter is not NameExpr, it is FieldAccessExpr
	at com.github.javaparser.ast.expr.Expression.asNameExpr(Expression.java:323)
	at org.checkerframework.specimin.TargetMethodFinderVisitor.visit(TargetMethodFinderVisitor.java:431)
	at org.checkerframework.specimin.TargetMethodFinderVisitor.visit(TargetMethodFinderVisitor.java:47)
	at com.github.javaparser.ast.expr.FieldAccessExpr.accept(FieldAccessExpr.java:90)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
	at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:651)
	at org.checkerframework.specimin.TargetMethodFinderVisitor.visit(TargetMethodFinderVisitor.java:372)
	at org.checkerframework.specimin.TargetMethodFinderVisitor.visit(TargetMethodFinderVisitor.java:47)
	at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:116)
	at com.github.javaparser.ast.visitor.ModifierVisitor.lambda$visit$79(ModifierVisitor.java:852)
	at java.base/java.util.Optional.map(Optional.java:260)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:852)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
[...]

PR #294, opened 2024-05-30
Correctly preserve enum constants that are used in a switch statement, as if they were fields

kelloggm:
Based on trying to reproduce #148. The crash that that issue reports no longer is reproducible, but that input does trigger this bug.


LoiNguyenCS (approved):
LGTM

PR #295, opened 2024-05-30
The scope of primitive arrays is not guaranteed to be a NameExpr

kelloggm:
Fixes a set of crashes observed by @jonathan-m-phillips on Randoop of the following form:
```
Exception: java.lang.IllegalStateException,
Message: this.cumulativeSize is not NameExpr, it is FieldAccessExpr\,
Example: com.github.javaparser.ast.expr.Expression.asNameExpr(Expression.java:332)\
```


LoiNguyenCS (approved):
LGTM

PR #297, opened 2024-06-04
Adding intial comments for a naive user

NiharikaJamble:


src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -155,54 +172,111 @@ private static void performMinimizationImpl(
...
+     * =============================================================================STEP 3: Parsing Target Files:====================================================================================
+       >>> It parses the target Java files specified in the targetFiles list and stores them in a map (parsedTargetFiles). 
+       	   This map associates file paths <KEY> with parsed Abstract Syntax Trees (ASTs) of the Java files <VALUE>.
+           Abstract syntax trees are data structures widely used in compilers to represent the structure of program code.     
+       >>>  WHAT IS A COMPILATION UNIT?
+              A CompilationUnit is a data structure used in Java programming to represent the entire content of a Java source file.
+              When we use a parser like StaticJavaParser to read a .java file, it creates a CompilationUnit 
+              that represents all the content of that file in a structured way. 
+              This allows our program to understand and work with the code without just treating it as plain text.
+              We can think of it as a digital blueprint of the Java file.
```
kelloggm, line 206: This explanation of ASTs and compilation units is very basic, and reads condescendingly to me, almost as if the text expects the reader to know nothing. I don't think that's reasonable or accurate: readers of the code that makes a program analysis tool work should be expected to have a basic familiarity with simple concepts in program analysis like ASTs, or should be expected to look them up. Copying the definition of an AST into our code is not necessary or helpful, to most readers.

Moreover, it's important to think about the _cost_ of adding a comment: when editing the code, it's reasonable to read the comments attached to it. Therefore, any edits here will incur a reading cost proportionate to the size of the comments that you add. Each comment therefore must justify itself - it should convey information to an editor that is _non-obvious_ when reading the code without the comment.

I don't think this comment (or most of the comments you have added here, to be frank) meet that bar. While they might be useful for someone reading this code for the first time, they will be a drag on anyone working on this code after their first day or two. So, I don't think I can in good conscience permit them to be merged: we need to optimize the code for its most important and common reader, which is someone maintaining it (not reading it for the first time).

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -155,54 +172,111 @@ private static void performMinimizationImpl(
...
+    /*
+     * ==============================================================================STEP 4: Decompiling JAR Files (Optional):========================================================================
+       >>> If JAR paths are provided (jarPaths list is not empty), it decompiles the JAR files to Java source files in the root directory using the ConsoleDecompiler tool.
+     */
     if (!jarPaths.isEmpty()) {
       List<String> argsToDecompile = new ArrayList<>();
-      argsToDecompile.add("--silent");
-      argsToDecompile.addAll(jarPaths);
-      argsToDecompile.add(root);
+      argsToDecompile.add("--silent");// why is this required?
```
kelloggm, line 219: I don't think it is required, but the name suggests it turns off some kind of logging from the library. That's generally a good choice: we want to produce our own logs.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -155,54 +172,111 @@ private static void performMinimizationImpl(
...
       ConsoleDecompiler.main(argsToDecompile.toArray(new String[0]));
       // delete unneccessary legal files
       FileUtils.deleteDirectory(new File(root + "META-INF"));
     }
 
-    // the set of Java classes in the original codebase mapped with their corresponding Java files.
+    /*
+     * ===================================================================================STEP 5: Mapping Classes:====================================================================================
+       >>> WHAT IS A PRMARY CLASS?
+       		A primary class is a class that has the same name as the Java file where the class is declared.
```
kelloggm, line 230: typo: "PRMARY"

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -155,54 +172,111 @@ private static void performMinimizationImpl(
...
+          4.4 We also loop through all enum declarations in the file and store their names and file paths in the existingClassesToFilePath map.
+     */
+
     Map<String, Path> existingClassesToFilePath = new HashMap<>();
-    // This map connects the fully-qualified names of non-primary classes with the fully-qualified
-    // names of their corresponding primary classes. A primary
-    // class is a class that has the same name as the Java file where the class is declared.
     Map<String, String> nonPrimaryClassesToPrimaryClass = new HashMap<>();
+    
+ // Initialize SourceRoot to help with parsing the Java files in the specified root directory
```
kelloggm, line 249: Comments are not necessary when they don't add any information beyond what the code already says. This is an example of such a comment, but most of the one-line comments here (and throughout this PR) have this property. Please go through and remove all of the comments that are duplicative with the code text itself.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -216,6 +290,12 @@ private static void performMinimizationImpl(
             enumDeclaration.getFullyQualifiedName().get(), pathOfCurrentJavaFile);
       }
     }
+
+    /*
+     * ==============================================================================STEP 6: Symbol Resolution and correcting Java types:========================================================================
+       >>> Visitors for Symbol Resolution:It initializes and utilizes various visitor classes (UnsolvedSymbolVisitor, JavaTypeCorrect, etc.)
+           to resolve unsolved symbols, correct Java types, handle inheritance, and remove unsolved annotations.
```
kelloggm, line 297: This text adds nothing over just reading the code. This is one of the most complex parts of this method, and so probably does deserve more of an explanation, but you haven't given one here. In contrast, you did give long explanations of the simpler parts. To me, that seems exactly backwards: simple parts should get little or no explanation (just read the code), but complex parts that take time to understand deserve real explanations that will speed up that understanding.

kelloggm (changes requested):
While some of these comments may be helpful (and it may be wise to preserve some of them in your own notes), at present this is far from ready to be merged. The core problem is that these comments do not follow the best practice for what a comment is supposed to do:
* it should contain information that is not obvious just from reading the program text
* it should seek to explain why or how something is done, not what is done: another programmer can read the code to see what is done, but not why
* comments should be optimized for reading: that is, they should be short, to the point, and not duplicative

Please take another pass through this and try to remove or modify all comments that do not follow these principles.

Issue #298, opened 2024-06-05
290 total errors on taregtMethod minimization


jonathan-m-phillips:
While trying to run the following commands, we are receiving a total of 290 errors.

`./gradlew run --args=--outputDirectory "/my/tempDir" --root "/path/to/checker-framework/javacutil/src" --targetFile "main/java/org/checkerframework/javacutil/BasicAnnotationProvider.java" --targetMethod "org.checkerframework.javacutil.BasicAnnotationProvider#isSideEffectFree(ExecutableElement)"`

> Note: Recompile with -Xlint:unchecked for details.
> Note: Some messages have been simplified; recompile with -Xdiags:verbose to get full output
> 100 errors
> 1 warning
> only showing the first 100 errors, of 290 total; use -Xmaxerrs if you would like to see more

From what I am seeing, the majority of the errors are stating the following:

> error: incompatible types:

> error: cannot find symbol

> error: method does not override or implement a method from

This issue was found running on ASHE. When trying to compile any files that should have been minimized from `org.checkerframework.javacutil.BasicAnnotationProvider#isSideEffectFree(ExecutableElement)`, we are only getting one file outputted during the minimization, `main/java/org/checkerframework/javacutil/BasicAnnotationProvider.java`. `#isSideEffectFree(ExecutableElement)` is dependent on AnnotationUtils.areSameByName being available. The compilation is as follows:
```
Specimin output:
/tempy/dir/main/java/org/checkerframework/javacutil/BasicAnnotationProvider.java:13: error: cannot find symbol
            boolean found = AnnotationUtils.areSameByName(am, "org.checkerframework.dataflow.qual.SideEffectFree");
                            ^
  symbol:   variable AnnotationUtils
  location: class BasicAnnotationProvider
1 error
```

The issue persists throughout the entirety of the file.

kelloggm:
@jonathan-m-phillips I'm not able to reproduce this problem (i.e., "290 errors"). In particular, Specimin seems to run to completion on the given target with no errors (although the output does not compile). More details on exactly what I did below:

I ran the following command on the `main` branch of Specimin:
```
[...]

PR #300, opened 2024-06-06
Fix package name computation

kelloggm:
I don't have a test case for this problem, since I noticed it while investigating another, unrelated issue (#298).

I think it should be clear what the issue is and that this fix is correct, though.


LoiNguyenCS (approved):
This change looks good to me, but I'm puzzled how Specimin has been passing the [unsolvedstaticmethod](https://github.com/njit-jerse/specimin/tree/packagename-from-parts/src/test/resources/unsolvedstaticmethod) test despite this bug. Maybe there is other code elsewhere doing the same thing as `updateClassSetWithStaticMethodCall`.

LoiNguyenCS:
Ah now I see it. Package name is only necessary when the involved method uses lambda expressions as arguments. Since unsolvedstaticmethod does not involve lambda expressions, this bug still went unnoticed until now.

PR #302, opened 2024-06-11
Fully solve all annotations

theron-wang:
This PR offers a solution to #55, which generates annotations when no jar file is provided. For example, given this input:

```.java
package com.example;

public class Simple {
    @Bar(10)
    @Baz(foo = "foo", bar = 1)
    public static int baz(String bar) {
        return 1;
    }
}
```

The output will contain a file with the same code, as well as two files containing the synthetic annotations (see `SyntheticAnnotationsTest`). Therefore, all annotations on required methods/fields will not be removed.

I apologize for the last commit (`sync`); I forgot to merge _before_ committing, so I could only do it afterwards.

Thank you!


kelloggm:
> I apologize for the last commit (sync); I forgot to merge before committing, so I could only do it afterwards.

No worries - we're not particularly worried about keeping a clean commit history. Since all changes to the `main` branch happen via PRs, the history of PRs in GitHub is more useful than the commit history, anyway.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1911,6 +1923,38 @@ public Visitable processMethodDeclaration(MethodDeclaration node) {
...
+        // it doesn't matter what type they are synthetically generated as
+        if (anno.isNormalAnnotationExpr()) {
+          for (MemberValuePair pair : anno.toNormalAnnotationExpr().get().getPairs()) {
+            unsolvedAnnotation.addMethod(
+                new UnsolvedMethod(
+                    pair.getNameAsString(), "Object", Collections.emptyList(), true));
+          }
+        } else if (anno.isSingleMemberAnnotationExpr()) {
+          unsolvedAnnotation.addMethod(
+              new UnsolvedMethod("param1", "Object", Collections.emptyList(), true));
```
kelloggm, line 1952: We should use `value`, following [JLS 9.7.3](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.7.3).

kelloggm (commented):
Sorry I forgot about this PR while you were away @theron-wang.

As you go over the failing tests, keep in mind that some of them might encode Specimin's current behavior (which you're changing), so you might need to change the expected output.

theron-wang:
@kelloggm Finally ready for review :). I found that the large hierarchy of annotations preserved from the original codebase and `jar` files is necessary for correct behavior; the preservation status test fails for CF-6030 if the annotation is minimized. However, if you don't think that annotations should be preserved this generously, we can probably find a solution to this.

src/test/java/org/checkerframework/specimin/SyntheticAnnotationsTest.java
```diff
@@ -0,0 +1,18 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.Test;
+
+/**
+ * This test checks that if Specimin will work properly where there is a super constructor while the
+ * parent class file is not in the root directory physically
```
kelloggm, line 8: This comment doesn't seem correct (maybe copied from a different test?), and it should be updated.

src/test/resources/annoingenerictarget/expected/org/checkerframework/checker/initialization/qual/Initialized.java
```diff
@@ -0,0 +1,5 @@
+package org.checkerframework.checker.initialization.qual;
+
+@java.lang.annotation.Target({ java.lang.annotation.ElementType.TYPE_USE, java.lang.annotation.ElementType.METHOD })
```
kelloggm, line 3: Why `TYPE_USE` and `METHOD`, specifically? It seems like you've hardcoded these two `ElementType`s, but annotations can be present in other places, too. Can you add a test with an annotation in a different position (e.g., on a class declaration)?

One solution to this problem is to make every synthetic annotation legal in every context. That's what I'd suggest, at least initially, as a conservative but over-approximate solution. However, it might be possible to be more precise by recording the contexts in which the annotation is actually used and then only including those; I don't think that needs to be in this PR, but we can add it as a TODO for the future.

kelloggm (changes requested):
I reviewed only the changes to the tests; these are my questions about the new specification. Once they're settled, I'll take another look at the code.

src/test/resources/annoingenerictarget/expected/org/checkerframework/checker/initialization/qual/Initialized.java
theron-wang: In testing, these two `ElementType` were sufficient for most (if not all) cases. This is probably due to this ([ElementType docs](https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/ElementType.html)):

>The constant [TYPE_USE](https://docs.oracle.com/javase/8/docs/api/java/lang/annotation/ElementType.html#TYPE_USE) corresponds to the 15 type contexts in JLS 4.11, as well as to two declaration contexts: type declarations (including annotation type declarations) and type parameter declarations.

With the exception of `void` methods (hence the inclusion of `METHOD`), the annotations should be interpreted by the compiler to be applied to the type of the declaration instead of the declaration itself, which shouldn't cause any errors since there aren't such things as void field or class declarations, for example.

I've added a new test case `SyntheticAnnotationTarget` in which annotations are applied in many different contexts (like class declarations) to test for compile errors.

If you think this behavior is fine, or if I am missing something, please let me know! I can definitely work on including all `ElementType`s if necessary.

kelloggm: I'm satisfied with the new test, thanks!

My only (mild) remaining concern with respect to this is that I know that annotation processors can tell whether an annotation is a type annotation or a declaration annotation, so which of the two it is can impact their behavior. For the case of unsolved annotations, however, there's no way to resolve this ambiguity so I think the approach you're taking is fine.

src/main/java/org/checkerframework/specimin/AnnotationParameterTypesVisitor.java
```diff
@@ -0,0 +1,400 @@
...
+        || parent instanceof MethodDeclaration
+        || parent instanceof FieldDeclaration)) {
+      parent = parent.getParentNode().orElseThrow();
+    }
+    return parent;
+  }
+
+  /**
+   * Determines if the given Node is a target/used method or class. Should be used following
+   * findClosestParentMemberOrClassLike().
```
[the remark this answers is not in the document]
kelloggm: Also, this method seems like it would be useful in some of the other `Visitor` classes. Maybe we should make a visitor superclass that contains this method and some of the fields that many of the visitors have?

(I'm not expecting you to do that in this PR, but I'd appreciate it if you could add a TODO comment to this method.)

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1911,6 +2041,63 @@ public Visitable processMethodDeclaration(MethodDeclaration node) {
     return result;
   }
 
+  /**
```
kelloggm, line 2044: This method should be static, and should be in `JavaLangUtils`, not `UnsolvedSymbolVisitor` (which has too much functionality as it is).

kelloggm (changes requested):
My review is a bit shallower than I would've liked, because I'm not feeling well and need to rest. I'll take another, deeper look once these comments have been addressed.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -425,13 +425,6 @@ private static void performMinimizationImpl(
     Set<String> updatedUsedClass = solveMethodOverridingVisitor.getUsedClass();
     updatedUsedClass.addAll(totalSetOfAddedInheritedClasses);
 
-    // remove the unsolved annotations in the newly added files.
-    for (CompilationUnit cu : parsedTargetFiles.values()) {
-      cu.accept(annoRemover, null);
```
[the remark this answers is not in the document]
theron-wang: I believe my initial thought process was that it would remove all unsolved annotations before going into `AnnotationParameterTypesVisitor`, and then it would remove any `@Override`s in any files added by `AnnotationParameterTypesVisitor`. Now that I think of it, there shouldn't be any `@Override`s added so the last usage (which you commented on) should be its only usage to remove any possible unsolved annotations/`@Override`s. Thanks!

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -442,6 +435,65 @@ private static void performMinimizationImpl(
...
+      for (CompilationUnit cu : compilationUnitsToSolveAnnotations) {
+        cu.accept(annotationParameterTypesVisitor, null);
+      }
+
+      // add all files related to the target annotations
+      for (String annoFullName : annotationParameterTypesVisitor.getClassesToAdd()) {
+        String directoryOfFile = annoFullName.replace(".", "/") + ".java";
+        File thisFile = new File(root + directoryOfFile);
+        // classes from JDK are automatically on the classpath, so UnsolvedSymbolVisitor will not
+        // create synthetic files for them
```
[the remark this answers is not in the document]
theron-wang: Sorry, the code used here was directly copy-pasted from above (line 370-ish), and I didn't update the comment. I removed the comment here (lines 461/462)since it isn't relevant to the code.

src/main/java/org/checkerframework/specimin/AnnotationParameterTypesVisitor.java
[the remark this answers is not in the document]
theron-wang: > I can see that that is the use case you wrote it for, but I think this method could be used otherwise without danger, no?

Yep, you're right. Sorry that I didn't clarify in the method, I just meant that it is intended to be used afterwards. Since it's not dangerous I removed the comment as it is unnecessary.

> (I'm not expecting you to do that in this PR, but I'd appreciate it if you could add a TODO comment to this method.)
Added👍

theron-wang:
> My review is a bit shallower than I would've liked, because I'm not feeling well and need to rest. I'll take another, deeper look once these comments have been addressed.

Don't worry about it! Take your time and feel better before you review this again.

src/main/java/org/checkerframework/specimin/AnnotationParameterTypesVisitor.java
```diff
@@ -0,0 +1,380 @@
...
+      String memberName = fullName.substring(fullName.lastIndexOf(".") + 1);
+      staticImports.put(memberName, fullName);
+    }
+    return decl;
+  }
+
+  @Override
+  public Visitable visit(AnnotationMemberDeclaration decl, Void p) {
+    // Ensure that enums/fields that are used by default are included
+    // Also, in edge cases, preserve method type since a definition with a default value may be
```
kelloggm, line 99: Can you be more specific in this comment about which edge cases you're referring to?

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
```diff
@@ -4,6 +4,7 @@
 import com.github.javaparser.ast.ImportDeclaration;
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.body.AnnotationDeclaration;
```
kelloggm, line 7: I would not expect spotless (the auto-formatter) to pass with this unused import declaration.

kelloggm: I don't think there should be any changes in this file.

kelloggm (changes requested):
A few more comments

kelloggm:
@theron-wang is this ready for another review?

theron-wang:
@kelloggm Yes, sorry! I forgot to re-request a review.

kelloggm (approved):

PR #310, opened 2024-06-20
Don't crash when a constructor call uses an imported array type from another package

### LINKED ISSUE #86: Unfound parent for this class: <Class_Name> (2)
**NOTE:** This may be solved with Issue #85 

When running Specimin on [MethodSignature.java](https://github.com/mernst/randoop/blob/master/agent/replacecall/src/main/java/randoop/instrument/MethodSignature.java), the exception below was thrown.

```
Exception in thread "main" java.lang.RuntimeException: Unfound parent for this class: MethodSignature
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.getParentClass(UnsolvedSymbolVisitor.java:1869)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.updateSyntheticClassForSuperCall(UnsolvedSymbolVisitor.java:1148)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:532)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:76)
	at com.github.javaparser.ast.expr.NameExpr.accept(NameExpr.java:79)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:142)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.expr.AssignExpr.accept(AssignExpr.java:135)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:477)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.stmt.ExpressionStmt.accept(ExpressionStmt.java:71)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
	at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:169)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:449)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:76)
	at com.github.javaparser.ast.stmt.BlockStmt.accept(BlockStmt.java:70)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:335)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:563)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:76)
	at com.github.javaparser.ast.body.ConstructorDeclaration.accept(ConstructorDeclaration.java:90)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
	at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:243)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:327)
	at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:76)
	at com.github.javaparser.ast.body.ClassOrInterfaceDeclaration.accept(ClassOrInterfaceDeclaration.java:92)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1129)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.NodeList.accept(NodeList.java:277)
	at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1168)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:306)
	at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:51)
	at com.github.javaparser.ast.CompilationUnit.accept(CompilationUnit.java:127)
	at org.checkerframework.specimin.SpeciminRunner.performMinimization(SpeciminRunner.java:131)
	at org.checkerframework.specimin.SpeciminRunner.main(SpeciminRunner.java:64)

```

`--args=--outputDirectory "my/temp/dir" --root "/my/absolute/to/root/randoop/agent/replacecall/src/main/java" --targetFile "randoop/instrument/MethodSignature.java" --targetMethod "randoop.instrument.MethodSignature#of(String, org.apache.bcel.classfile.Method)"`

```
  /**
   * Creates a {@link MethodSignature} object for a {@code org.apache.bcel.classfile.Method} object.
   *
   * @param classname the class containing the method
   * @param method the Method object
   * @return the {@link MethodSignature} representation of the Method object
   */
  public static MethodSignature of(String classname, org.apache.bcel.classfile.Method method) {
    return new MethodSignature(classname, method.getName(), method.getArgumentTypes());
  }
```

**I would recommend running Specimin on this entire file before closing this out!**

kelloggm:
Fixes #86.

There are many parts of this PR, which fix the different parts of this bug. This PR is a draft, because I'd like to try to break out some of these into smaller PRs, which can be reviewed separately; I'm opening this here so that we have a reference branch for all of it together.

The various things in the PR are:
* the first part of the bug is the handling of array types when Specimin is manipulating strings that represent type names. All of the improvements are to account for the fact that types may have array brackets, which need to be stripped before e.g., looking up the type in a map. While I was editing these parts of the code, I made similar improvements so that generics are also similarly handled.
* some changes to whether fully-qualified or simple names are passed around were required. This is to deal with the fact that in the real example, there are actually two classes named `Method`: one imported, and another that is only referenced by its FQN. To get this right, we need to keep FQNs around a bit longer when reasoning about method call expressions. Related, I wrote some new logic for splitting up an FQN into the package and class names using the heuristic that packages are named with lowercase and classes are named with uppercase. This is Java convention but isn't actually guaranteed, so I added a note to the README about this.
* `TargetMethodFinderVisitor` did not directly add to the list of used classes the types of the parameters of used methods or constructors. When that is the only place that such a type is used, then they'd have been excluded. This rarely occurs in practice (usually, there is an in-scope variable with the same type at the call site that serves as the argument), but when it does it causes a crash. I added logic to both the handling of method calls and constructor calls to explicitly add the types of each parameter when a method is used.
* The logic for adding a used type in `TargetMethodFinderVisitor` didn't account for arrays, either, so I fixed that as well.
* `UnsolvedSymbolVisitor` did not consider used constructors as "potential used members" (only methods and fields), which caused some logic to not fire for the parameters of used constructors
* `UnsolvedSymbolVisitor`'s logic for creating synthetic classes was only triggered for `ClassOrInterfaceType`s, but it needs to be triggered for array and wildcard types, too (to create synthetic classes for the component types and bounds, respectively). I factored the logic out and then shared it between the `visit` methods for all of `ArrayType`, `WildcardType`, and `ClassOrInterfaceType`
* when `JavaTypeCorrect` found a problem that requires a change of type, we have logic in `UnsolvedSymbolVisitor` to look up the right package for the simple name that javac says that the type should be. However, javac can provide a more complex type than just a simple name, such as an array type or a type containing a generic. In these cases, before looking up the package associated with the type, we need to strip off the type modifiers (the array brackets, etc) before checking the class-to-package map


kelloggm:
I spent some time trying to break up this PR and I've decided that it's not worthwhile. @LoiNguyenCS if you can't make sense of it, let me know and I'll try a bit harder. I think it's ready for review.

README.md
```diff
@@ -42,10 +42,19 @@ Options may be specified in any order. When supplying repeatable options more th
...
 Here is a sample command to run the tool: `./gradlew run --args='--outputDirectory "tempDir" --root "src/test/resources/twofilesimple/input/" --targetFile "com/example/Foo.java" --targetFile "com/example/Baz.java" --targetMethod "com.example.Foo#bar()" --jarpath "path/to/jar/directory"'`
 
+# Important limitations and caveats
+
+The current implementation makes use of heuristics to distinguish simple from fully-qualified names
+at several points. In particular, it assumes that package and class names follow Java convention:
+* class names always begin with an uppercase letter
+* package names always begin with a lowercase letter
+
+Specimin will likely produce incorrect output if the input program does not follow this convention.
```
LoiNguyenCS, line 52: We should clarify that this limitation only applies to approximate slicing.

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
```diff
@@ -173,6 +173,22 @@ public Node visit(ImportDeclaration decl, Void p) {
...
+    if (isUsedMethodParameterType(classFullName)) {
+      return super.visit(decl, p);
+    }
+
+    decl.remove();
+    return decl;
+  }
+
+  /**
+   * Helper method to check if the given fully-qualified class name is used as a parameter by any of
```
LoiNguyenCS, line 185: I think there're missing words here. It should be "Helper method to check if the given fully-qualified class name is used as _**the type**_ of a parameter by any of..."

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
```diff
@@ -624,6 +625,13 @@ private void preserveMethodDecl(ResolvedMethodDeclaration decl) {
     catch (UnsolvedSymbolException e) {
       return;
     }
+
+    for (int i = 0; i < decl.getNumberOfParams(); ++i) {
```
LoiNguyenCS, line 629: We've already have the `visit(Parameter, Void)` to handle parameters of target members. I think we can skip this part.

LoiNguyenCS (approved):
LGTM. I only have a few minor suggestions for clarification and simplification.

src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
kelloggm: Without this change, the test `OtherPackageArrayTest` fails: the class `Type` isn't created, and the output isn't compilable.

I suspect that there might be another bug in our implementation of `visit(Parameter, Void)`. I don't see a real problem with having redundancy here, but we should also try to figure out why that method doesn't work as expected in this case.

PR #314, opened 2024-06-24
Correctly handle types that are more complex than arrays in #310

kelloggm:
This PR extends #310 to handle other, more complex types. In particular, the goal of this PR was to make sure that we use FQN for all kinds of "internal" types (component types of arrays, bounds of type variables, etc.) in a more principled way than we do in #310.

This should probably be reviewed before or concurrently with #310.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -3376,52 +3376,39 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
 
   /**
    * Lookup the fully-qualified names of each type in the given string, and replace the simple type
-   * names in the given string with their fully-qualified equivalents.
+   * names in the given string with their fully-qualified equivalents. Return the result.
    *
-   * @param javacType
-   * @return
+   * @param javacType a type from javac
+   * @return that same type, with simple names in the class-to-package map replaced with FQNs
    */
   private String lookupFQNs(String javacType) {
-    Node parsedJavac = StaticJavaParser.parseType(javacType);
-    parsedJavac.accept(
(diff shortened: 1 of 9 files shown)


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -3392,6 +3371,43 @@ public boolean updateTypeForSyntheticClasses(
...
+   * names in the given string with their fully-qualified equivalents. Return the result.
+   *
+   * @param javacType a type from javac
+   * @return that same type, with simple names in the class-to-package map replaced with FQNs
+   */
+  private String lookupFQNs(String javacType) {
+    // It's possible that the type could start with a new (synthetic) type variable's declaration.
+    // That won't be parseable as a type, so strip it first and then re-add it.
+    String typeVarDecl, rest;
+    if (javacType.startsWith("<")) {
```
LoiNguyenCS, line 3385: I am confused. How can a type start with `<`?

LoiNguyenCS (commented):
I am still a bit unsure about how this code works.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: Technically, it can't. However, we post-process what we get from javac to add synthetic type variable declarations to some return types (where they'll be placed in front of the method). So, for example, we might have something like `<SyntheticTypeVar> SyntheticTypeVar` as the input to this method; the first part is the declaration of the type variable, and the second part is a use of the type variable. (There's a test that demonstrates this - I think `LambdaBodyStaticUnsolved2Test`.)

I'll clarify the comment above.

LoiNguyenCS (approved):
LGTM

PR #319, opened 2024-07-12
Add Windows support for testers

theron-wang:
This PR creates the `check_differences.bat` file to substitute the lack of the `diff` command on Windows, allowing for full Windows support when running `./gradlew test`. It also ports `typecheck_test_outputs.sh` over to a `typecheck_test_outputs.bat`, which allows Windows users to call `./gradlew expectedTestOutputsMustCompile`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ task requireJavadoc(type: JavaExec) {
 task expectedTestOutputsMustCompile(type: Exec) {
     // TODO: should this task run in CI, or as part of the regular tests?
     if (System.getProperty('os.name').toLowerCase().startsWith('windows')) {
-        commandLine "cmd", "/c", "your_batch_file.bat"
+        commandLine "cmd", "/c", "typecheck_test_outputs.bat"
     } else {
         commandLine "sh", "typecheck_test_outputs.sh"
     }
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,11 @@ task requireJavadoc(type: JavaExec) {
 
 task expectedTestOutputsMustCompile(type: Exec) {
     // TODO: should this task run in CI, or as part of the regular tests?
-    commandLine "sh", "typecheck_test_outputs.sh"
+    if (System.getProperty('os.name').toLowerCase().startsWith('windows')) {
+        commandLine "cmd", "/c", "your_batch_file.bat"
+    } else {
+        commandLine "sh", "typecheck_test_outputs.sh"
+    }
 }
 
 tasks.compileJava {
--- a/check_differences.bat
+++ b/check_differences.bat
@@ -1,0 +1,122 @@
+@echo off
+
+rem On Windows, instead of using the diff command for SpeciminTestExecutor, we can use this
+rem command instead since diff is not included by default. This script outputs 0 if no differences
+rem are found, 1 if a difference in file structure or file content (all whitespace removed) is found
+
+goto :MAIN
+
+rem Function to remove all whitespace from a file
+:RemoveWhitespace
+    setlocal enabledelayedexpansion
+        set "content="
+        rem Simply calling (%~1) is not enough; in edge cases, empty lines containing delimiters (i.e. ;)
+        rem may be removed. Therefore, we need findstr to add line numbers so no line is empty.
+        for /F "tokens=* delims=" %%i in ('findstr /n "^" %~1') do (
+            set line=%%i
+            set line=!line:*:=!
+            if "!line!" NEQ "" set line=!line: =!
+            set content=!content!!line!
+        )
+        echo !content! >> %~2
+    endlocal
+exit /b 0
+
+:MAIN
+
+setlocal enabledelayedexpansion
+
+if "%1"=="" (
+    echo No path provided.
+    exit /b 1
+)
+if "%2"=="" (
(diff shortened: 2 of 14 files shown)


build.gradle
```diff
@@ -53,7 +53,11 @@ task requireJavadoc(type: JavaExec) {
 
 task expectedTestOutputsMustCompile(type: Exec) {
     // TODO: should this task run in CI, or as part of the regular tests?
-    commandLine "sh", "typecheck_test_outputs.sh"
+    if (System.getProperty('os.name').toLowerCase().startsWith('windows')) {
+        commandLine "cmd", "/c", "your_batch_file.bat"
```
kelloggm, line 57: I think this should be `typecheck_test_outputs.bat` rather than `your_batch_file.bat`.

Have you tried running this task and checking to make sure it does what you expect? I would think that this problem would show up immediately if you did.

check_differences.bat
```diff
@@ -0,0 +1,122 @@
+@echo off
+
+rem On Windows, instead of using the diff command for SpeciminTestExecutor, we can use this
+rem command instead since diff is not included by default. This script outputs 0 if no differences
+rem are found, 1 if a difference in file structure or file content (all whitespace removed) is found
```
kelloggm, line 5: Did you write this script yourself, or did you source it from somewhere? If the former, we need to add some tests for it (it's complex enough that I'm sure there are edge cases, and I'm not enough of a Windows batch file expert to be sure that this script is correct through inspection alone); if the latter, we should 1) give credit where credit is due, and 2) if possible, depend on wherever this script is sourced from directly, so that we'll receive bugfixes/updates if there are any.

kelloggm (changes requested):
In addition to the concerns I raised in my comments, I have one other large concern: Specimin's CI runs in Unix, so none of these scripts are tested anywhere. One advantage of having two different sets of scripts that are intended to accomplish the same purpose is that we can test them against each other (this is technically called _differential testing_), and I think it would be wise for us to do that.

build.gradle
theron-wang: Yes, I did, sorry. I accidentally deleted the file before I committed so I had to re-copy and paste the template code, but it looks like I forgot to change it again, thanks!

check_differences.bat
theron-wang: I wrote this script myself (with some Googling to help me with syntax). I can definitely add test cases to this, and definitely should. Thanks!

theron-wang:
@kelloggm Sorry, I’m a little confused as to how we can achieve this.

> One advantage of having two different sets of scripts that are intended to accomplish the same purpose is that we can test them against each other (this is technically called differential testing), and I think it would be wise for us to do that.

Since the two scripts run on different OSes, how can we compare their outputs? Would we run it on Windows and then include a check in the CI? Aside from that I’d definitely be happy to set it up.

kelloggm:
> Since the two scripts run on different OSes, how can we compare their outputs? Would we run it on Windows and then include a check in the CI? Aside from that I’d definitely be happy to set it up.

I think what I would do is add a second job in CI that runs on Windows. It would check that 1) the tests pass on Windows as expected, and 2) the results on Windows are the same as the ones on the Linux CI job (not sure how easy that is in GitHub Actions, so we could skip it if we have to)

theron-wang:
> I think what I would do is add a second job in CI that runs on Windows. 

Sounds great! I added some test cases in `check_differences/`. I also wrote another `test_runner.bat` file to run the tests; I understand that you are not too familiar with batch so we could run the tests another way as well. Please let me know if you want to add more test cases or change some of the current ones, and then we can work on setting up the second job.

Also, I didn't include any Java examples in the test case because running `./gradlew test` already verifies that everything works properly. However, we could include some Java-related test cases (right now, only 1/4 of the test cases are Java) if you think that's better.

kelloggm:
> Also, I didn't include any Java examples in the test case because running ./gradlew test already verifies that everything works properly

That seems fine to me.

kelloggm:
This looks good to me, but there is still one thing to do: add a GitHub actions job that runs the tests on Windows (and the tests you wrote for the Windows diff script, too). We can do that by creating a parallel GitHub actions [job](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow). I'd expect that test to 1) run on Windows, and 2) test three things:
* run `gradlew.bat test` to show that the tests pass on Windows
* run the tests for your diff script that you've included here
* run `gradlew.bat expectedTestOutputsMustCompile` to show that that script also works

kelloggm:
This looks good to me, but there is still one thing to do: add a GitHub actions job that runs the tests on Windows (and the tests you wrote for the Windows diff script, too). We can do that by creating a parallel GitHub actions [job](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow). I'd expect that test to 1) run on Windows, and 2) test three things:
* run `gradlew.bat test` to show that the tests pass on Windows
* run the tests for your diff script that you've included here
* run `gradlew.bat expectedTestOutputsMustCompile` to show that that script also works

theron-wang:
@kelloggm workflow created and runs successfully. Please let me know if you would like me to change anything else!

kelloggm (approved):
LGTM, thanks :)

PR #320, opened 2024-07-12
Move some static methods from PrunerVisitor to JavaParserUtil where they belong

kelloggm:
I noticed that these methods shouldn't be in `PrunerVisitor` while reviewing the changes in #302.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
+++ b/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
@@ -1,9 +1,14 @@
 package org.checkerframework.specimin;
 
 import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
+import com.github.javaparser.ast.body.EnumDeclaration;
 import com.github.javaparser.ast.body.TypeDeclaration;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import com.github.javaparser.resolution.types.ResolvedReferenceType;
+import org.checkerframework.checker.signature.qual.FullyQualifiedName;
+
+import java.util.Optional;
 
 /**
  * A class containing useful static functions using JavaParser.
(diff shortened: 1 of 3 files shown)


kelloggm:
@theron-wang I can't assign you to review this until you accept the write access invite that I've sent you, but once you've done that you should be the reviewer on this PR

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -68,4 +73,64 @@ public static ResolvedReferenceType classOrInterfaceTypeToResolvedReferenceType(
...
+     * a class, interface, or enum (i.e., something that would be a {@link
+     * javax.lang.model.element.TypeElement} in javac's internal model). This method will throw if no
+     * such element exists.
+     *
+     * @param node a node that is contained in a class-like structure
+     * @return the nearest enclosing class-like node
+     */
+    public static Node getEnclosingClassLike(Node node) {
+      Node parent = node.getParentNode().orElseThrow();
+      while (!(parent instanceof ClassOrInterfaceDeclaration || parent instanceof EnumDeclaration)) {
```
theron-wang, line 113: This method should also handle `AnnotationDeclaration` since annotations are considered an interface under `javax.lang.model.element.TypeElement`. This change is made in #302, (in `PrunerVisitor`, however), so we could make the change in that PR instead of here.

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -68,4 +73,64 @@ public static ResolvedReferenceType classOrInterfaceTypeToResolvedReferenceType(
...
+     *
+     * <p>This method will fail if it is called on a node that is not contained in a class or
+     * interface.
+     *
+     * @param node a node contained in a class or interface
+     * @return the fully-qualified name of the inner-most containing class or interface
+     */
+    @SuppressWarnings("signature") // result is a fully-qualified name or else this throws
+    public static @FullyQualifiedName String getEnclosingClassName(Node node) {
+      Node parent = getEnclosingClassLike(node);
```
theron-wang, line 89: If we decide to handle `AnnotationDeclaration`, in `getEnclosingClassLike` (see my other comment), we should account for that case in this method as well.

theron-wang (approved):
Looks good to me, aside from two minor comments!

kelloggm:
I've added the handling of annotations that you suggested.

PR #321, opened 2024-07-15
Refactor visitors to share state and simplify constructing one visitor from the previous one

kelloggm:
This doesn't need a review from both @LoiNguyenCS and @theron-wang, just from one of you (whoever gets to it first is fine).

I also removed some fields from the visitors that weren't ever used.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/PrunerVisitor.java
@@ -26,7 +26,6 @@ import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import com.github.javaparser.ast.type.PrimitiveType;
 import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.type.TypeParameter;
-import com.github.javaparser.ast.visitor.ModifierVisitor;
 import com.github.javaparser.ast.visitor.Visitable;
 import com.github.javaparser.resolution.UnsolvedSymbolException;
 import com.github.javaparser.resolution.declarations.ResolvedConstructorDeclaration;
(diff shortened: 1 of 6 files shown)


src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -477,7 +471,7 @@ private static void performMinimizationImpl(
         if (typesToChange.containsKey(simpleName)) {
           continue;
         }
-        if (!finder.getUsedTypeElement().contains(classFullyQualfiedName)) {
+        if (!finder.getUsedTypeElements().contains(classFullyQualfiedName)) {
```
theron-wang, line 474: We should fix the spelling of `classFullyQualfiedName` --> `classFullyQualifiedName`

src/main/java/org/checkerframework/specimin/PrunerVisitor.java
```diff
@@ -37,45 +36,18 @@
...
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * This visitor removes every member in the compilation unit that is not a member of its {@link
- * #methodsToLeaveUnchanged} set or {@link #membersToEmpty} set. It also deletes the bodies of all
- * methods and replaces them with "throw new Error();" or remove the initializers of fields
- * (minimized if the field is final) within the {@link #membersToEmpty} set.
+ * #targetMethods} set or {@link #usedMembers} set. It also deletes the bodies of all methods and
+ * replaces them with "throw new Error();" or remove the initializers of fields (minimized if the
+ * field is final) within the {@link #usedMembers} set.
```
theron-wang, line 48: `isUsedMethodParameterType()` documentation should also be updated to use `usedMembers` instead of `membersToEmpty`

theron-wang (approved):
Looks good to me, aside from a few small tweaks!

PR #324, opened 2024-07-18
Add `specimin-evaluation` tests to Windows CI

theron-wang:
This PR adds 4/20 of the `specimin-evaluation` test cases to the Windows CI to ensure that `main.py` successfully executes on various types of test cases on Windows.
- cf-1291: test typical integration test
- cf-577: download + unzip + rename/file access errors
- cf-691: download + unzip 2
- cf-4614: slightly different than expected CF outputs

This also fixes a silent error in `typecheck_test_outputs.bat` from #319 where not all test cases were being checked. The reason behind this was how I was handling `continue` in batch; I used `goto` within a `for` loop, which I did not know broke the script out of the loop context (hence stopping at the test case `methodref2`). 

The CI should pass once the companion PR njit-jerse/specimin-evaluation#3 is approved; right now we should expect at least a few of these tasks to fail.


theron-wang:
I accidentally indented one line for the Linux version of the script in `extract_and_rename` which is probably causing the error in the main CI test. For Windows, I am not entirely sure why it is not working (it works on my computer), but I will try to find out.

theron-wang:
@kelloggm Windows errors fixed: I forgot to set the `SPECIMIN` variable in the workflow. If you could check out njit-jerse/specimin-evaluation#4 which addresses the Ubuntu CI errors, that would be greatly appreciated. Sorry about these oversights!

kelloggm:
> Sorry about these oversights!

We should set up a CI check in the `specimin-evaluation` repository that just runs Specimin's CI on any proposed change, so that we don't make this mistake again in the future. @theron-wang would you be interested in setting that up?

theron-wang:
> would you be interested in setting that up?

Yes, of course. Should this run both the Ubuntu and Windows CIs? Thanks.

kelloggm:
> Yes, of course. Should this run both the Ubuntu and Windows CIs? Thanks.

Yes.

kelloggm (approved):

PR #325, opened 2024-07-19
Further refactoring of visitors to reduce duplicated tracking of state

kelloggm:


src/main/java/org/checkerframework/specimin/SpeciminStateVisitor.java
```diff
@@ -62,7 +90,11 @@ public SpeciminStateVisitor(
       Set<String> usedMembers,
       Set<String> usedTypeElements,
       Map<String, Path> existingClassesToFilePath) {
-    this.targetMethods = targetMethods;
+    this.targetMethods = new HashSet<>();
+    for (String methodSignature : targetMethods) {
+      // remove spaces
+      this.targetMethods.add(methodSignature.replaceAll("\\s", ""));
```
theron-wang, line 96: The AnnotationParameterTypesVisitor compares method signatures _with_ the spaces, so this could be causing some issues regarding that

src/main/java/org/checkerframework/specimin/SpeciminStateVisitor.java
```diff
@@ -91,4 +126,134 @@ public SpeciminStateVisitor(SpeciminStateVisitor previous) {
...
   }
+
+  /**
+   * Given a method declaration, this method return the declaration of that method without the
+   * return type and any possible annotation.
+   *
+   * @param methodDeclaration the method declaration to be used as input
+   * @return methodDeclaration without the return type and any possible annotation.
+   */
+  public static String removeMethodReturnTypeAndAnnotations(String methodDeclaration) {
```
theron-wang, line 137: does this belong in `JavaParserUtil`?

theron-wang (commented):
I was able to review partially, but not completely (I will finish later!)

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -271,17 +242,16 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
   public UnsolvedSymbolVisitor(
       String rootDirectory,
       Map<String, Path> existingClassesToFilePath,
-      List<String> targetMethodsSignatures,
-      List<String> targetFieldsSignature) {
+      Set<String> targetMethodsSignatures,
```
theron-wang, line 245: We may want to rename this to `targetMethodNames` since this set contains user-provided signatures like `com.example.Class#method()` rather than the typical `getQualifiedSignature()` signatures like `com.example.Class.method()`.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -219,39 +210,19 @@ public class UnsolvedSymbolVisitor extends ModifierVisitor<Void> {
...
       classToItsUnsolvedInterface = new HashMap<>();
 
-  /**
-   * List of signatures of target methods as specified by users. All signatures have spaces removed
-   * for ease of comparison.
-   */
-  private final Set<String> targetMethodsSignatures;
-
-  /** List of signatures of target fields as specified by users. */
-  private final Set<String> targetFieldsSignatures;
```
theron-wang, line 229: We could also remove `syntheticReturnTypes` and `syntheticTypes` since their values aren't used anywhere.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
theron-wang: We could also just change formats here (replace `#` with `.`), since this set does get passed down to other visitors which use `getQualifiedSignature()`.

theron-wang (changes requested):
Just a few comments regarding the new handling of the `targetMethods` set. The compile errors in the CI are probably caused by `AnnotationParameterTypesVisitor` not being able to preserve types because the target method signature formats are different.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: > We may want to rename this to targetMethodNames since this set contains user-provided signatures like com.example.Class#method() rather than the typical getQualifiedSignature() signatures like com.example.Class.method().

I don't think that this is correct: we never do such a conversion as far as I know (and there are many comparisons between the result of a call to `getQualifiedSignature` and the elements of `SpeciminStateVisitor#targetMethods`, which are the same as the contents of this set with the spaces removed.

theron-wang: Yep, I agree. However, a conversion (not really) occurs in `TargetMethodFinderVisitor` [here](https://github.com/njit-jerse/specimin/blob/9ea46fe9dd6cf55a2405680c14e02a9927f43428/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java#L274) so I believe the `targetMethods` set (since it is never cleared) contains both the user-specified signatures and `getQualifiedSignature` signatures past that visitor. This doesn't affect `UnsolvedSymbolVisitor` so it doesn't really matter. Sorry!

theron-wang (approved):
Looks good to me!

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
kelloggm: > a conversion (not really) occurs in TargetMethodFinderVisitor [here](https://github.com/njit-jerse/specimin/blob/9ea46fe9dd6cf55a2405680c14e02a9927f43428/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java#L274) so I believe the targetMethods set (since it is never cleared) contains both the user-specified signatures and getQualifiedSignature signatures past that visitor

I'll open another PR maybe to address this duplication - I don't think it needs to be mixed with this already large one. Thanks for pointing this out.

kelloggm: I looked into doing this, and it creates more problems than it solves. In particular, it's difficult to standardize on one internal format for the target methods set for three, competing reasons:
* `UnsolvedSymbolVisitor` needs to run before it is possible to resolve methods
* method signatures need to have fully-qualified parameter types to match the output of `ResolvedMethodDeclaration#getQualifiedSignature`
* we don't want to force users to fully qualify parameter types in the input.

Put another way, there are three main differences between the "getQualifiedSignature" format and the current input format:
* the current input format uses a `#` between the class and method name, but `getQualifiedSignature` uses a `.`
* the input format permits internal whitespace
* the input format permits simple names for parameter types

The first two of these differences can be overcome with string manipulation, but the last cannot. Maybe there is a better refactoring here, but I don't immediately see it - maybe creating a special class for "method signatures" that stores both the input format signature and the qualified signature (once it's available)? But that seems overly complex to me.

PR #327, opened 2024-07-22
Include methods for all implemented interfaces + annotation type-checker behavior preservation

theron-wang:
This PR addresses an issue I discovered while running Specimin on [NullAway issue #102](https://github.com/uber/NullAway/issues/102), where methods in an abstract class present in the parent interface of the implemented interface were not preserved. For example:

Expected:
```java
package com.example;

import java.util.Iterator;

public class Foo<E> {
    public Iterator<E> iterator() {
        return new AbstractLinkedIterator() {
            @Override
            E computeNext() {
                throw new Error();
            }
        };
    }

    public abstract class AbstractLinkedIterator implements Iterator1<E> {
        abstract E computeNext();

        @Override
        public E next() {
            throw new Error();
        }

        @Override
        public boolean hasNext() {
            throw new Error();
        }
    }

    public interface Iterator1<E> extends Iterator<E> {
    }
}
```

Actual:
```java
package com.example;

import java.util.Iterator;

public class Foo<E> {
    public Iterator<E> iterator() {
        return new AbstractLinkedIterator() {
            @Override
            E computeNext() {
                throw new Error();
            }
        };
    }

    public abstract class AbstractLinkedIterator implements Iterator1<E> {
        abstract E computeNext();
    }

    public interface Iterator1<E> extends Iterator<E> {
    }
}
```

The current behavior is to only look for method declarations in the direct parent interface `Iterator1`, but those methods are not present there. This causes them to be removed and thus compile errors ensue. However, I modified `MustImplementMethodsVisitor` to explore all interface ancestors, so the `next()` and `hasNext()` methods are still preserved. The test case can be found in `InterfaceChainTest`.

I also included a small tweak to annotation preservation in this PR; previously, we preserved JDK annotations by seeing if it started with `java.lang`, but I changed it to use `JavaLangUtils.inJdkPackage()` instead after discovering a conflict when looking at [NullAway issue #97](https://github.com/uber/NullAway/issues/97).

Thank you!

### DIFF
--- a/src/main/java/org/checkerframework/specimin/AnnotationParameterTypesVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/AnnotationParameterTypesVisitor.java
@@ -241,7 +241,7 @@ public class AnnotationParameterTypesVisitor extends SpeciminStateVisitor {
     try {
       String qualifiedName = anno.resolve().getQualifiedName();
       if (anno.resolve() instanceof ReflectionAnnotationDeclaration
-          && !qualifiedName.startsWith("java.lang")) {
+          && !JavaLangUtils.inJdkPackage(qualifiedName)) {
         // This usually means that JavaParser has resolved this through the import, but there
         // is no file/CompilationUnit behind it, so we should discard it to prevent compile errors
         anno.remove();
--- a/src/main/java/org/checkerframework/specimin/JavaLangUtils.java
+++ b/src/main/java/org/checkerframework/specimin/JavaLangUtils.java
@@ -287,9 +287,13 @@ public final class JavaLangUtils {
   public static boolean inJdkPackage(String qualifiedName) {
     // TODO: can we get a list of such packages from the JDK instead of using this relatively-coarse
     // heuristic?
+    if (qualifiedName.startsWith("javax.annotation")) {
+      return false;
+    }
+
     return qualifiedName.startsWith("java.")
-            || qualifiedName.startsWith("javax.")
-            || qualifiedName.startsWith("com.sun.")
-            || qualifiedName.startsWith("jdk.");
+        || qualifiedName.startsWith("javax.")
+        || qualifiedName.startsWith("com.sun.")
+        || qualifiedName.startsWith("jdk.");
   }
 }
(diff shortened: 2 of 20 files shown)


kelloggm (changes requested):
Overall LGTM, but tests are failing

kelloggm (approved):

theron-wang:
@kelloggm Seems like the failing integration test can't resolve `javax.annotations.Nonnull`; should that case run on Java 8 instead of Java 11? (I couldn't find a reference to Java 11 in the codebase's original `build.gradle` file, only [`1_8`](https://github.com/tahiat/calcite/blob/cf-3619/build.gradle.kts#L396))

kelloggm:
> Seems like the failing integration test can't resolve `javax.annotations.Nonnull` should that case run on Java 8 instead of Java 11?

I don't think the Java version should have anything to do with it, because `javax.annotations.Nonnull` (JSR 305) was [never included](https://stackoverflow.com/questions/2289694/what-is-the-status-of-jsr-305) in the JDK. It shouldn't count as a JDK name, and the code in this PR should create the appropriate annotation.

theron-wang:
@kelloggm Hmm, alright. The reason I wondered if Java 8 would work was because when I ran NullAway issue 97 in Java 8 (uses `javax.annotation.Nullable`), build.gradle didn't seem to need to reference anything else and it compiled fine. I will keep this as it was before then, comparing to `java.lang`. Should we remove `javax` from `JavaLangUtils.inJdkPackage` then, since it implies that `javax` is included in the jdk?

kelloggm:
> I will keep this as it was before then, comparing to java.lang. Should we remove javax from `JavaLangUtils.inJdkPackage` then, since it implies that javax is included in the jdk?

I don't think that's the right path forward - it'd be better to add exceptions to the current check in `JavaLangUtils.inJdkPackage` for the exception that we know about (`javax.annotations.*`) rather than exclude it entirely. And, it's a good idea for all of these checks to be unified!

The problem with `javax` as a package name is that it includes both Java extensions (the "x" in "javax") that _were_ incorporated into the JDK (e.g., all `javax.swing.*` packages...) and also a few that _were not_ incorporated into the JDK, like JSR 305 annotations.

theron-wang:
Thanks for clearing that up, I never knew how weird `javax` was haha. I've added a check for `javax.annotation` in `JavaLangUtil.inJdkPackage()` and also addressed the root cause of the issue I was facing in NullAway 97 by adding all `@Target` values to synthetic annotations as you mentioned previously in #302. Thanks!

PR #328, opened 2024-07-22
Fix Windows tests not running

theron-wang:
Windows diff test was previously not running due to an early return statement and an error in working directory (as seen in #327 when windows passed but ubuntu didn't).


kelloggm:
@theron-wang the tests are now failing on this branch :)

theron-wang:
Not sure why paths are getting their backslashes removed (don't know why :(, doesn't happen on my computer). Hopefully converting them to forward slashes and quoting them works. Sorry!

kelloggm (approved):
LGTM. I checked that the tests are actually running in CI :)

PR #329, opened 2024-07-23
Preserve `MustImplementMethodsVisitor` method type ancestors

theron-wang:
This PR addresses an issue I found in [NullAway issue #103](https://github.com/uber/NullAway/issues/103), where compile-time errors were caused by missing methods and types. The missing methods were caused by the following:

**1) Methods not preserved**

Foo extends (abstract) Bar extends [Some JDK interface]

If Bar contains an abstract method also present in the JDK interface, those implementations in Foo were removed while the abstract definition was maintained in Bar, causing compile-time errors. A check was added to ensure that all super abstract methods did not contain anything that needed to be preserved (i.e. when in Foo, instead of just checking Bar, we check Bar and the JDK interface).

**2. Classes not preserved**

If a non-target class needs its method implementations to be preserved, and some of its return/parameter types are solvable but were untracked so far by `usedTypeElements`, their ancestors were not preserved. For example, if a method visited by `MustImplementMethodsVisitor` is of type `PeekingIterator<E>` which extends `Iterator<E>`, the `extends Iterator<E>` portion is kept in the original definition but its import statement was removed, leading to compile-time errors. The new approach to this is, if a type element was not previously in `usedTypeElements`, all of its ancestors (interfaces and super classes) would also be marked for preservation.

A test case can be found in `Issue103Test`.

Thanks!

### DIFF
--- a/src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
@@ -1,14 +1,5 @@
 package org.checkerframework.specimin;
 
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
-import org.checkerframework.checker.nullness.qual.Nullable;
-
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
@@ -27,6 +18,13 @@ import com.github.javaparser.resolution.declarations.ResolvedParameterDeclaratio
 import com.github.javaparser.resolution.types.ResolvedReferenceType;
 import com.github.javaparser.resolution.types.ResolvedType;
 import com.github.javaparser.resolution.types.parametrization.ResolvedTypeParametersMap;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * If a used class includes methods that must be implemented (because it extends an abstract class
--- a/src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
@@ -123,6 +123,11 @@ public class MustImplementMethodsVisitor extends SpeciminStateVisitor {
       return true;
     }
 
+    if (overridden != null) {
+      return false;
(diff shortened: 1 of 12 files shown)


src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
```diff
@@ -63,15 +64,16 @@ public Visitable visit(MethodDeclaration method, Void p) {
     // implementing required methods from interfaces in the target code,
     // but unfortunately I think it's the best that we can do here. (@Override
     // is technically optional, but it is widely used.)
-    if (isPreservedAndAbstract(overridden)
+    if (ancestorMethodPreservedAndAbstract(method)
         || (overridden == null && overridesAnInterfaceMethod(method))) {
       ResolvedMethodDeclaration resolvedMethod = method.resolve();
-      Set<String> returnAndParamTypes = new HashSet<>();
+      Map<String, ResolvedType> returnAndParamTypes = new HashMap<>();
```
kelloggm, line 70: I think the values in this map also need to include thrown exceptional types. The new code exposes a problem caused by not doing that in CF-6060, which it looks like is why that integration test fails.

kelloggm (changes requested):
There are two failing integration tests:
* CF-6060 appears to be failing because exception types thrown by preserved interface methods aren't themselves preserved. I think this is an easy fix (see comment).
* CF-6030 is failing because a meta-annotation on a preserved annotation isn't being preserved. I'm not sure what that has to do with this PR, so you'll have to investigate @theron-wang

theron-wang:
> CF-6030 is failing because a meta-annotation on a preserved annotation isn't being preserved. I'm not sure what that has to do with this PR, so you'll have to investigate

Yep, I believe this was a side-effect of the approach I took with this PR; basically, a previously-unincluded method which had a parent abstract method which had another parent interface method in the JDK (it's strange, sorry) was now marked to be preserved, along with its annotation. The annotation definition referenced an inner annotation which was not included due to an issue with #302 which I fixed in this PR.

src/main/java/org/checkerframework/specimin/MustImplementMethodsVisitor.java
```diff
@@ -184,6 +265,7 @@ private boolean overridesAnInterfaceMethodImpl(
       for (String name : typeParametersMap.getNames()) {
         String interfaceViewpointName = name.substring(name.lastIndexOf('.') + 1);
         String localViewpointName = typeParametersMap.getValueBySignature(name).get().describe();
+        localViewpointName = Pattern.quote(localViewpointName);
```
kelloggm, line 268: A comment justifying why quoting this is necessary will be helpful later, so that we don't accidentally delete it.

kelloggm (approved):
LGTM other than one minor request. I don't need to review again before merging.

PR #330, opened 2024-07-24
CI specimin integration

NiharikaJamble:
This change will run a workflow to run accuracy check for changes made in specimin via a pull request or commit on main branch of project specimin

### DIFF
--- a/check_differences/check_differences.bat
+++ b/check_differences/check_differences.bat
@@ -1,5 +1,4 @@
 @echo off
-
 rem On Windows, instead of using the diff command for SpeciminTestExecutor, we can use this
 rem command instead since diff is not included by default. This script outputs 0 if no differences
 rem are found, 1 if a difference in file structure or file content (all whitespace removed) is found
@@ -9,18 +8,18 @@ goto :MAIN
 rem Function to remove all whitespace from a file
 :RemoveWhitespace
     setlocal enabledelayedexpansion
-        set "content="
-        rem Simply calling (%~1) is not enough; in edge cases, empty lines containing delimiters (i.e. ;)
-        rem may be removed. Therefore, we need findstr to add line numbers so no line is empty.
-        for /F "tokens=* delims=" %%i in ('findstr /n "^" %~1') do (
-            set line=%%i
-            set line=!line:*:=!
-            if "!line!" NEQ "" set line=!line: =!
-            set content=!content!!line!
-        )
-        echo !content! >> %~2
+    set "content="
+    rem Simply calling (%~1) is not enough; in edge cases, empty lines containing delimiters (i.e. ;)
+    rem may be removed. Therefore, we need findstr to add line numbers so no line is empty.
+    for /F "tokens=* delims=" %%i in ('findstr /n "^" "%~1"') do (
+        set "line=%%i"
+        set "line=!line:*:=!"
+        if "!line!" NEQ "" set "line=!line: =!"
+        set "content=!content!!line!"
+    )
+    echo !content! > "%~2"
     endlocal
-exit /b 0
+    exit /b 0
 
 :MAIN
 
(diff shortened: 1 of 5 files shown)


.github/workflows/specimin_evaluation_CI.yml
```diff
@@ -0,0 +1,204 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Debug - Show Current Branch
+        run: |
+          set -ex
+          echo "Current branch: $(git branch --show-current)"
```
kelloggm, line 26: This step (and other debug steps) should be removed before we merge this script.

.github/workflows/specimin_evaluation_CI.yml
```diff
@@ -0,0 +1,204 @@
...
+            cat /home/runner/work/specimin/specimin/CI_repository_list.csv
+          else
+            echo "File /home/runner/work/specimin/specimin/CI_repository_list.csv does not exist"
+            exit 1
+          fi
+
+      - name: Clone ASHE Project 
+        run: |
+          set -ex
+          git clone https://github.com/njit-jerse/ASHE_Automated-Software-Hardening-for-Entrypoints ASHE
```
kelloggm, line 69: Rather than cloning the ASHE repo directly, this should use the `git-clone-related` program available here: https://github.com/plume-lib/git-scripts/blob/main/git-clone-related instead. That program permits CI to work correctly when a developer needs to make changes that must be coordinated across multiple repositories.

.github/workflows/specimin_evaluation_CI.yml
```diff
@@ -0,0 +1,204 @@
...
+          
+      - name: Create ASHE Clone SPACE Directory
```
kelloggm, line 71: What is the purpose of this "Clone SPACE Directory"? It seems superfluous to me.

.github/workflows/specimin_evaluation_CI.yml
```diff
@@ -0,0 +1,204 @@
...
+            exit 1
+          fi
+
+      - name: Give write permissions to config.properties
+        run: |
+          set -ex
+          chmod +w ASHE/src/main/resources/config.properties
+          ls -l ASHE/src/main/resources/config.properties
+
+      - name: Update ASHE Config File to update SPECIMIN path
```
kelloggm, line 105: "Specimin" is a name, not an acronym. So, it should not be styled as "SPECIMIN", since it doesn't stand for anything.

.github/workflows/specimin_evaluation_CI.yml
```diff
@@ -0,0 +1,204 @@
...
+          grep 'Fully successful from minimization to compilation' $(pwd)/ASHE/logs/specimin_statistics.txt | awk '{print $NF}' > current_run_accuracy_percentage.txt
+          cat current_run_accuracy_percentage.txt
+
+      - name: Read and update evaluation accuracy
+        id: read_and_update
+        run: |
+          set -ex
+          current_accuracy=$(cat current_run_accuracy_percentage.txt)
+          echo "Current accuracy: $current_accuracy"
+          previous_run_accuracy=${{ secrets.LATEST_SPECIMIN_EVAL_PERCENTAGE }}
```
kelloggm, line 151: I'm concerned that this is a single global variable for all pull requests. Consider the following scenario in the future:
* on the `main` branch, the last CI run produced 90% accuracy
* from the `feature-1` branch, developer A opens a PR with 85% accuracy. This CI task fails, because 85 < 90 (all good so far!)
* from the `feature-2` branch, developer B opens another PR with 87% accuracy. Will this CI task pass? I think it will, because the latest run will be the run of `feature-1` that failed with 85% accuracy, and 87 > 85. 

So, I think that this really needs to only be updated if the current branch is `main`: otherwise, there could be confusion between PRs.

kelloggm (changes requested):
Looking over the log, I have one concern:
* from [here](https://github.com/njit-jerse/specimin/actions/runs/10101037185/job/27933709623#step:20:38), I see that there is a syntax error in your call to `bc`. That suggests that the result of this test is only correct coincidentally.

PR #331, opened 2024-07-25
Fix ability to target constructors with arguments

kelloggm:
Was broken by #325, I think. We should have had a test for this. Should fix the issue @theron-wang observed with na-323 and na-347, which both require a constructor with an argument to be targeted.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/TargetMethodFinderVisitor.java
@@ -196,6 +196,8 @@ public class TargetMethodFinderVisitor extends SpeciminStateVisitor {
     String constructorMethodAsString = method.getDeclarationAsString(false, false, false);
     // the methodName will be something like this: "com.example.Car#Car()"
     String methodName = this.currentClassQualifiedName + "#" + constructorMethodAsString;
+    // remove spaces
+    methodName = methodName.replaceAll("\\s", "");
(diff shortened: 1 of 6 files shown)


theron-wang (approved):
The exception is fixed! However, there are now compile-time errors (probably caused by the default values in final fields) but we can do that in another PR.

Here's the stack trace for future reference: 
```
C:\path\to\specimin-evaluation\ISSUES\na-347\output\otr4j\src\main\java\net\java\otr4j\messages\SignatureX.java:23: error: cannot assign a value to final variable longTermPublicKey
        this.longTermPublicKey = requireNonNull(ourLongTermPublicKey);
            ^
C:\path\to\specimin-evaluation\ISSUES\na-347\output\otr4j\src\main\java\net\java\otr4j\messages\SignatureX.java:24: error: cannot assign a value to final variable dhKeyID
        this.dhKeyID = requireNotEquals(0, ourKeyID);
            ^
C:\path\to\specimin-evaluation\ISSUES\na-347\output\otr4j\src\main\java\net\java\otr4j\messages\SignatureX.java:25: error: cannot assign a value to final variable signature
        this.signature = requireLengthExactly(DSA_SIGNATURE_LENGTH_BYTES, signature);
            ^
3 errors

FAILURE: Build failed with an exception.
```

kelloggm:
> The exception is fixed! However, there are now compile-time errors (probably caused by the default values in final fields) but we can do that in another PR.

Great. I will look into that other issue separately, and merge this one now.

PR #333, opened 2024-07-25
Don't add initializers to final fields assigned by a target constructor

kelloggm:
~~Should~~Does fix na-347 (confirmed locally that na-347 passes).


theron-wang (approved):
Looks good to me!

PR #334, opened 2024-07-25
Functional interfaces should not include a method if it's not used

theron-wang:
This PR fixes an issue present with na-791a where, even though a functional interface's method is not used, it is still being included in the output, causing compilation errors. Since the `@FunctionalInterface` annotation is removed, we should treat all interfaces equally.


kelloggm (changes requested):
Do you have a test case for this change? If not, we should wait until the NullAway bugs are running in CI, so that you can use na-791a as the test case to demonstrate that this fix works.

theron-wang:
Test added. Thanks!

kelloggm (approved):
Thanks for adding the test :)

PR #335, opened 2024-07-25
QoL improvement: typecheck just one test

kelloggm:
(on unix only, because I don't know how to write a bat script)


theron-wang (approved):
Looks good! I think it would be best if there was a note in `DEVELOPERS.md` on how to run this, though.

PR #336, opened 2024-07-26
Parent synthetic classes should preserve method exceptions

theron-wang:
The issue with na-389 was that a child method had a `throws` clause, but its synthetic, parent class definition did not keep these exceptions, leading to a compilation error. This PR addresses this by ensuring that `UnsolvedMethod`s generate with the original exceptions.

A test case can be found in `InheritMethodExceptionTest`.


kelloggm (approved):
Nice catch! Fix LGTM.

PR #337, opened 2024-07-26
Make synthetic classes used in a try with resources statement implement AutoCloseable

kelloggm:
Based on a problem I noticed while investigating why Specimin wasn't handling na-176 properly. There are probably other issues with that bug, but this one seemed like a good place to start.

### DIFF
--- a/src/test/resources/trywithresources/input/com/example/Simple.java
+++ b/src/test/resources/trywithresources/input/com/example/Simple.java
@@ -1,12 +1,23 @@
 package com.example;
 
 import org.example.Resource;
+import org.example.OtherResource;
+import org.example.ThirdResource;
 
 class Simple {
+
+    private ThirdResource r;
+
     // Target method.
-    void bar() throws Exception {
+    void bar(OtherResource o) throws Exception {
         try (Resource r = new Resource()) {
             // do something
+        }
+        try (o) {
+            // do something else
+        }
+        try (r) {
+
         }
     }
 }
--- a/src/test/resources/trywithresources/input/com/example/Simple.java
+++ b/src/test/resources/trywithresources/input/com/example/Simple.java
@@ -4,7 +4,7 @@ import org.example.Resource;
 
 class Simple {
     // Target method.
-    void bar() {
+    void bar() throws Exception {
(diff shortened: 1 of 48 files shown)


src/test/resources/trywithresources/input/com/example/Simple.java
```diff
@@ -0,0 +1,23 @@
+package com.example;
+
+import org.example.Resource;
+import org.example.OtherResource;
+import org.example.ThirdResource;
+
+class Simple {
+
+    private ThirdResource r;
```
theron-wang, line 9: For this test case, we might want to change this to `private final ThirdResource r = new ThirdResource()` to prevent the compilation error we see in the CI. Not sure how Specimin will handle this (if it will remove the initializer or not) but we could force Specimin to preserve all used fields if they are used with try-with-resources.

theron-wang (changes requested):
Besides one comment regarding the test case's behavior, the code changes look good to me. By the way, before I realized you were working on this issue, I was looking into this issue and created a minimal reproducible test case for the `IllegalStateException`, if it helps:

```java
package test;

public class Foo {
    public void foo() {
        Baz.test();
    }

    public void bar() {
        Baz<String> baz;
    }
}
```

src/test/resources/trywithresources/input/com/example/Simple.java
kelloggm: I changed it to just initialize it to null, which is what Specimin would do anyway. The actual value of the resource doesn't matter here, since we're not going to run the code - javac just checks that it's final.

kelloggm:
> Besides one comment regarding the test case's behavior, the code changes look good to me. By the way, before I realized you were working on this issue, I was looking into this issue and created a minimal reproducible test case for the `IllegalStateException`, if it helps

Definitely! I actually haven't investigated that part of the crash yet at all - I looked at the code that was causing the problem and immediately noticed the try-with-resources and realized that Specimin didn't handle that case at all, so I just started working on fixing that. I'll look into this problem next, starting with your test case.

theron-wang (approved):
Looks good to me!

PR #338, opened 2024-07-26
Don't crash when encountering an erased type in a static use context

kelloggm:
Based on the test case provided by @theron-wang.


theron-wang (approved):
Looks good to me!

PR #340, opened 2024-07-26
Re-add support for target fields

theron-wang:
While looking into na-705, I found that Specimin ignored target fields introduced in #247, possibly when we were refactoring. The original test case `TargetField` was also not included in the original PR, so I included the test for the CI. This PR doesn't fix the issue present in na-705, however.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
+++ b/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
@@ -354,8 +354,8 @@ public class SpeciminRunner {
     // what specifications they use, and the second phase takes that information
     // and removes all non-used code.
 
-    TargetMethodFinderVisitor finder =
-        new TargetMethodFinderVisitor(enumVisitor, nonPrimaryClassesToPrimaryClass);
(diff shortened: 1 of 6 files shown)


src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -365,8 +365,16 @@ private static void performMinimizationImpl(
...
       throw new RuntimeException(
           "Specimin could not locate the following target methods in the target files:\n"
-              + unfoundMethodsTable(unfoundMethods));
+              + unfoundMembersTable(unfoundMethods, true));
     }
+
+    Map<String, Set<String>> unfoundFields = finder.getUnfoundFields();
+    if (!unfoundFields.isEmpty()) {
+      throw new RuntimeException(
+          "Specimin could not locate the following target methods in the target files:\n"
```
kelloggm, line 374: This should say "target fields"

kelloggm (approved):
LGTM, thanks for noticing this bug - it was quite subtle.

kelloggm (approved):

PR #341, opened 2024-07-29
Fully qualify types in generic synthetic method return types

theron-wang:
Previously, when a synthetic method with a generic return type had a type parameter in the same package as the target method, it would not keep its fully qualified name in its synthetic definition, leading to compilation errors seen in na-791b.

For example, `ImmutableSet.copyOf` returned an `ImmutableSet<Feature>`, but in its synthetic definition, `copyOf` returned `com.google.common.collect.ImmutableSet<Feature>` instead of `com.google.common.collect.ImmutableSet<com.github.benmanes.caffeine.cache.Feature>`.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -3667,8 +3667,10 @@ public class UnsolvedSymbolVisitor extends SpeciminStateVisitor {
       // If it's not imported, it's probably in the same package
       // TODO: handle already fully qualified generic type arguments. Right now JavaTypeCorrect
       // only outputs the simple class name, so there is no way to get its fully qualified name
-      // here without ambiguity (i.e. org.example.Foo and com.example.Foo are different signatures)
+      // here without ambiguity (i.e. org.example.Foo and com.example.Foo have different signatures)
       // with the same simple class name
+      // Check MethodReturnFullyQualifiedGenericTest for more details; the current return type in
+      // Bar.java is com.example.InOtherPackage2 rather than com.foo.InOtherPackage2 (expected)
       fullyQualifiedName.append(currentPackage).append(".").append(typeArgument.toString());
     }
   }
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -3627,6 +3627,9 @@ public class UnsolvedSymbolVisitor extends SpeciminStateVisitor {
         fullyQualifiedName.append("? extends ");
         lookupTypeArgumentFQN(fullyQualifiedName, asWildcardType.getExtendedType().get());
       }
+    } else if (isAClassPath(erased)) {
+      // If it's already a fully qualified name, don't do anything
+      fullyQualifiedName.append(typeArgument.asString());
     } else {
       // If it's not imported, it's probably in the same package
       fullyQualifiedName.append(currentPackage).append(".").append(typeArgument.toString());
--- a/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
@@ -3665,6 +3665,10 @@ public class UnsolvedSymbolVisitor extends SpeciminStateVisitor {
       fullyQualifiedName.append(typeArgument.asString());
     } else {
       // If it's not imported, it's probably in the same package
+      // TODO: handle already fully qualified generic type arguments. Right now JavaTypeCorrect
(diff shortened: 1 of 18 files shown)


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -3565,18 +3565,74 @@ private String lookupFQNs(String javacType) {
...
+      if (asWildcardType.getSuperType().isPresent()) {
+        fullyQualifiedName.append("? super ");
+        lookupTypeArgumentFQN(fullyQualifiedName, asWildcardType.getSuperType().get());
+      } else if (asWildcardType.getExtendedType().isPresent()) {
+        fullyQualifiedName.append("? extends ");
+        lookupTypeArgumentFQN(fullyQualifiedName, asWildcardType.getExtendedType().get());
+      }
+    } else {
+      // If it's not imported, it's probably in the same package
+      fullyQualifiedName.append(currentPackage).append(".").append(typeArgument.toString());
```
kelloggm, line 3632: What if the type argument was already fully-qualified in the input. For example, what if we added the following method to your test case:
```
    public Bar<com.foo.InOtherPackage2> foo() {
        return Bar.getOtherPackage2();
    }
```

I think the current implementation will hit this line and append the current package to the front of the name, resulting in something like `"com.example.com.foo.InOtherPackage2"`. This would probably be easy enough to fix by just checking if the name is fully-qualified already before you call this method.

Regardless, please add the test case.

kelloggm (changes requested):
Overall looks great, but I have one concern that needs to be addressed.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
theron-wang: @kelloggm I looked into this, and unfortunately it's more complicated than that. The `lookupFQN()` method, in this case, is called after `JavaTypeCorrect`, which gets rid of the fully qualified name. Since it's not imported, I can't resolve the fully qualified name through `classAndPackageMap`. I also don't think we can simply add the types to `classAndPackageMap` in `visit(ClassOrInterfaceType)`, since there could be multiple classes with the same simple name (i.e. `com.example.Foo` and `com.other.Foo`). Do you have any ideas on how to solve this?

kelloggm: In the general case I think you're right that there's no unambiguous way to handle this.

I don't think handling this case correctly is critical at the moment. Let's move the `InOtherPackage2` version of the test into a separate test runner, allow that test runner to fail intentionally, and then add a TODO in `lookupFQN` with your notes about this problem. Then we can merge this PR, which does improve the current situation even if it is imperfect.

Then, we can come back to this problem later when we're not worried about a deadline at the end of the week :)

theron-wang:
I added another test case, as you suggested. For the expected output I put the incorrect one, but I changed the check compilation scripts to ignore the compilation error for this case.

src/test/java/org/checkerframework/specimin/MethodReturnFullyQualifiedGenericTest.java
```diff
@@ -0,0 +1,20 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.Test;
+
+/**
+ * This test checks that Specimin includes the fully qualified class name in generics when
+ * generating a synthetic method. For example, if a return type is Bar<com.foo.Foo>, the actual
+ * synthetic return type should be com.qualified.Bar<com.foo.Foo>. At this moment, we should expect
+ * this test case to fail.
```
kelloggm, line 10: Rather than "we should expect this test to fail", I would phrase this as "this test encodes Specimin's current behavior, which doesn't produce compilable output. If you break this test, it might not be a bad thing."

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -3598,18 +3598,81 @@ private String lookupFQNs(String javacType) {
...
+      }
+    } else if (isAClassPath(erased)) {
+      // If it's already a fully qualified name, don't do anything
+      fullyQualifiedName.append(typeArgument.asString());
+    } else {
+      // If it's not imported, it's probably in the same package
+      // TODO: handle already fully qualified generic type arguments. Right now JavaTypeCorrect
+      // only outputs the simple class name, so there is no way to get its fully qualified name
+      // here without ambiguity (i.e. org.example.Foo and com.example.Foo are different signatures)
+      // with the same simple class name
```
kelloggm, line 3671: Add a reference to the test that demonstrates the "bad" behavior to this comment.

kelloggm (approved):
I don't need to review again after you make the requested changes to the wording of a few comments. Nice work :)

theron-wang:
Thanks :) Just changed the comments.

PR #343, opened 2024-07-29
NullAway modularity model skeleton...

kelloggm:
...and one simple, demonstrative use. This might be sufficient to fix at least one of the NullAway bugs (NullAway considers the absence of an assignment in a constructor when analyzing the constructor). However, I'll need to make some associated changes to the evaluation scripts so that the right modularity model gets selected for the NullAway bugs.


src/main/java/org/checkerframework/specimin/TargetMemberFinderVisitor.java
```diff
@@ -246,6 +246,16 @@ public Visitable visit(ConstructorDeclaration method, Void p) {
           resolvedMethod.getPackageName() + "." + resolvedMethod.getClassName(),
           usedTypeElements,
           nonPrimaryClassesToPrimaryClass);
+      if (modularityModel.preserveAllFieldsIfTargetIsConstructor()) {
+        // This cast is safe, because a constructor must be contained in a class declaration.
+        ClassOrInterfaceDeclaration thisClass =
+            (ClassOrInterfaceDeclaration) JavaParserUtil.getEnclosingClassLike(method);
+        for (FieldDeclaration field : thisClass.getFields()) {
+          for (VariableDeclarator variable : field.getVariables()) {
+            usedMembers.add(currentClassQualifiedName + "#" + variable.getNameAsString());
```
theron-wang, line 255: Do we need to make sure the field types are included in `usedTypeElements`? Currently `visit(VariableDeclarator)` only checks if it's in the `targetFields` set, not `usedMembers` so these types may not be included in the output

theron-wang (approved):
Mostly looks good, just one question.

src/main/java/org/checkerframework/specimin/TargetMemberFinderVisitor.java
kelloggm: Good catch! It's actually trickier than that - I also needed to trigger synthetic class generation for the types of the fields in `UnsolvedSymbolVisitor`. I've added a test for this case.

PR #344, opened 2024-07-29
include `java/` files in output if they are targeted

theron-wang:
In na-97, where we target a `java/` file, the targeted file (`java/util/atomic/concurrent/Striped64.java`, I believe) was being excluded from the output because of these lines in `SpeciminRunner`:

```java
if (target.getKey().startsWith("java/")) {
    continue;
}
```

The reason it worked on Windows was probably because Windows uses backslashes, and it didn't work on Unix because of the forward slash comparison here. I changed it so that the target file is preserved in the output no matter what, even if it starts with `java/` (or `java\`). This should make na-97 work on both OSes.

I didn't include a test case for this, since NullAway bugs are now in the evaluation script, but if you'd like, I could add one. Thanks!

### DIFF
--- a/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
+++ b/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
@@ -461,10 +461,20 @@ public class SpeciminRunner {
 
     // cache to avoid called Files.createDirectories repeatedly with the same arguments
     Set<Path> createdDirectories = new HashSet<>();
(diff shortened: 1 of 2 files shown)


kelloggm:
CI should fail on this change, since we expect it to fix na-97 but it doesn't change the preservation status file. That will be a good test of the updated CI.

kelloggm (approved):

theron-wang:
Whoops, I think I have to merge the latest changes first.

PR #345, opened 2024-07-30
Call the correct generic "solve a type" method in UnsolvedSymbolVisitor when using the NA modularity model

kelloggm:
This fixes NA-347 (with the modularity model changes in specimin-evaluation#9)


theron-wang (approved):
LGTM!

PR #346, opened 2024-07-31
Remove extra `ElementType`s in all synthetic annotations

theron-wang:
This change allows na-705 to be reproducible. I discovered this one change was able to reproduce the error when running Specimin with `--jarPath path/to/checker-qual.jar`. ~~Since `@Retention(RetentionPolicy.RUNTIME)` is the most generous, it should work for all cases (i.e. `CLASS` and `SOURCE`).~~ (see comment below)

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnsolvedClassOrInterface.java
+++ b/src/main/java/org/checkerframework/specimin/UnsolvedClassOrInterface.java
@@ -471,18 +471,9 @@ public class UnsolvedClassOrInterface {
       sb.append(
           "@java.lang.annotation.Target({ \n"
               + "\tjava.lang.annotation.ElementType.TYPE, \n"
-              + "\tjava.lang.annotation.ElementType.FIELD, \n"
               + "\tjava.lang.annotation.ElementType.METHOD, \n"
-              + "\tjava.lang.annotation.ElementType.PARAMETER, \n"
-              + "\tjava.lang.annotation.ElementType.CONSTRUCTOR, \n"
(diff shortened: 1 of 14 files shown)


kelloggm (approved):
Nice find!

theron-wang:
Never mind, it looks like `@Retention(RetentionPolicy.RUNTIME)` was a red herring (the checker jar also only used `@Target(TYPE, TYPE_USE)`. The issue for this was an over-correction in #327, where I included all `ElementType`s in `@Target`. In fact, the only change I needed to make was to add `ElementType.Type`. I believe that the extra definition of `ElementType.Field`, in this case, caused the annotation to be incorrectly processed by NullAway.

kelloggm:
@theron-wang I think this is still off - now not only is na-705 failing, but na-97 is, too.

theron-wang:
@kelloggm Looks like `ElementType.FIELD` is needed to reproduce na-97, but its presence in na-705 prevents the NullAway from reproducing the bug. I tried removing `@Target` altogether for both examples, but na-705 still is not being reproduced. I don't think there is way to differentiate between the two cases (we can't check by context either, since both na-97 and na-705 `@Nullable` annotations apply to fields), so the best way to still reproduce na-705 is to run it with the checker-qual.jar. What are your thoughts? We could just close this PR and add a checker-qual dependency in specimin-evaluation for na-705, but that is an unsatisfying solution.

theron-wang:
Closed as per njit-jerse/specimin-evaluation#12.

PR #351, opened 2024-08-08
CI_for_Specimin

NiharikaJamble:
This CI evaluates accuracy of specimin for any push/pull events on main.
It will check file CI_Latest_run_percentage.txt for the latest accuracy percentage of specimin. 

The CI will pass if the current accuracy( accuracy calculated for current change) matches the existing accuracy of main branch, if not the CI will result in failure.

In case the calculated accuracy doesn't match, and is higher than the value present in the file, it needs to be manually updated to the txt file.


.github/workflows/specimin_evaluation_CI.yml
```diff
@@ -0,0 +1,132 @@
+name: specimin_evaluation_CI
```
kelloggm, line 1: Can we change the name of this job to be more descriptive of what it actually does? Maybe something like "check_compilable_percentage"?

There is already another CI job called "specimin-evaluation", so I'm trying to avoid confusion.

kelloggm (approved):
Other than the name, this looks good to me.

NiharikaJamble:
Ive made the necessary changes

kelloggm:
@NiharikaJamble the `.txt` file that you added with the accuracy isn't correct - see the job's failure. This is good validation that this is working as expected; once you update that file to be correct and CI passes, this will merge.

PR #352, opened 2024-08-08
Handle method references

theron-wang:
Method references should be solvable, but only if their parameter types match the type arguments in its corresponding functional interface (for example, `Function<?, ?>` doesn't match `String foo(int x)` but `Function<Integer, ?>` does). The test case (a bit of a large one) is in `SyntheticMethodRefsTest`. Thanks!


src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -3562,6 +3731,137 @@ public boolean updateTypeForSyntheticClasses(
...
+          for (String parameter : getReferenceTypesFromCommaSeparatedString(parametersAsString)) {
+            parameters.add(lookupFQNs(parameter.trim()));
+          }
+
+          // 1st phase: Assume non-void until javac says it is void
+          String fixed =
+              resolveFunctionalInterfaceWithFullyQualifiedParameters(
+                  parameters, false, currentPackage);
+
+          for (int argument : argumentsToFix) {
```
kelloggm, line 3769: This for loop is the reason you're getting a warning from the interning checker. The type of `argument` should be `Integer`, not `int` (primitives can't be stored in iterables, so this is hiding a boxing conversion, which is undesirable).

You can call `getValue()` on the `Integer` if you need the primitive.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -3562,6 +3731,137 @@ public boolean updateTypeForSyntheticClasses(
...
+
+  /**
+   * Given a String of types (separated by commas), return a List of these types, with any
+   * primitives converted to their object counterparts. Use this instead of {@code .split(", ")} to
+   * properly handle generics.
+   *
+   * @param commaSeparatedTypes A string of comma separated types
+   * @return a list of strings representing the types in commaSeparatedTypes
+   */
+  private List<String> getReferenceTypesFromCommaSeparatedString(String commaSeparatedTypes) {
```
kelloggm, line 3833: Is there a reason this can't be a static utility method?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -3562,6 +3731,137 @@ public boolean updateTypeForSyntheticClasses(
...
+    if (commaSeparatedTypes == null || commaSeparatedTypes.isBlank()) {
+      return new ArrayList<>();
```
kelloggm, line 3835: Use `Collections.EMPTY_LIST` to avoid an unnecessary allocation here.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -3562,6 +3731,137 @@ public boolean updateTypeForSyntheticClasses(
...
+    for (int i = 0; i < tokens.length; i++) {
+      if (JavaLangUtils.isPrimitive(tokens[i].trim())) {
+        tokens[i] = JavaLangUtils.getPrimitiveAsObject(tokens[i].trim());
+      }
+    }
+
+    // Parse as a generic type, then get the type arguments
+    // This way we can properly differentiate between commas within type arguments
+    // versus actual commas in javac error messages
+    Type parsed = StaticJavaParser.parseType("ToParse<" + String.join(", ", tokens) + ">");
```
kelloggm, line 3852: This is a neat trick! Nicely done :)

kelloggm (changes requested):
Overall this is great work, thanks! I have a few comments and I'd like to review again after they're addressed, but I think it'll be easy to merge this.

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
theron-wang: Ah, that makes sense. I was wondering where that warning was coming from; thanks!

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
theron-wang: Thanks!

src/main/java/org/checkerframework/specimin/JavaLangUtils.java
```diff
@@ -49,6 +51,9 @@ private JavaLangUtils() {
    */
   private static final Set<String> knownFinalJdkTypes = new HashSet<>();
 
+  /** A list of map of primitive names (int, short, etc.) to their object representations */
```
kelloggm, line 54: Typo: "a list of map of" -> "a map of"

kelloggm (approved):
LGTM

PR #354, opened 2024-08-13
Remove unused imports

theron-wang:
Removes unused imports in all output files.

Take, for example, `ManyImports` in the included test case.
Previously:

```java
package com.example;

import java.util.List;
import java.io.*;
import static java.util.Map.*;
import static java.lang.Math.sqrt;
import static java.lang.Math.PI;

public class ManyImports {
}
```

Now:

```java
package com.example;

public class ManyImports {
}
```

Even if the import type is a used type element, it should still be removed if the current, minimized file does not contain any instances.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/UnusedImportRemoverVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/UnusedImportRemoverVisitor.java
@@ -4,6 +4,7 @@ import com.github.javaparser.ast.ImportDeclaration;
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.PackageDeclaration;
 import com.github.javaparser.ast.expr.AnnotationExpr;
+import com.github.javaparser.ast.expr.Expression;
 import com.github.javaparser.ast.expr.FieldAccessExpr;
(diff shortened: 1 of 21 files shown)


theron-wang:
@kelloggm No rush on the review, just whenever you're back and have time!

src/main/java/org/checkerframework/specimin/UnusedImportRemoverVisitor.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    String fullyQualified = JavaParserUtil.erase(scope.calculateResolvedType().describe());
+
+    if (!fullyQualified.contains(".")) {
+      // If there is no ., it is not a class (i.e. this.values.length)
+      return;
+    }
+
+    String wildcard = fullyQualified.substring(0, fullyQualified.lastIndexOf('.')) + ".*";
```
kelloggm, line 254: I would extract this call to substring into a helper method, since it's repeated elsewhere (e.g., at line 130 above and just below at line 267).

kelloggm (approved):
Overall this is a great set of changes. Once my minor comments are addressed, this can be merged without another review.

theron-wang:
Changes made; thanks.

PR #355, opened 2024-08-15
Remove unused `ElementType`s in `@Target`

theron-wang:
Removes unused `ElementType`s in `@Target` annotations. The current approach is to only remove `ElementType`s, not add them, so we don't accidentally make a non `TYPE_USE` annotation into a `TYPE_USE` annotation, for example.

I also addressed two small bugs I encountered:
- When all the wildcard import are JDK imports, synthetic classes should be generated in the current package instead of the first wildcard import package
- Don't throw an exception when an annotation is applied on a package declaration

Instead of creating a new test case for this, I just modified one of the existing ones (`SyntheticAnnotationTargetTest`), since it's purpose already seemed to fit this PR.

Thanks!

### DIFF
--- a/src/main/java/org/checkerframework/specimin/AnnotationTargetRemoverVisitor.java
+++ b/src/main/java/org/checkerframework/specimin/AnnotationTargetRemoverVisitor.java
@@ -25,9 +25,12 @@ import com.github.javaparser.ast.type.TypeParameter;
 import com.github.javaparser.ast.visitor.ModifierVisitor;
 import com.github.javaparser.ast.visitor.Visitable;
 import java.lang.annotation.ElementType;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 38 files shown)


src/main/java/org/checkerframework/specimin/AnnotationTargetRemoverVisitor.java
```diff
@@ -0,0 +1,242 @@
...
+      AnnotationExpr targetAnnotation = pair.getValue().getAnnotationByName("Target").orElse(null);
+
+      if (targetAnnotation == null) {
+        // This is most likely an existing definition. If there is no @Target annotation,
+        // we shouldn't add it
+        continue;
+      }
+
+      boolean useFullyQualified =
+          targetAnnotation.getNameAsString().equals("java.lang.annotation.Target");
```
kelloggm, line 62: Since you've used the string `"java.lang.annotation.Target"` several times in this PR, it probably makes sense to pull it into a constant static field, just to avoid the possibility of typos, etc.

kelloggm (approved):
LGTM

kelloggm:
@theron-wang can you resolve the merge conflict? It should be easy: you just need to decide whether the code in this PR or the code in #354 (which I just merged) should run first.

Issue #368, opened 2024-10-24
Deletes input source files under some circumstances


gmerr3:
File 1(A.java):
```
import java.io.IOException;
import java.net.InetAddress;
import java.net.InetSocketAddress;
import java.nio.channels.AsynchronousSocketChannel;

public class A{
	public AsynchronousSocketChannel client;
	public A(AsynchronousSocketChannel client){
		this.client=client;
	}
	public InetAddress test(){
		try {
			if(client.getRemoteAddress() instanceof InetSocketAddress isa)
				return isa.getAddress();
		} catch (IOException e) {
			
		}
		return null;
	}
}
```
File 2(B.java):
```

import java.io.IOException;
import java.net.InetAddress;

public class B extends A {

    @Override
    public InetAddress test(){
        return super.test();
    }
}
```
Command:
```
gradlew run  --args="--outputDirectory \"/tmp/specimin328479325\" --root \"/path/to/src/main/java\" --targetFile \"B.java\" --targetMethod \"B#test()\""
```
The source file A.java is deleted when running the above command.
It is deleted at this point:
```
at org.checkerframework.specimin.UnsolvedSymbolVisitor.deleteOldSyntheticClass(UnsolvedSymbolVisitor.java:3201)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.updateSyntheticSourceCode(UnsolvedSymbolVisitor.java:3182)
at org.checkerframework.specimin.SpeciminRunner.performMinimizationImpl(SpeciminRunner.java:303)
at org.checkerframework.specimin.SpeciminRunner.performMinimization(SpeciminRunner.java:163)
at org.checkerframework.specimin.SpeciminRunner.main(SpeciminRunner.java:90)
```

gmerr3:
fixed in #371

Issue #370, opened 2024-10-31
empty output for classes in default package


gmerr3:
Test case - same as #361 but remove references to TimeUnit.
Output: file is created but is blank. Does not happen if file is put in a package.

gmerr3:
fixed in #371

Issue #361, opened 2024-09-20
Crashes with "qualifiedNameToFilePath only works for classes in the original directory" on java.util.concurrent.TimeUnit


gmerr3:
Command:
```
./gradlew run --args="--outputDirectory \"./tmp\" --root \"$(pwd)\" --targetFile \"Scheduler.java\" --targetMethod \"Scheduler#schedule(Runnable, int)\"" --info --stacktrace
```
Code(Scheduler.java):
```
import static java.util.concurrent.Executors.newSingleThreadScheduledExecutor;

import java.util.concurrent.ScheduledExecutorService;
import java.util.concurrent.TimeUnit;

public class Scheduler{
	public static final ScheduledExecutorService s = newSingleThreadScheduledExecutor(x -> new Thread(x, "scheduler thread"));

	private Scheduler() {

	}

	public static void schedule(Runnable task, int delay) {
		s.schedule(task, delay, TimeUnit.MILLISECONDS);
	}
}
```

Stack trace:
```
Exception in thread "main" java.lang.RuntimeException: qualifiedNameToFilePath only works for classes in the original directory
at org.checkerframework.specimin.UnsolvedSymbolVisitor.qualifiedNameToFilePath(UnsolvedSymbolVisitor.java:1567)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.updatedAddedTargetFilesForPotentialEnum(UnsolvedSymbolVisitor.java:1591)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:1073)
        at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:105)
        at com.github.javaparser.ast.expr.FieldAccessExpr.accept(FieldAccessExpr.java:95)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1111)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:52)
        at com.github.javaparser.ast.NodeList.accept(NodeList.java:275)
        at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1149)
        at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:642)
[...]

PR #371, opened 2024-11-05
Fix issues #368, #370, #361

gmerr3:
Fix issue #368 , issue #370 , and issue #361 .
Tests have been added.
One existing test(Issue103) fails, but it was already failing in the main branch.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
+++ b/src/main/java/org/checkerframework/specimin/SpeciminRunner.java
@@ -1,6 +1,7 @@
 package org.checkerframework.specimin;
 
 import com.github.javaparser.ParseProblemException;
+import com.github.javaparser.ParseResult;
 import com.github.javaparser.ParserConfiguration;
 import com.github.javaparser.StaticJavaParser;
 import com.github.javaparser.ast.CompilationUnit;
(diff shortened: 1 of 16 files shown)


src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -236,7 +237,10 @@ private static void performMinimizationImpl(
     Map<String, String> nonPrimaryClassesToPrimaryClass = new HashMap<>();
     SourceRoot sourceRoot = new SourceRoot(Path.of(root));
     sourceRoot.tryToParse();
-    for (CompilationUnit compilationUnit : sourceRoot.getCompilationUnits()) {
+    //getCompilationUnits does not seem to include all files, causing some to be deleted
+    for (ParseResult<CompilationUnit> res : sourceRoot.getCache()) {
+      CompilationUnit compilationUnit =
+          res.getResult().orElseThrow(() -> new RuntimeException("" + res.getProblems()));
```
kelloggm, line 243: Why `"" + ...` rather than just a call to `toString()`?

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
```diff
@@ -1588,6 +1588,9 @@ public boolean updatedAddedTargetFilesForPotentialEnum(FieldAccessExpr expr) {
       return false;
     }
     if (resolved.isEnumConstant()) {
+      if (JavaLangUtils.inJdkPackage(resolved.getType().describe())) {
+        return true;
```
kelloggm, line 1592: I'm surprised this is `return true` rather than `return false`, since in this case no update actually occurs.

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
gmerr3: done

src/main/java/org/checkerframework/specimin/UnsolvedSymbolVisitor.java
gmerr3: updated, does not seem to change functionality

kelloggm: probably means we should have another test case, but this seems fine to me

kelloggm (approved):

gmerr3:
not sure why it gives 89.01 here(and on main) but at https://github.com/gmerr3/specimin/actions/runs/11711689956/job/32620745538 it gives 89.49

PR #374, opened 2024-11-13
Ci logging improvement

M3CHR0M4NC3R:
As requested, here are changed to CI so the individual cases can be viewed. I'd like a second set of eyes to verify that the file paths are correct in this as I don't have an environment to run the CI script outright like github does.


kelloggm:
> I'd like a second set of eyes to verify that the file paths are correct in this as I don't have an environment to run the CI script outright like github does.

It looks like this is not working - the CI job is failing with:
```
Run if [ -f "$(pwd)/CI_constant_saved_run_percentage" ]; then
File CI_constant_saved_run_percentage does not exist
```

I suspect that the capitalization + file extension is the problem (the file in this PR is called `CI_Constant_saved_run_percentage.txt`) - note the capital "C" in "Constant" and the `.txt`, which differ from what the script is looking for. I suggest renaming the file so that its name is all lowercase.

kelloggm:
Otherwise, the output of this PR looks great - I love that CI now gives a list of failures and failed compilations.

kelloggm:
It looks like something is still wrong: the two accuracy numbers are pulled correctly (and show up in the script's logs in the "parse accuracy %" and "read latest run % from file" steps), but in the comparison one of them is missing.

kelloggm:
@M3CHR0M4NC3R it looks like CI is still failing only because the number actually does need to be updated. Please do that, and then we can merge this.

kelloggm:
@M3CHR0M4NC3R this PR should've been merged already. Can you resolve the conflict and then merge it?

kelloggm (approved):

PR #381, opened 2025-01-04
changed repository link for ASHE used in CI runs

M3CHR0M4NC3R:
changed repository link for ASHE used in CI runs to the one owned by NJIT-jerse. Just ran this a few times to test, should be good.

PR #383, opened 2025-01-08
Deinterlace multithreaded logs

M3CHR0M4NC3R:
When I multithreaded ASHE to speed up CI checks, I made the logs print out in a random order. This should filter it by thread and then sort the results alphabetically so it remains consistent regardless of how many threads the CI tests run on.
This should also let me figure out what failed during CI for #381.
If this PR passes CI don't merge it yet, I need to diff the results from this with my own results before that happens.


kelloggm (approved):
LGTM

M3CHR0M4NC3R:
now correctly displays which tests are failing after CI is done.

kelloggm:
@M3CHR0M4NC3R is this ready to merge? If so, please go ahead and do so.

PR #390, opened 2025-01-23
Handle `Map.Entry` properly in import removal, and remove duplicated logic

kelloggm:
Prior to this PR, there were two pieces of code in Specimin that were both trying to handle removing extraneous imports:
* some code in `PrunerVisitor`, and
* the `UnusedImportRemoverVisitor` class

I deleted the logic in `PrunerVisitor`, which was more complex but had the same overall impact. I enhanced the logic in `UnusedImportRemoverVisitor` to fix the bug evinced by the new test case, wherein the `import java.util.Map` was incorrectly removed. I also had to add some `continue` statements elsewhere in `PrunerVisitor`; that part makes me nervous, because a seemingly-unrelated test failed without them but did not fail until I removed the logic in `PrunerVisitor`. However, I much prefer this design and don't have time to dig deep into that, so I'm okay with ignoring that weirdness for now: the new `continue` statements are pretty clearly correct (it's not sensible to keep going through the loop body's logic if we've already decided to remove something).

~~I expect CI to fail, because this PR should resolve some compilation issues in plume-util. I'll update the "passed" baseline number once I know what the new result is.~~


M3CHR0M4NC3R (approved):
I am able to reproduce these results on my local machine as well. The accuracy was also reported at 91.24. The accuracy file will conflict with the other pull requrests so that will need to be checked again afterwards. Otherwise this looks good to me.

kelloggm:
> I am able to reproduce these results on my local machine as well. The accuracy was also reported at 91.24

Thanks for confirming. This is good to go, in that case.

PR #393, opened 2025-02-04
Avoid failed compilations caused by JDK classes without zero argument constructors

kelloggm:
E.g., in https://github.com/plume-lib/plume-util/blob/master/src/main/java/org/plumelib/util/EntryReader.java, if all the constructors are removed, we get compilation errors of the following form:
```
public class EntryReader extends LineNumberReader implements Iterator<String> {
       ^
    constructor LineNumberReader.LineNumberReader(Reader) is not applicable
      (actual and formal argument lists differ in length)
    constructor LineNumberReader.LineNumberReader(Reader,int) is not applicable
      (actual and formal argument lists differ in length)
1 error
```

I don't actually expect this PR to move the needle on plume-util compilation numbers, because fixing this problem exposed another issue in compiling the `EntryReader` class mentioned above (Specimin is incorrectly assuming that private classes can be referenced outside their containing class). That's a separate issue that I'll address in a new PR.


kelloggm:
I've just realized that I made an error here. This PR is not yet ready to go - the script I was using to test it was checking the wrong target method (which is why I was seeing that other error). Still, the issue is probably fixable, but I'll come back to it later.

kelloggm:
@M3CHR0M4NC3R this is now ready for review

kelloggm:
CI is failing, so this is not ready for review.

PR #394, opened 2025-02-06
Don't use private, protected, or package-private argument types as parameter types

kelloggm:
This resolves at least a few issues in plume-util, so I expect that I'll need to update the compilation numbers once CI runs. The fixed issues are of the form:
```
org/plumelib/util/WeakHasherMap.java:5: error: Subsequence has private access in Intern
    public <SyntheticUnconstrainedType> SyntheticUnconstrainedType get(org.plumelib.util.Intern.Subsequence<?> parameter0) {
                                                                                               ^
```

This PR changes types like `Subsequence` into `java.lang.Object`, which is always safe to use as a parameter type if a reference type was being used in the first place (the only place we can't use `Object` as a param type is when the type is a primitive).


M3CHR0M4NC3R (approved):
Looks good to me, was able to reproduce this locally with an improved result.

PR #395, opened 2025-02-06
Fix crash when a type parameter in a field declaration has an array type

kelloggm:
This caused some crashes in PlumeUtil's `Intern.java`, so I'll need to update the % of passing methods.


kelloggm:
While this PR is _probably_ okay to merge, it only fixes one crash of the five that I expected it to fix. I'm investigating the other four now, which have different symptoms even though they look superficially similar.

kelloggm:
Merging without review because I lack qualified reviewers for this right now.

Issue #396, opened 2025-02-06
CI should store the lists of failed minimizations and compilations rather than the percentage


kelloggm:
Currently, CI only stores the specific % of compilations/minimizations that succeed in plume-util. This makes that CI step very brittle, because even good changes need to make updates to it frequently. Moreover, it's hard to merge multiple PRs that are all good quickly, since each independently updates the CI %.

Instead, CI should store the list of expected failures, so that a PR writer can see exactly what new failures (or fixes) their PR causes. So, instead of storing e.g., the two-digit floating point value as we currently do in `CI_constant_saved_run_percentage.txt`, the text file should save a _list_ of failed mimizations/compilations, like the ones produced in CI already under the "show only failed minimizations" / "show only failed compilations" steps.

@M3CHR0M4NC3R how hard would this be to implement? If it's easy, we should do it ASAP.

kelloggm:
Was fixed by #406 ages ago, should have been closed then.

PR #406, opened 2025-03-15
instead of a percentage of successes, CI reads a diff

M3CHR0M4NC3R:
this should help with #396. CI will do a diff over the failures in minimization and compilation, and output results of that. This solution still requires that someone edit files that CI references when a PR is merged, but is less brittle. Also, this will give ASHE the ability to fail earlier if a test fails that isn't in the saved list.


M3CHR0M4NC3R:
I cannot run CI locally at this time, so this is remaining a draft until I can verify that my changes work.

.github/workflows/check_compilable_percentage.yml
```diff
@@ -108,37 +109,36 @@ jobs:
       - name: Show only failed minimizations
         run: |
           grep 'failed_minimization' ASHE/logs/specimin_run_results.txt | sort
+          grep 'failed_minimization' ASHE/logs/specimin_run_results.txt | sort > run_failed_minimizations.txt
       - name: Show only failed compilations
         run: |
           grep 'failed_compilation' ASHE/logs/specimin_run_results.txt | sort
-      - name: Parse accuracy percentage
-        id: parse_accuracy_percentage
+          grep 'failed_compilation' ASHE/logs/specimin_run_results.txt | sort > run_failed_compilations.txt
```
kelloggm, line 116: Use `tee` (https://en.wikipedia.org/wiki/Tee_(command)) instead of running this command twice.

kelloggm (approved):
Overall this is clearly an improvement. It looks like CI is failing, which is a bit of a surprise (but a good test that the infra works).

I restarted the CI job, because I have a suspicion that it is flaky. That's a separate issue that we need to address soon, though.

PR #409, opened 2025-03-25
outputs a specific error message when a file has no suitable methods

M3CHR0M4NC3R:
For CI: Should output a message for a problem CI is showing intermittently so I can track it down.


kelloggm (approved):
LGTM. Why is this PR a draft @M3CHR0M4NC3R?

kelloggm (approved):

PR #407, opened 2025-03-20
Fix a bug with the handling of type parameter bounds

kelloggm:
Discovered while investigating #402, though not sufficient to fix that one on its own.

Issue #403, opened 2025-03-12
string index out of bounds crash


M3CHR0M4NC3R:
I have uncovered the following crash while running ashe/specimin on a large set of inputs. The branch OOB-report contains a test "stringindexoob" that replicates this crash.
```
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index 5 out of bounds for length 5
at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
at java.base/java.util.Objects.checkIndex(Objects.java:365)
at java.base/java.util.ArrayList.get(ArrayList.java:428)
at java.base/java.util.Collections$UnmodifiableList.get(Collections.java:1502)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.updateClassSetWithStaticMethodCall(UnsolvedSymbolVisitor.java:3440)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:1165)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:106)
at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:135)
at com.github.javaparser.ast.visitor.ModifierVisitor.lambda$visit$100(ModifierVisitor.java:1020)
at java.base/java.util.Optional.map(Optional.java:260)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1020)
at org.checkerframework.specimin.SpeciminStateVisitor.visit(SpeciminStateVisitor.java:177)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:924)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:106)
at com.github.javaparser.ast.body.VariableDeclarator.accept(VariableDeclarator.java:153)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1111)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:52)
[...]

PR #410, opened 2025-04-01
Address various issues related to the crash in #402

### LINKED ISSUE #402: failed to solve return type crash
I uncovered a crash while running specimin/ashe on a large number of projects. The branch returnTypeFail contains the test of the same name that replicates this crash.
```
Exception in thread "main" java.lang.RuntimeException: failed to solve the return type (JsonInclude.Value) of getDefaultPropertyInclusion(Class<?>)
at org.checkerframework.specimin.TargetMemberFinderVisitor.visit(TargetMemberFinderVisitor.java:383)
at org.checkerframework.specimin.TargetMemberFinderVisitor.visit(TargetMemberFinderVisitor.java:53)
at com.github.javaparser.ast.body.MethodDeclaration.accept(MethodDeclaration.java:195)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1111)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:52)
at com.github.javaparser.ast.NodeList.accept(NodeList.java:275)
at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1149)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:239)
at org.checkerframework.specimin.SpeciminStateVisitor.visit(SpeciminStateVisitor.java:216)
at org.checkerframework.specimin.TargetMemberFinderVisitor.visit(TargetMemberFinderVisitor.java:231)
at org.checkerframework.specimin.TargetMemberFinderVisitor.visit(TargetMemberFinderVisitor.java:53)
at com.github.javaparser.ast.body.ClassOrInterfaceDeclaration.accept(ClassOrInterfaceDeclaration.java:152)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1111)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:52)
at com.github.javaparser.ast.NodeList.accept(NodeList.java:275)
at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1149)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:307)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:52)
at com.github.javaparser.ast.CompilationUnit.accept(CompilationUnit.java:134)
at org.checkerframework.specimin.SpeciminRunner.performMinimizationImpl(SpeciminRunner.java:418)
at org.checkerframework.specimin.SpeciminRunner.performMinimization(SpeciminRunner.java:164)
at org.checkerframework.specimin.SpeciminRunner.main(SpeciminRunner.java:91)
Caused by: UnsolvedSymbolException{context='null', name='JsonInclude.Value', cause='null'}
at com.github.javaparser.ast.type.ClassOrInterfaceType.convertToUsage(ClassOrInterfaceType.java:358)
at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:648)
at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:659)
at com.github.javaparser.symbolsolver.JavaSymbolSolver.toResolvedType(JavaSymbolSolver.java:427)
at com.github.javaparser.ast.type.ClassOrInterfaceType.resolve(ClassOrInterfaceType.java:336)
at org.checkerframework.specimin.TargetMemberFinderVisitor.visit(TargetMemberFinderVisitor.java:373)
... 21 more
```

kelloggm:
Fixes #402.

I think the test for #403 was incorrectly added to the branch for #402. I've added an `@Ignore` flag to it, which we can remove when we fix that properly. I fixed some issues related to it, but the test still doesn't pass (although it no longer crashes); this PR is too big already, though, so I'll leave that for later.

This PR incorporates the changes in #407 and should not be reviewed properly until after #407 is merged.

Summary of changes:
* make `SpeciminRunner` keep the list of previous iterations in a linked hash set, to help with debugging (I want them printed in the order they occur). This should have minimal overhead, because the number of elements in the set is small (usually < 10, even for large inputs; it is bounded by the program's longest def-use chain, like CF WPI).
* make the JavaTypeCorrect logic for updating types based on feedback from the compiler also search synthetic inner classes for methods and fields
* make inner class logic non-idempotent when a more-precise version of the inner class is discovered (necessary if it e.g. is found to have a method in a later iteration, as is the case in the `returnTypeFail` test case for `JsonInclude.Value`)
* fix the class and package map to avoid having the last package part in a wildcard import be in the map as if it were a type. This was just a bug, and is the proximate cause of the String OOB in #403 (iff there is a variable with the same name as the last part of the wildcard import's package name)
* make the `updateMissingClass` method return the class it actually updated, so that further uses of the classes don't use an malformed class that should be an outer/inner pair. This change had to be threaded through a lot of `UnsolvedSymbolVisitor`, but probably fixes a few latent miscompilation bugs
* unified a bunch of logic for splitting apart FQNs into package and class names
* fixed a bug in the logic for processing type names from javac that broke on unconstrained wildcards (it handled `extends` and `super` bounds, but not `?` by itself); this one is related to #403's test, not to this PR's test, but oh well
* some other misc updates to support inner classes, using the heuristic that class names are capitalized and package names are not


kelloggm:
Closing in favor of #468, which contains the same test case. This is fixed after the refactor in #421.

Issue #405, opened 2025-03-15
crashed while trying to get parent for WRAPPER_ARRAY


M3CHR0M4NC3R:
I have uncovered a bug while running specimin/ashe on a large input. The branch "WRAPPER_ARRAY" contains a new testcase to replicate this issue. Here is the stack trace.
```
Specimin output:

> Configure project :
No explicit dependency on the Checker Framework found, using default version 3.49.0

> Task :createCheckerFrameworkManifest UP-TO-DATE
> Task :spotlessInternalRegisterDependencies UP-TO-DATE
> Task :spotlessJava
> Task :spotlessJavaApply
> Task :spotlessMisc UP-TO-DATE
> Task :spotlessMiscApply UP-TO-DATE
> Task :spotlessApply
> Task :compileJava UP-TO-DATE
> Task :processResources UP-TO-DATE
> Task :classes UP-TO-DATE

> Task :run FAILED
Exception in thread "main" java.lang.RuntimeException: crashed while trying to get the parent for WRAPPER_ARRAY
at org.checkerframework.specimin.UnsolvedSymbolVisitor.updateSyntheticClassForSuperCall(UnsolvedSymbolVisitor.java:2384)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:979)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:106)
at com.github.javaparser.ast.expr.NameExpr.accept(NameExpr.java:80)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:1111)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:52)
at com.github.javaparser.ast.NodeList.accept(NodeList.java:275)
at com.github.javaparser.ast.visitor.ModifierVisitor.modifyList(ModifierVisitor.java:1149)
at com.github.javaparser.ast.visitor.ModifierVisitor.visit(ModifierVisitor.java:878)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:651)
at org.checkerframework.specimin.UnsolvedSymbolVisitor.visit(UnsolvedSymbolVisitor.java:106)
at com.github.javaparser.ast.stmt.SwitchEntry.accept(SwitchEntry.java:177)
[...]

PR #411, opened 2025-04-03
Correctly handle switches that use an unresolved enum as the selector

kelloggm:
This fixes the proximate cause of #405. However, there is some other, seemingly-unrelated crash that's then exposed. So, I wrote some test cases for this problem, specifically, and am PRing this fix now as a logical unit.

Here's an explanation, which is copied from one of the new test cases:
```
// The Java scoping rules for switches over enum constants
// are surprisingly complicated and poorly documented by the JLS.
// Specifically, Java permits an unqualified enum constant that
// is a member of the type used as the selector expression in
// a switch statement to be used either as a rule or as a case,
// without an explicit import. Specimin needs to correctly
// identify those cases as enum constants, because otherwise
// there usually is not _anything_ in scope that matches, which
// causes us to produce non-sensical output. However, I don't
// _think_ that the unqualified enum constant names can be used
// in the bodies of the expressions in the rest of the switch,
// and have implemented Specimin's logic assuming that is so.
```


kelloggm:
Superseded by #463, which is a port to the new architecture.

PR #419, opened 2025-07-02
Removed redundant import

adityajidhonde:
The previous logic in 'UnusedImportRemoverVisitor#removeUnusedImports' failed to remove redundant imports for nested classes within the same package.
Eg. if 'com.example.foo.MainClass' imports 'com.example.foo.OuterClass.Inner', the old condition '!relativeName.contains(".")' (where 'relativeName' is "OuterClass.Inner") erroneously kept the import.
Now if we add class from same package it does not add it as external package(import com.example.bar.Inner).


kelloggm:
@adityajidhonde there are only two lines of changes in this PR, and they aren't obviously related to the description. Does this PR contain the intended set of changes?

kelloggm:
PS don't worry about the failing CI job - it is known to be brittle. I'd like to fix that at some point, but for now we're okay merging PRs if the other two CI jobs succeed.

PR #421, opened 2025-07-15
Refactor

theron-wang:
Includes the refactored algorithm for fully solvable projects. At the moment, I removed the modularity model code, but I'll add that back in before the end. I also don't have file output code yet, since I'm waiting to finish unsolved symbol generation (with the new alternates option).

Most of the code in `TargetMemberFinderVisitor` is unchanged from before; I just removed some methods that I don't think are necessary in the new implementation. Most of the new logic is located in `Slicer` and `TypeRuleDependencyMap`.


src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -222,14 +210,31 @@ public static String getValueTypeFromAnnotationExpression(Expression value) {
...
+
+  /**
+   * See {@code getEnclosingClassLike(Node)} for more details. This does not throw if a parent is
+   * not found; instead, it returns null.
+   *
+   * @param node a node that is contained in a class-like structure
+   * @return the nearest enclosing class-like node
+   */
+  public static @Nullable TypeDeclaration<?> getEnclosingClassLikeOptional(Node node) {
+    Optional<Node> parent = node.getParentNode();
```
kelloggm, line 229: Rather than two copies of code that logically does the same thing, could you implement this version in terms of the other (e.g., by catching an exception) or vice-versa? I think it should be relatively easy to do so.

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,137 @@
...
+import org.checkerframework.specimin.unsolved.UnsolvedSymbolAlternates;
+import org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator;
+
+public class Slicer {
+  /** The slice of nodes. */
+  private final Set<Node> slice = new HashSet<>();
+
+  /**
+   * The slice of unsolved symbol definitions. These values need not be unique; the map is provided
+   * for simple lookups when adding new symbols.
```
kelloggm, line 29: I'm concerned that these values don't need to be unique - maps don't really make sense unless the keys aren't going to overlap. Moreover, I'd generally expect that unsolved symbol names are unique (at least, within a single compilation unit): if not, the original Java code wouldn't compile, either - at least, assuming that the keys here are fully-qualified names. Are they?

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,137 @@
...
+      // It's guaranteed that these are relevant to the slice due to the algorithm.
+
+      // We need to add this to the slice (or we potentially face missing
+      // declarations or broken code). We also need to ask the type rule
+      // dependency map if we can break this node down further.
+
+      slice.add(node);
+    }
+
+    worklist.addAll(TypeRuleDependencyMap.getRelevantElements(node));
```
kelloggm, line 109: The type rule dependency map will eventually need to be an object that is passed in, I think.

src/main/java/org/checkerframework/specimin/Slicer.java
kelloggm: > However, using sets as the keys would prompt a new SomeClass, with only FQN org.example.SomeClass, to be generated

I don't think this is necessarily a problem - we could replace the key with an updated one when we discover this constraint, no?

> This is an edge edge case, so we may never encounter this. Do you think it's worth accounting for?

I think it's a good idea to try to get this right the first time, since we're rewriting the core of the analysis. So, let's try to make sure we've thought of and accounted for all of the edge cases.

kelloggm: > My understanding is that you're suggesting a map of type Map<Set<String>, UnsolvedSymbol>, right?

I was imagining a custom data structure for the key that's backed by a string set but has its own looser equality method. Regardless, I think we're on the same page with respect to what this needs to handle; the exact way that you implement it is less important. You should do what you find most intuitive.

theron-wang:
@kelloggm I just pushed some commits for unsolved symbol generation. At this point, I haven't added support for type parameters and some more complex inference logic yet.

I'm aiming to make `UnsolvedSymbolGenerator.addInformation` into a method that can be called multiple times on the same generated symbol, each in different contexts. For example, `inferContextImpl` will generate the symbols, but `addInformation` will see if anything needs to be changed from the original definition. If a type is found in a `try` block, then it implements `AutoCloseable`; if a type is found in an `implements` clause, then it's changed to an interface; etc.

I haven't made the changes you asked for yet, but I'll make sure to address them before I give you the final version.

theron-wang:
Approximately 60 cases pass now. Some of these cases fail because they have slightly different definitions, though.

For example, take `UnsolvedStaticSimpleField`:

Expected:
```java
package org.sampling;

public class Baz {

    public static OrgSamplingBazMyFieldSyntheticType myField;
}
```

Generated:
```java
package org.sampling;
public class Baz {
    public static org.sampling.OrgSamplingBazMyFieldSyntheticType myField;
}
```

The only difference is the FQN, which I don't see any harm in keeping in the output. Do you think we should just change the test cases to reflect this or should we trim the FQN? Trimming the FQN would require a change in the way unsolved symbols are currently outputted, though, since we'd have to pass information from the unsolved type into each unsolved member, in addition to what we currently do: pass information from the unsolved members into the unsolved type (if you look at the `toString` methods this will make more sense).

kelloggm:
> The only difference is the FQN, which I don't see any harm in keeping in the output. Do you think we should just change the test cases to reflect this or should we trim the FQN?

Generally, I think output with the FQN is better. So, in this case, I'd argue that the new output is actually an improvement on the old output, and you should definitely change the test case.

Going the opposite direction (FQN -> no FQN) would be a regression, though.

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -373,4 +360,101 @@ static String removeMethodReturnTypeAndAnnotationsImpl(String declAsString) {
     String result = methodWithoutReturnType.replace("< ", "<");
     return result;
   }
+
+  /**
+   * Checks to see if an expression (usually a field or method) is static.
```
kelloggm, line 365: > usually a field or method

Since these are expressions, this must be a _reference_ to a field or method, rather than a field or method definition.

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -373,4 +360,101 @@ static String removeMethodReturnTypeAndAnnotationsImpl(String declAsString) {
...
+   */
+  public static boolean isAStaticMember(Expression expr) {
+    CompilationUnit cu = expr.findCompilationUnit().get();
+
+    String nameOfExpr;
+    String nameOfScope = null;
+    Expression scope = null;
+
+    if (expr.isNameExpr()) {
+      nameOfScope = nameOfExpr = expr.asNameExpr().getNameAsString();
```
kelloggm, line 378: This double-assignment pattern is a bit unintuitive to read and is usually considered bad style in Java. It's more verbose to use three lines for this, but I'd prefer it.

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -373,4 +360,101 @@ static String removeMethodReturnTypeAndAnnotationsImpl(String declAsString) {
...
+    if (nameOfScope == null) {
+      return false;
+    }
+
+    for (ImportDeclaration importDecl : cu.getImports()) {
+      if (importDecl.isStatic() && importDecl.getNameAsString().endsWith("." + nameOfExpr)) {
+        return true;
+      }
+
+      if (!importDecl.isStatic() && importDecl.getNameAsString().endsWith("." + nameOfScope)) {
```
kelloggm, line 415: This condition is exactly identical to the condition above, but with `!` before the call to `isStatic()`. That's unexpected, given the method name, and seems suspicious to me. Certainly it justifies an explanatory comment, if it's actually reasonable to do.

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -373,4 +360,101 @@ static String removeMethodReturnTypeAndAnnotationsImpl(String declAsString) {
     String result = methodWithoutReturnType.replace("< ", "<");
     return result;
   }
+
+  /**
+   * Checks to see if an expression (usually a field or method) is static.
+   *
+   * @param expr The expression
+   * @return Whether it is static or not
```
kelloggm, line 368: Is this method really algorithmic, in the sense that it always gets the right answer? It seems to me that it is based on heuristics, and if that's a fair description it should be in the Javadoc.

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,212 @@
...
+  private final Set<Node> postProcessingWorklist = new HashSet<>();
+
+  private final Map<String, Path> existingClassesToFilePath;
+  private final String rootDirectory;
+  private final Map<String, CompilationUnit> toSlice;
+
+  private final UnsolvedSymbolGenerator unsolvedSymbolGenerator;
+  private final TypeRuleDependencyMap typeRuleDependencyMap;
+
+  public Slicer(
```
kelloggm, line 56: This constructor needs Javadoc explaining its arguments and, more importantly, where they come from.

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,212 @@
...
+      TypeRuleDependencyMap typeRuleDependencyMap,
+      UnsolvedSymbolGenerator unsolvedSymbolGenerator) {
+    this.existingClassesToFilePath = existingClassesToFilePath;
+    this.rootDirectory = rootDirectory;
+    this.toSlice = toSlice;
+    this.typeRuleDependencyMap = typeRuleDependencyMap;
+    this.unsolvedSymbolGenerator = unsolvedSymbolGenerator;
+  }
+
+  public void slice() throws IOException {
```
kelloggm, line 69: Is this method _always_ supposed to be called on a member of this class (I suspect so). If that's the case, make the contract simpler: this method can be private, and the constructor can call it.

Alternatively, make the constructor private and provide a static method that both constructs the object and runs the slice() method, and then returns the result to the caller.

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,212 @@
...
+import java.util.IdentityHashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.checkerframework.specimin.unsolved.AddInformationResult;
+import org.checkerframework.specimin.unsolved.UnsolvedSymbolAlternates;
+import org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator;
+
+public class Slicer {
```
kelloggm, line 28: I would expect a slicer to take a program and a dependency map as input and output the slice. It's not clear that this class actually has that contract, though: the constructor has some other arguments, and the slice() method is void. How does a user of this class get the output? The Javadoc should explain how the class is intended to be used.

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,212 @@
...
+    Set<CompilationUnit> usedCompilationUnits = new HashSet<>();
+
+    // Step 1: build the slice; see which nodes to keep
+    while (!worklist.isEmpty()) {
+      Node element = worklist.remove();
+      handleElement(element, usedCompilationUnits);
+    }
+
+    if (!generatedSymbolSlice.isEmpty()) {
+      // Step 2: Handle the post-processing worklist
```
kelloggm, line 79: What is the post-processing worklist? It's not mentioned anywhere; maybe `slice()` needs Javadoc explaining how it differs from the algorithm in the paper?

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,212 @@
...
+      toSlice.put(
+          qualifiedNameToFilePath(used.getPrimaryType().get().getFullyQualifiedName().get())
+              .toString(),
+          used);
+    }
+
+    // Step 4: remove all nodes except the ones marked for preservation OR package/import
+    // declarations
+    for (CompilationUnit cu : toSlice.values()) {
+      sliceCompilationUnit(cu);
```
kelloggm, line 100: Confusingly-named method, given what the comment above says that it does.

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+
+  public Set<UnsolvedSymbolAlternates<?>> getGeneratedSymbolSlice() {
+    return generatedSymbolSlice;
+  }
+
+  public void addToWorklist(Node node) {
+    worklist.add(node);
+  }
```
kelloggm, line 110: Why do these methods (this one and `getGeneratedSymbolSlice()`) need to be public? They seem like class internals.

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+
+  /**
+   * Adds all nodes to the beginning of the worklist. This ensures that synthetic types are
+   * generated before they are used. Use instead of worklist.addAll.
+   *
+   * @param c The list of nodes.
+   */
+  private void addToBeginningOfWorklist(List<Node> c) {
```
kelloggm, line 154: Rather than this method, we probably should use a different data structure - maybe a FILO queue? 

In general, the need to write a method like this one is almost always an indication ("code smell") that you should do something differently at the architecture or design level - in this case, it indicates that the data structure we chose doesn't quite match the needs of the algorithm.

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -0,0 +1,212 @@
...
+      sliceNode(typeDecl);
+    }
+  }
+
+  /**
+   * Recursively removes all nodes not in the slice.
+   *
+   * @param node The node to slice
+   */
+  private void sliceNode(Node node) {
```
kelloggm, line 179: I'd prefer that we reserve the term "slice" for the overall process, and use a different term for the internal step of removing nodes that this method implements.

kelloggm (changes requested):
I haven't reviewed the classes in the `unsolved` package yet; I'll try to get to them later.

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
theron-wang: There's a slight difference in the two conditions (`nameOfScope` vs `nameOfExpr` at the end of each line), but I did add some comments to make this a bit more clear.

src/main/java/org/checkerframework/specimin/Slicer.java
theron-wang: Yeah, this didn't make me too happy when I wrote it. I changed it around so this method isn't necessary: the target member information goes FILO (worklist does removeLast instead of removing from the beginning), but the type rule dependency map adds to the end, so that part would be FIFO. The reasoning behind this is because certain statements, like A a = new A(), need to be handled before a.method() (so type A is generated). The old method was to circumvent the lack of an addAll to the beginning of the Deque, but doing removeLast resolves the need for such a method. Is this still a problem or do you think it's fine?

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -202,217 +216,61 @@ private static void performMinimizationImpl(
...
     // getCompilationUnits does not seem to include all files, causing some to be deleted
     for (ParseResult<CompilationUnit> res : sourceRoot.getCache()) {
       CompilationUnit compilationUnit =
           res.getResult().orElseThrow(() -> new RuntimeException(res.getProblems().toString()));
       Path pathOfCurrentJavaFile =
           compilationUnit.getStorage().get().getPath().toAbsolutePath().normalize();
-      String primaryTypeQualifiedName = "";
-      if (compilationUnit.getPrimaryType().isPresent()) {
-        // the get() is safe because primary type here is definitely not a local declaration,
-        // which does not have a fully-qualified name.
```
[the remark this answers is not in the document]
theron-wang: I don't think this handling is removed, at all; it's less explicitly handled now, of course, but I still think the functionality is there. First, the addition a few lines below (`fqnToCompilationUnits.put(declaredClassQualifiedName, compilationUnit);`) makes sure that both non-primary and primary classes point to the same compilation unit. In the type rule dependency map, when a reference to a type is found, it uses the `fqnToCompilationUnits` map to find the right CU and also adds the non-primary class to the slice. The newer code doesn't handle these cases separately.

src/main/java/org/checkerframework/specimin/TargetMemberFinderVisitor.java
```diff
@@ -1,65 +1,43 @@
...
  */
-public class TargetMemberFinderVisitor extends SpeciminStateVisitor {
+public class TargetMemberFinderVisitor extends ModifierVisitor<Void> {
   /**
    * The names of the target methods. The format is
    * class.fully.qualified.Name#methodName(Param1Type, Param2Type, ...). All the names will have
    * spaces remove for ease of comparison.
    */
-  private final Set<String> targetMethodNames;
+  private final Set<String> targetMethods;
```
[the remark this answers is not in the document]
theron-wang: Since `SpeciminStateVisitor` no longer exists, and `targetMethods` is no longer unavailable as a field name, I just thought it sounded better to rename it to `targetMethods`. There is no difference between the old field and the current field. I did change it back to `targetMethodNames`, though.

theron-wang:
@kelloggm How much do you think we need to stick to the old naming conventions with generated field/method types? I'm encountering a few failing tests where the refactor generates shorter names than the old version. For example, `UnsolvedMethodWithArrayParameter` fails because the return type for `UnsolvedType.processArray(intArray);` is differently named. The old has it named `OrgTestingUnsolvedTypeProcessArrayReturnType`, while the new has it named `ProcessArrayReturnType`, both in the same package. If it's not too big of a deal, I won't fix this, but if you think it's important to stick to the original, then I can, too. It definitely is possible to replicate the original behavior in this instance, but I'm not sure if it's worth the effort.

kelloggm:
> How much do you think we need to stick to the old naming conventions with generated field/method types?

In general, it's not a strict contract: these generated method/field names are arbitrary. As long as we're confident that the name is unambiguous, it's fine.

> The old has it named OrgTestingUnsolvedTypeProcessArrayReturnType, while the new has it named ProcessArrayReturnType, both in the same package

In this case, I think there is something to worry about: what if there are two methods called `processArray` with different argument lists in the same class, both of which are used by the target? E.g., suppose that we have:
```
processArray(int[] foo) { ...}
```
and
```
processArray(int[] foo, boolean bar) { ... }
```
with calls to each. I don't think the previous code would get this correct, either, but since we're thinking about this we should try our best to make sure that the name will always be unique.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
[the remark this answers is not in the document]
kelloggm: Okay, I understand. This is a better design, and removes a complication from the algorithm. Nice job!

src/main/java/org/checkerframework/specimin/Slicer.java
kelloggm: > Is this still a problem or do you think it's fine?

That sounds fine. I'm not too concerned about efficiency here, just making sure the code is readable and we don't need to reinvent anything. I'll double-check it when I re-review this code.

kelloggm (commented):
Partial review of `unsolved`, mostly focused on `FullyQualifiedNameGenerator`. I'll keep going in a bit, need to turn my attention to a different task now.

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
```diff
@@ -0,0 +1,984 @@
...
+/**
+ * Helper class for {@link UnsolvedSymbolGenerator}. Generates all FQNs based on an expression or
+ * type.
+ */
+public class FullyQualifiedNameGenerator {
+  /**
+   * When evaluating an expression, there is only one possible type. However, the location of an
+   * expression could vary, depending on the parent classes/interfaces of the class which holds the
+   * expression. When determining the location of a {@link FieldAccessExpr}, {@link MethodCallExpr},
+   * or {@link NameExpr}, use this method instead of {@link #getFQNsForExpressionType} because it
```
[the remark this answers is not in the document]
theron-wang: No, because `#getFQNsForExpressionType` can still be used with those types. This returns the location of the expression, while `#getFQNsForExpressionType` returns its type. If you called the two functions on a MethodCallExpr, `foo()`, for example, the two methods would both be valid but would return different results. `#getFQNsForExpressionType` would return the return type (maybe `FooReturnType`) but `#getFQNsForExpressionLocation` would return the FQNs of potential locations (i.e., static import, ancestor class/interface).

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
```diff
@@ -0,0 +1,984 @@
...
+
+          if (toAst instanceof Parameter param && param.getType().isUnionType()) {
+            UnionType unionType = param.getType().asUnionType();
+
+            Map<String, Set<String>> result = new LinkedHashMap<>();
+
+            for (ReferenceType type : unionType.getElements()) {
+              try {
+                // If a type in the union type is resolvable, the location of the expression will
+                // be in a built-in Java superclass. In this case, return an empty map.
```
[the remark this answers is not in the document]
theron-wang: If a union type is `UnsolvedException | NullPointerException`, then any method called on the NameExpr representing an exception of this type will be in `Exception` or `Throwable`. This was my original thought process and why I left the comment. 

I do realize that in this case, there may be some ambiguity: if it were `UnsolvedException | MyException`, and `MyException` is resolvable, and `MyException` extends some unsolvable class, then it could be in this unsolvable parent class. I'll leave a TODO to address that later.

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
```diff
@@ -0,0 +1,984 @@
...
+   * Gets FQNs of an annotation.
+   *
+   * @param anno The annotation
+   * @return A set of possible FQNs
+   */
+  public static Set<String> getFQNsFromAnnotation(AnnotationExpr anno) {
+    // If an annotation is @Foo.Bar, we need to find the import with org.example.Foo, not
+    // org.example.Foo.Bar.
+    // Hence, look for the import with the "earliest" scope (with @Foo.Bar, this would be Foo).
+    String getImportedName = anno.getNameAsString();
```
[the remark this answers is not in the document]
theron-wang: I don't think so, I believe this is the minimum. The method above deals with `ClassOrInterfaceType` which returns another `ClassOrInterfaceType` with the scope, while this method deals with `Name` as the scope. The annotation parent handling is also a little different so that can't be moved into the same method either.

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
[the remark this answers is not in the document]
kelloggm: > When determining the location of a {@link FieldAccessExpr}, {@link MethodCallExpr}, or {@link NameExpr}, use this method instead of {@link #getFQNsForExpressionType}

That makes this statement confusing, because that statement implies that substituting the two methods is a sensible thing to do for other kinds of expressions. We should rephrase this to make it clear that these methods return different things (i.e., what you said in your last comment).

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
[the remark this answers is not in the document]
kelloggm: Please also add the reasoning in your first paragraph to the comment in the code.

kelloggm (commented):
More comments; I'm now most of the way through `unsolved`, although I didn't look at all of `UnsolvedSymbolGenerator` closely yet.

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedClassOrInterfaceAlternates.java
```diff
@@ -0,0 +1,272 @@
...
+      String fqn,
+      List<UnsolvedClassOrInterfaceAlternates> allGenerated,
+      Map<String, UnsolvedSymbolAlternates<?>> generatedSymbolsMap) {
+    String packageName = fqn.substring(0, fqn.lastIndexOf('.'));
+    String className = fqn.substring(packageName.length() + 1);
+
+    UnsolvedClassOrInterfaceAlternates generated;
+
+    if (generatedSymbolsMap.containsKey(packageName)) {
+      return (UnsolvedClassOrInterfaceAlternates) generatedSymbolsMap.get(packageName);
```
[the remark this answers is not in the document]
theron-wang: > When I read this class' constructor, I (incorrectly, it seems) assumed that the keys to this map were FQNs.

You're right about this assumption.

I think the variable `packageName` is misleading here. Perhaps `qualifier` would be a better name; for inputs like `org.example.MyClass.MyOtherClass`, `org.example.MyClass` would be in `packageName`, but it's clearly not a package name. I also wrote this method before we clarified that we're assuming class paths represent classes, not packages, so the previous implementation treated `org.example.MyClass.MyOtherClass`, `org.example.MyClass`, and `org.example` all as classes. ll update that.

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedMethod.java
```diff
@@ -0,0 +1,217 @@
...
+    if (!(o instanceof UnsolvedMethod)) {
+      return false;
+    }
+    UnsolvedMethod other = (UnsolvedMethod) o;
+    // This set of fields is based on the JLS' overloading rules. According to the documentation of
+    // Oracle: "You cannot declare more than one method with the same name and the same number and
+    // type of arguments, because the compiler cannot tell them apart. The compiler does not
+    // consider return type when differentiating methods, so you cannot declare two methods with the
+    // same signature even if they have a different return type."
+    return other.name.equals(this.name) && other.parameterList.equals(parameterList);
```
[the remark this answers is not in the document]
theron-wang: I don't think this is too big of a deal, since most places where equality is necessary deal with `UnsolvedMethodAlternates` instead of `UnsolvedMethod`. The only case, as of right now, where `UnsolvedMethod` equality is used is in the `UnsolvedSymbolEnumerator`. Do you think its acceptable to just leave a note in the Javadoc to use caution with this method for this case?

src/main/java/org/checkerframework/specimin/Slicer.java
```diff
@@ -182,7 +182,9 @@ private void handleElement(Node node) {
       try {
         // Resolve isn't perfect: methods/constructors, even if in the same file, will not resolve
         // if there are unresolvable argument types
-        resolved = asResolvable.resolve();
+        if (!(node instanceof FieldDeclaration)) {
+          resolved = asResolvable.resolve();
```
kelloggm, line 186: Why is this call to `resolve` skipped if `node` is a `FieldDeclaration`? Adding an explanation to the comment would be helpful.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
```diff
@@ -314,4 +339,149 @@ public List<Node> getRelevantElements(Object resolved) {
...
+        } else if (resolvedTypeDecl
+            instanceof ReflectionInterfaceDeclaration reflectionInterfaceDeclaration) {
+          isInterface = true;
+          methods = reflectionInterfaceDeclaration.getDeclaredMethods();
+        } else {
+          continue;
+        }
+
+        for (ResolvedMethodDeclaration resolvedMethodDecl : methods) {
+          // Don't preserve equals(Object) and hashCode()
```
kelloggm, line 411: Why are these two methods, specifically, excluded? I would expect that any justification for excluding `equals` and `hashCode` would also apply to other object methods, like `clone`, no? If these two methods are special for some reason, this comment should explain why.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
```diff
@@ -314,4 +339,149 @@ public List<Node> getRelevantElements(Object resolved) {
...
+            continue;
+          }
+
+          if (resolvedMethodDecl.getName().equals("hashCode")
+              && resolvedMethodDecl.getNumberOfParams() == 0) {
+            continue;
+          }
+
+          for (MethodDeclaration methodDecl : originalTypeDecl.findAll(MethodDeclaration.class)) {
+            if (areAstAndResolvedMethodLikelyEqual(resolvedMethodDecl, methodDecl)) {
```
kelloggm, line 424: Why can't we use FQNs for the parameters in `areAstAndResolvedMethodLikelyEqual`? That would make this correct beyond a doubt.

src/test/resources/constraintType/expected/com/example/Simple.java
```diff
@@ -1,5 +1,6 @@
 package com.example;
 
+import java.util.Set;
```
kelloggm, line 3: `Set` isn't used in this example. While there's not necessarily _harm_ in unnecessary imports, I'd prefer that Specimin not generate them.

Removing unnecessary imports is something that IDEs and formatters already do well, though. Maybe the simplest solution is to run an automated formatter (e.g., [google-java-format](https://github.com/google/google-java-format)) over Specimin's output automatically?

kelloggm (commented):
This review is only on the changes in commit c369d72

kelloggm (commented):
Review of commit 6e928e0

src/test/resources/methodreturnfullyqualifiedgeneric/expected/com/bar/Bar.java
```diff
@@ -2,7 +2,7 @@
 
 public class Bar<T> {
 
-    public static com.bar.Bar<com.example.InOtherPackage2> getOtherPackage2() {
+    public static com.bar.Bar<com.foo.InOtherPackage2> getOtherPackage2() {
```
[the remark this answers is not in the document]
kelloggm: https://github.com/njit-jerse/specimin/issues/422

src/test/resources/unsolvedmethodtypecorrect/expected/com/example/Simple.java
```diff
@@ -1,15 +1,13 @@
...
 class Simple {
     void bar() {
         Foo f = new Foo();
         f.getLocals().set(0);
         baz();
-    }
-
-    void baz() {
-        throw new java.lang.Error();
+        final LocalVariables locals = f.getLocals();
```
kelloggm, line 11: This change doesn't look correct to me. The original input file is:
```
package com.example;

import org.example.LocalVariables;
import org.example.Foo;

class Simple {
    // Target method.
    void bar() {
        Foo f = new Foo();
        f.getLocals().set(0);
        baz();
    }

    void baz() {
        // To trigger JavaTypeCorrect.
        Foo f = new Foo();
        final LocalVariables locals = f.getLocals();
    }
}
```

Putting the contents of the `baz()` method into the `bar()` method definitely seems incorrect.

kelloggm (commented):
Review of changes in commit 81648d5

theron-wang:
@kelloggm If you're not reviewing already, you can hold off on doing so for the next day or so; I'm going to try and get rid of all the miscellaneous bugs before I tackle some of the larger problems. I expect that once I'm done with these bug fixes, non-UnsolvedSymbolGenerator files will stay relatively unchanged, so you shouldn't have to review another time.

src/test/resources/unsolvedmethodtypecorrect/expected/com/example/Simple.java
kelloggm: > I modified the test case (the input as well) to move the line into bar() so that we'd still be able to test for the same thing, which is why the output here looks the way it does.

Thanks, I missed that you'd changed both.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
kelloggm: > equals and hashCode are in the Set interface

But they are already defined on the object because they are defined in `java.lang.Object`, so being in an interface is immaterial. I think this exception should apply to all of the methods in `java.lang.Object`, not just to equals and hashCode. It's okay to just add a TODO though - I don't think this is very important.

theron-wang:
@kelloggm Ready for review, except for UnsolvedSymbolGenerator. I bumped up the gradle version to try to get it to run faster, but to no avail (30s to 60s still for each change).

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
```diff
@@ -133,6 +144,54 @@ public List<Node> getRelevantElements(Node node) {
...
+                      .filter(s -> s.isExplicitConstructorInvocationStmt())
+                      .findFirst()
+                      .get();
+
+              elements.add(firstStatement);
+            }
+          }
+        } catch (UnsolvedSymbolException ex) {
+          // No preservation of super() call necessary because the synthetic type will
+          // have a default constructor.
```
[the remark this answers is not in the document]
kelloggm: (we should just add this test case, I think)

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
```diff
@@ -376,21 +378,61 @@ else if (expr.isCastExpr()) {
...
+          || operator == Operator.LESS_EQUALS
+          || operator == Operator.GREATER_EQUALS) {
+        return new FullyQualifiedNameSet("boolean");
+      } else {
+        // Treat all other cases; type on one side is equal to the other
+        FullyQualifiedNameSet leftType = getFQNsForExpressionTypeImpl(binary.getLeft(), canRecurse);
+        FullyQualifiedNameSet rightType =
+            getFQNsForExpressionTypeImpl(binary.getRight(), canRecurse);
+
+        // Remaining operators only work with primitive/String types
```
kelloggm, line 410: What about boxed types like `java.lang.Integer`? Those are also possible here, although maybe you can just treat them as their primitive equivalents.

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
```diff
@@ -376,21 +378,61 @@ else if (expr.isCastExpr()) {
...
+            && JavaLangUtils.isJavaLangOrPrimitiveName(leftType.erasedFqns().iterator().next())) {
+          return leftType;
+        }
+        if (rightType.erasedFqns().size() == 1
+            && JavaLangUtils.isJavaLangOrPrimitiveName(rightType.erasedFqns().iterator().next())) {
+          return rightType;
+        }
+
+        // TODO: alternates here? cases here are all primitives (except for +, which could be
+        // String)
```
kelloggm, line 423: It's easy to cover all of the alternates here (we know the list of primitives), so we should just do so to avoid a headache later.

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedFieldAlternates.java
```diff
@@ -50,12 +87,40 @@ public static UnsolvedFieldAlternates create(
...
+  /**
+   * Updates field types and must preserve nodes. Saves the intersection of the previous and the
+   * input, since we know more information to narrow potential field types down.
+   *
+   * @param typesToPreserveNodes A map of field types to nodes that must be preserved
+   */
+  public void updateFieldTypesAndMustPreserveNodes(
+      Map<MemberType, CallableDeclaration<?>> typesToPreserveNodes) {
+    // Update in-place; intersection = removing all elements in the original set
+    // that isn't found in the updated set
```
kelloggm, line 105: Can we have one implementation of this intersection logic that's shared by the different `XAlternates` classes?

kelloggm (changes requested):

kelloggm:
> @kelloggm Ready for review, except for UnsolvedSymbolGenerator. I also bumped up the gradle version to try to get it to run faster, but to no avail (30s to 60s still for each change).

I've reviewed everything except `UnsolvedSymbolGenerator.java`.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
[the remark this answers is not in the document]
theron-wang: The case that you pointed out does indeed work, but you did help me realize another problem:

```java
class Foo extends Bar {
    public Foo() {
       super("any argument", "some other argument");
    }

    public static Bar bar() {
        Foo foo = new Foo();
        return new Bar("a consistent argument");
    }
}
```

This slightly modified version (where `bar()` is the only target method) doesn't preserve the super(...) call and that super constructor is not preserved. Maybe we should make it so if the parent is unresolvable, then always preserve these super calls? Or should we always explicitly generate a default constructor?

I think my original code was under the assumption that the parent class would not have an object creation expression in the target method, so the generated parent would have no constructor (and thus have the default constructor).

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
kelloggm: > I can do this after I address the other failing cases, though.

It's not a priority, I agree. Leaving the TODO here is fine: we can come back to it or not later.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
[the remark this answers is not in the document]
kelloggm: > I think my original code was under the assumption that the parent class would not have an object creation expression in the target method, so the generated parent would have no constructor (and thus have the default constructor).

I don't think we can make this assumption, as your test case shows (nice adaptation, by the way!).

> Maybe we should make it so if the parent is unresolvable, then always preserve these super calls?

I prefer this approach. It may result in us over-preserving constructors, but I'd rather do that than generate default constructors that might not actually be present in the original version of the code. At least we know that these super constructors must exist, because we have legitimate calls to them in code from the user.

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -766,14 +862,51 @@ public static List<MethodDeclaration> tryResolveMethodCallWithUnresolvableArgume
     return candidates;
   }
 
+  /**
+   * Given an enum constant declaration, returns all possible constructors that match the arity and
+   * known types of the arguments.
```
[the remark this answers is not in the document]
theron-wang: I didn't change it to null, but I did clarify with this statement in the Javadoc: This returns an empty list if no matching constructors were found or if the declaring type could not be solved.

Is that alright? The `tryResolve` methods are different in purpose than the other `try` methods so I think it shouldn't be a big deal for consistency if it returns null or not (it would make it a bit more annoying for the method's user if it were null instead of empty list).

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
```diff
@@ -609,7 +632,37 @@ private void getAllMustImplementMethodsImpl(
...
-          continue;
+          methods = Set.of();
+        }
+
+        // Compose previousTypeParametersMap with the current typeParametersMap.
+        // For example, if previousTypeParametersMap is T --> String and current is E --> T,
+        // then the composed map should be E --> String.
+        List<Pair<ResolvedTypeParameterDeclaration, ResolvedType>> typeParametersMap;
+
+        // Now, substitute previous type parameters through the current map
```
kelloggm, line 643: Do we have a test case that requires this behavior? If not, we should write one to double check that this is correct. (I recall that this wasn't in the last commit, so if adding it caused a test to go from failing to passing, we're already ok. If not, though, you should write a corresponding test.)

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
```diff
@@ -988,6 +999,121 @@ private void handleConstructorCall(
...
+          getOrCreateMemberTypeFromFQNs(fullyQualifiedNameGenerator.getFQNsFromType(exception));
+      exceptions.add(exceptionType);
+    }
+
+    AccessSpecifier specifier = methodDecl.getAccessSpecifier();
+    String accessModifier =
+        switch (specifier) {
+          case PUBLIC -> "public";
+          case PROTECTED -> "protected";
+          case PRIVATE -> "private";
```
kelloggm, line 1101: The `PRIVATE` case isn't possible, right, since this method must be an override? I would throw here rather than proceeding if we do hit this case, because I think it's an error condition.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
kelloggm: Nevermind, I found MustImplementMethodsComplexTest

kelloggm (commented):
Review of 7a6fc1c

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -64,6 +76,35 @@ private JavaParserUtil() {
     throw new UnsupportedOperationException("This class cannot be instantiated.");
   }
 
+  /**
+   * Set this TypeSolver instance to be used by the JavaParserFacade, so we don't have to find it
+   * through reflection in JavaParserSymbolSolver. This field is initialized once in SpeciminRunner.
+   * Note that by the time you evaluate the value of this field, it should already be non-null.
```
kelloggm, line 82: The [MonotonicNonNull annotation](https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/MonotonicNonNull.html) exactly expresses this contract.

kelloggm: (you'll also want an `@EnsuresNonNull` annotation on `setTypeSolver`, I think)

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -1365,10 +1406,269 @@ public static String getInitializerRHS(String variableType) {
...
+
+      // Using what we already know, let's try to find the value of the implicit type argument
+
+      // First, try to find a Type node that corresponds with the return type of methodCallParent.
+      // This could be a variable declarator, a return type in a method declaration, or a parameter
+      // type from another declaration.
+
+      // Don't use a map here: in case a ResolvedType is the same for different parameters, we could
+      // have different pieces of information (i.e., two parameters are both resolved type T, but
+      // the args
```
kelloggm, line 1524: nit: awkward formatting

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -1365,10 +1406,269 @@ public static String getInitializerRHS(String variableType) {
...
+   * the corresponding method declaration from {@code Foo}. A similar case can be found in
+   * LambdaBodyStaticUnsolved2Test.
+   *
+   * <p>This method works under the assumption that the type argument type is solvable; if not, this
+   * will return null.
+   *
+   * @param expression The expression to find the declaration of (method, field)
+   * @return The resolved object, if it exists; null otherwise
+   */
+  public static @Nullable Object tryFindCorrespondingDeclarationForConstraintQualifiedExpression(
```
kelloggm, line 1434: I didn't read this that closely and will need to review this again when I have a bit more mental energy available. I don't see anything obviously wrong for now, though.

kelloggm (commented):
Quick review of commit 02a8153

theron-wang:
@kelloggm Ready for review, including UnsolvedSymbolGenerator

theron-wang:
I know you're away at the moment, but whenever you get a chance again, you can just review whatever I have at the moment.

kelloggm (commented):
I wrote these review comments on Friday and forgot to post them. I'll try to review the current status of the PR now as best I can, time permitting.

kelloggm (commented):
Review of commit ece78bb

kelloggm (commented):
Review of changes in commit 07bc8ea

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
```diff
@@ -2310,6 +2179,236 @@ else if (node instanceof MethodCallExpr methodCall) {
...
   }
 
+  /**
+   * Given a method call expression, try to match its return types to known child classes if this
+   * method declaration behind the call matches all of these requirements:
+   *
+   * <ul>
+   *   <li>The method declaration is unsolved and has a synthetic definition for it
+   *   <li>The unsolved declaring type of this method has a solvable child class with the same
+   *       method signature
```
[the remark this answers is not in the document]
theron-wang: Never mind, I think this test case matches the question I'm asking here: http://github.com/njit-jerse/specimin/tree/main/src/test/resources/voidreturndouble

Overrides are not preserved, but the return types are preserved; void is used instead of a synthetic return type in the test case linked above.

kelloggm (commented):
review of changes in commit ebd54b3

theron-wang:
@coderabbitai configuration

theron-wang:
@coderabbitai full review

theron-wang:
@coderabbitai full review

theron-wang:
@coderabbitai full review

theron-wang:
@coderabbitai full review

theron-wang:
@kelloggm Ready for re-review, with some notes:
* Most file changes (200+ in test files) since your last review are import changes from upgrading JUnit from 4 to 6
* CF-3032 no longer crashes, but it takes much longer (>100s) compared to other evaluation cases
* CF-3025 is now preserved
* MethodRefUndefTest (the one you added sometime this fall/winter) doesn't pass and is ignored for now
* For CI results, I canceled the last two workflows so you can just look at the third-to-last run (the one that passed) for the most recent logs
* All alternates are not generated (only best effort right now)
* I tried running CodeRabbit earlier, but I think there are too many files in this PR so I am not going to run it (525 files vs limit of 150)

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedMethodAlternates.java
```diff
@@ -0,0 +1,531 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.checkerframework.specimin.JavaParserUtil;
+
+/**
+ * /** Given a name, return type, a set of parameters and a set of potential encapsulating classes,
+ * this class allows for alternates of a same field to be generated in different locations. If a
```
kelloggm: nit: there's a typo here - this should say "alternates of the same **method**", not "field"

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
```diff
@@ -0,0 +1,4347 @@
...
+  /** Generates fully qualified names for symbols. */
+  private final FullyQualifiedNameGenerator fullyQualifiedNameGenerator;
+
+  /**
+   * Creates a new UnsolvedSymbolGenerator. Pass in a set of fqns to compilation units for
+   * resolution purposes.
+   *
+   * @param fqnsToCompilationUnits A set of fully-qualified names to compilation units
+   */
+  @SuppressWarnings("method.invocation")
```
kelloggm: nit: When making a warning suppression like this, you should also use the checker key so that it applies only to the checker whose output you intended to suppress. It's also a good idea to write a comment explaining why the suppression is okay - that is, why the warning was a false positive. In this case, since this code was written like a year ago, I think you can just leave a TODO comment.

src/main/java/org/checkerframework/specimin/SpeciminRunner.java
```diff
@@ -245,17 +257,20 @@ public void run() {
    * @param targetMethodNames A set of target method names to be preserved.
    * @param targetFieldNames A set of target field names to be preserved.
    * @param outputDirectory The directory for the output.
-   * @param modularityModel the modularity model
+   * @param ambiguityResolutionPolicy The ambiguity resolution policy.
+   * @param modularityModel The modularity model.
    * @param disableRootValidation whether to disable root validation
    * @throws IOException if there is an exception
    */
+  @SuppressWarnings("UnusedVariable") // Remove once ambiguityResolutionPolicy is used
```
kelloggm: Should this be removed now? I think probably yes; otherwise I'm concerned about why we have `AmbiguityResolutionPolicy` in this PR at all.

kelloggm (commented):
A few low-level comments on files that I happened to look at, but this was by no means a thorough review. I think we should merge this PR based on high-level metrics (like "it preserves everything in specimin-evaluation") rather than based on the low-level details, because I'm confident that this PR _does_ [improve the overall health of the codebase](https://google.github.io/eng-practices/review/reviewer/standard.html) and a low-level review will not be pleasant for anyone: it's too much code to go through, and @theron-wang didn't write most of it recently.

I do think we should do a few small experiments to make sure that this change doesn't cause serious regressions, and I am concerned that the new ambiguity resolution policies don't have dedicated tests. So, I propose that we do the following before we merge:
* write a few dedicated tests for each of the two new ambiguity resolution policies. These can be adaptations of existing tests (with different expected outputs), and need not be comprehensive - the standard I'm looking for here is "if we write code in the future that fundamentally breaks one of the other policies, we'll notice in CI". As far as I can tell, there aren't any new tests that use the new ambiguity resolution policies, but if I've missed them in this huge PR just let me know @theron-wang and point me at them.
* run a few experiments to be sure that this change doesn't make the percentage of arbitrary methods in arbitrary projects that compile after being Speciminified significantly worse. There is already infrastructure in the repo to do this (we had it running as a CI step for a while, but it's a bit fragile/noisy for that use case). In particular, the file [.github/workflows/check_compilable_percentage.yml](https://github.com/njit-jerse/specimin/blob/main/.github/workflows/check_compilable_percentage.yml) has logic to run this kind of experiment on a GitHub runner, and overall it should be a pretty straightforward activity. Pick a few repos (3 smallish ones of your choice is probably sufficient - I suggest picking ones on the order of 10-20k LoC) and run every method in each through Specimin both before and after this change. As long as the % of results that compile is _not worse_ in the condition with this PR, we can merge it. Let me know if you have any questions or concerns about this plan.

theron-wang:
@kelloggm I forgot about the two ambiguity resolution policies; that's also why you also don't see any tests relating to it in this PR. Do you want me to add it, or should I just remove it altogether?

kelloggm:
> I forgot about the two ambiguity resolution policies; that's also why you also don't see any tests relating to it in this PR

If you haven't actually implemented either of them, it's better to cut the option. I'd like to at least support the "all" ambiguity resolution policy, though - I think it will actually be useful in some cases. But, if it's not actually implemented yet, it's okay if that's in a follow-up PR. If that's what you're going to do, please make an issue so that we don't forget about it (in case you don't get to it this summer).

theron-wang:
@kelloggm 

I ran `check_compilable_percentage_CI` on these repos:
- https://github.com/M3CHR0M4NC3R/plume-util-static
- https://github.com/cbeust/jcommander
- https://github.com/jopt-simple/jopt-simple

CI run on current: https://github.com/theron-wang/specimin/actions/runs/28538258310/job/84605098564 (this is on my personal fork; if you can't see this and would like to, let me know)
CI run on refactor: https://github.com/njit-jerse/specimin/actions/runs/28535005216/job/84594018661

| Metric | Current | Refactor |
| --- | --- | --- |
| Full successes | 897 | 869 |
| Failed minimizations | 24 | 39 |
| Failed compilations | 61 | 74 |
| Validate accuracy | 20 | 39 |

Validate accuracy is higher on the refactor? I don't know why

kelloggm:
@theron-wang these results look fine. Not great - there are slightly more failures with the refactor than without it - but close enough that I don't think the regressions will be problematic. It would be great to try to fix some of these (or at least check if there is some low-hanging fruit), but that doesn't have to happen as part of this PR.

> Validate accuracy is higher on the refactor? I don't know why

I'm also surprised by that, but it's not important - validate accuracy is a relative measure compared to a baseline, and the baseline in the repo is probably out of date (since we're not using this CI step anymore). You can safely ignore it.

kelloggm (approved):
Once the "check compilable percentage" CI logic has been removed again I think this is ready to merge. Thanks for your hard work on this big rewrite!

theron-wang:
Sounds good -- I removed the repositories I added in yesterday and turned off the workflow through GitHub (as it was yesterday as well). I have turned on auto-merge so it should merge in a few minutes.

PR #433, opened 2025-11-05
Change the test executor so that it compares by AST instead of using diff

kelloggm:
This ignores all whitespace in the actual vs expected outputs, making it much easier to write the expected output for tests.

This change also obviates the need for `check_differences.bat` (i.e., it unifies the Windows and Unix test infra), which is a nice bonus.

This change was written with an AI assistant (Gemini 2.5 Pro), which wrote the new code in the test executor. I designed it myself. The prompt was (with only `SpeciminTestExecutor.java` in the context):
```
A lot of tests that use the test executor that I've included in the context are failing right now because of formatting mismatches, especially formatting mismatches related to newlines, between the actual and expected outputs. However, the outputs being tested here are themselves Java programs, so newlines don't matter. For example, one of the currently failing tests looks like this:

diff -r -w -B /var/folders/03/b2tvjr914yb7msv0crx4893w0000gn/T/specimin-test-13726960170581697823/com/example/ManyImports.java /Users/mjk76/Research/specimin/specimin/src/test/resources/unusedimports/expected/com/example/ManyImports.java
7c7,8
< public class ManyImports {}
---
> public class ManyImports {
> }

I want the tests that use this executor to ignore _all_ whitespace, including differences in newlines, as long as the Java programs are equivalent. I think that means that we need to get rid of our reliance on the diff command around line 122 (and the `check_differences.bat` script, which replicates the functionality for Windows). Instead, let's change this so that it reads in the _contents_ of these files and then compares them. The easiest way to do so is probably to use JavaParser, which is already on the classpath for this project: read in both the actual and expected programs, and then compare them at the AST level. Please propose a way to do that.
```

### DIFF
--- a/src/test/java/org/checkerframework/specimin/SpeciminTestExecutor.java
+++ b/src/test/java/org/checkerframework/specimin/SpeciminTestExecutor.java
@@ -141,7 +141,7 @@ public class SpeciminTestExecutor {
                     + actualCu);
           }
         } catch (Exception e) {
-          Assert.fail("Error parsing and comparing files: " + relativePath + "\n" + e);
+          Assert.fail("Error parsing and comparing files: " + relativePath.toString().replace('\\', '/') + "\n" + e);
         }
       }
     }


theron-wang (approved):
LGTM, I just added `replace('\\', '/')` to file path outputs so it's easier to read on Windows (backslashes don't show up so it looked like `comexampleSimple.java` instead of `com/example/Simple.java`)

PR #452, opened 2026-07-06
Fix Specimin crashes relating to JavaParser bugs

theron-wang:
Fixes some of the crashes in #421. Namely, these two cases:

* Resolution of annotation members
* MethodAmbiguityException thrown by JavaParser when there are multiple overloads of a method but there really is only one correct answer

Also, I moved resolution logic into a single class (`Resolver#resolve(Resolvable)`) which handles these cases (alongside some other cases that JavaParser cannot handle), so future work in Specimin should use that method instead of JavaParser's `Resolvable#resolve()`.


kelloggm:
> future work in Specimin should use that method instead of JavaParser's Resolvable#resolve()

We should add a lint rule to the build for this, probably as a custom ErrorProne check (since we use ErrorProne for linting), because the odds that every future PR will remember this are low.

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -5,16 +5,7 @@
...
-import com.github.javaparser.ast.body.CallableDeclaration;
-import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
-import com.github.javaparser.ast.body.ConstructorDeclaration;
-import com.github.javaparser.ast.body.EnumConstantDeclaration;
-import com.github.javaparser.ast.body.FieldDeclaration;
-import com.github.javaparser.ast.body.MethodDeclaration;
-import com.github.javaparser.ast.body.Parameter;
-import com.github.javaparser.ast.body.TypeDeclaration;
-import com.github.javaparser.ast.body.VariableDeclarator;
+import com.github.javaparser.ast.body.*;
```
kelloggm: Please avoid introducing wildcard imports.

kelloggm: (I consider them bad style, since they obscure what classes are actually used. Specimin itself struggles in their presence, for example.)

kelloggm (commented):
New test cases look reasonable to me.

kelloggm (approved):
LGTM

PR #453, opened 2026-07-11
Add support for the `permits` keyword

theron-wang:
This adds support for Java's `permits` keyword, noticed when I was working on the false positives project.

Specifically:
1. `permits` clauses are not removed, and the classes referenced within (if solvable) are included in the slice, similar to how classes in `extends`/`implements` are currently handled
2. Modifies unsolved type generation for types found in `permits` clauses. By default, they generate with two alternates--`final` and `non-sealed`--with best effort preferring `final` unless evidence is found against `final`; it then falls back to `non-sealed`. `sealed` is currently not in this list because this requires finding a valid `permits` clause for that unsolved symbol, which seems difficult to do.

**NOTE:** Merge this PR after #452, because this branch is based off the branch in that PR so I wouldn't have to deal with any merge conflicts later on. **_The last three commits (db29e065e2627096472df40ec641df53b4b5d991, 9ab9cb94014611375e214a9b3e0c24be7ccc3f64, 9e31b9da1cc6e2879926d3ffbaaa88fe9b4be070) are the only ones with changes relevant to this PR,_** so only those should be reviewed in this PR.


src/main/java/org/checkerframework/specimin/unsolved/UnsolvedClassOrInterface.java
```diff
@@ -228,6 +250,22 @@ public String toString(
       // a discussion of the difference.
       sb.append("static ");
     }
+
+    if (sealedness == Sealedness.FINAL) {
+      // final interface is not a thing
```
kelloggm: Should this throw rather than silently continuing?

kelloggm: (That is, would it be better for us Specimin to crash here with an error message that we can use to debug how this situation occurred? In general if there's a situation that we expect to be impossible, it's usually better to fail when if/when we end up in that situation rather than silently continuing with a result we know is incorrect.)

kelloggm (approved):
LGTM overall

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedClassOrInterface.java
theron-wang: Sounds good, I made it crash in these cases.

PR #458, opened 2026-07-15
Bug fixes

theron-wang:
This PR makes four changes:
* Static wildcard imports are not removed if they import from a used class
* Constructors are preserved/created with a minimal `super` call when its superclass does not have a parameterless constructor
* Support for record classes (behavior: only used members are preserved, and non-canonical constructors are trimmed down to a minimal `this` call to the canonical constructor)
* Method preservation for "adjacent" classes, where a used type implementing an interface with a known method that must be preserved also has to preserve that method, even if that method is never actually called by that specific used type, but is called by another used type that implements the same interface (see MustImplementMethodsAdjacentClassTest for an example)

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
+++ b/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
@@ -39,7 +39,6 @@ import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.type.TypeParameter;
 import com.github.javaparser.resolution.MethodUsage;
 import com.github.javaparser.resolution.Resolvable;
-import com.github.javaparser.resolution.TypeSolver;
 import com.github.javaparser.resolution.UnsolvedSymbolException;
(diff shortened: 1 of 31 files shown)


kelloggm (changes requested):
I did not review the code itself, only the tests. The three new tests look good, but there isn't a test for the change to the behavior with static imports. Please add one.

A PR like this one that has independent components, though, really ought to be separate PRs. That is, I'd expect there to be four PRs for these changes, not just one, that correspond exactly to the four bullet points in your PR description. The reason is a practical one: as a reviewer on a PR like this one that mixes up four concerns, I would have to carefully separate apart the concerns as I review, hunting around for all of the changes related to each separate change. This significantly reduces the value of the review: I will spend more of my review time trying to figure out what things are related to each other rather than actually critically evaluating the code. If there were instead four separate PRs, I wouldn't have to do that. So, please split this PR up into four separate PRs, one for each change, and then I'll review those separately.

PR #459, opened 2026-07-15
Static import fix

theron-wang:
Fixes:
* Static wildcard imports are not removed if they import from a used class


kelloggm (approved):
LGTM

PR #460, opened 2026-07-15
Ensure constructors

theron-wang:
Fixes:
* Constructors are preserved/created with a minimal super call when its superclass does not have a parameterless constructor

Once #459 merges and this PR is reviewed, set base to branch main and let the CI run again.

(I based this off of static-import-fix instead of main because I figured it would avoid merge conflicts down the line once this is merged)

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
+++ b/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
@@ -39,7 +39,6 @@ import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.type.TypeParameter;
 import com.github.javaparser.resolution.MethodUsage;
 import com.github.javaparser.resolution.Resolvable;
-import com.github.javaparser.resolution.TypeSolver;
 import com.github.javaparser.resolution.UnsolvedSymbolException;
 import com.github.javaparser.resolution.declarations.AssociableToAST;
 import com.github.javaparser.resolution.declarations.ResolvedConstructorDeclaration;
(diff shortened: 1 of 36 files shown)


kelloggm:
> (I based this off of static-import-fix instead of main because I figured it would avoid merge conflicts down the line once this is merged)

Unless the PRs touch nearby lines, this won't matter. It's better to base everything off of main unless the changes directly interact with each other.

src/main/java/org/checkerframework/specimin/SpeciminTypeSolvers.java
```diff
@@ -0,0 +1,153 @@
...
+   * is no way to modify the cache otherwise. Call this method when you have modified the AST and
+   * you cannot have a stale cache for that object.
+   *
+   * <p>We must manually set the cache, because clearing it is not enough. In JavaParserTypeSolver,
+   * it simply reparses the original source file, which does not contain any AST changes made during
+   * Specimin's run.
+   *
+   * @param fqn The type to clear from the cache
+   * @param resolvedReferenceTypeDeclaration The declaration to override in the cache. MUST BE THE
+   *     OBJECT RETURNED FROM {@link TypeDeclaration#resolve()}; OTHERWISE, NOTHING WILL BE CHANGED.
```
kelloggm: Given this requirement, would it be better to take a `TypeDeclaration` and call `resolve()` here rather than relying on the user to follow these instructions? Even if that's mildly less efficient (since we'll probably end up calling `resolve` twice), I think removing the need for the caller to reason about where exactly the `ResolvedReferenceTypeDeclaration` came from is probably worth it.

src/main/java/org/checkerframework/specimin/JavaParserUtil.java
```diff
@@ -1995,7 +1972,28 @@ public static Node findAttachedNode(
...
+    NodeWithSimpleName<?> detachedParentWithName =
+        detachedParent instanceof NodeWithSimpleName<?>
+            ? (NodeWithSimpleName<?>) detachedParent
+            : null;
+
+    return attached
+        .findFirst(
+            nodeClass,
+            // These simple heuristics don't guarantee that a similar but incorrect node is not
+            // selected, but they make it very rare.
```
kelloggm, line 1993: A more precise description of when and why, exactly, an incorrect node would be selected by these heuristics would be more helpful here. Think about who will read this comment in the future: someone who is debugging a case where the heuristics _did_ actually fail is the most likely! So, you should explain all you can about the circumstances in which you expect the heuristics to fail (and why) to help that imaginary future developer with debugging.

kelloggm (approved):
Overall looks good, just a few comments. I don't need to review again.

PR #461, opened 2026-07-15
Record classes

theron-wang:
Fixes:
* Support for record classes (behavior: only used members are preserved, and non-canonical constructors are trimmed down to a minimal `this` call to the canonical constructor)

Once #460 merges and this PR is reviewed, set base to branch main and let the CI run again.

(I based this off of ensure-constructors instead of main because I figured it would avoid merge conflicts down the line once this is merged)

### DIFF
--- a/src/main/java/org/checkerframework/specimin/Resolver.java
+++ b/src/main/java/org/checkerframework/specimin/Resolver.java
@@ -1,20 +1,29 @@
 package org.checkerframework.specimin;
 
+import com.github.javaparser.ast.AccessSpecifier;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Node;
-import com.github.javaparser.ast.body.CallableDeclaration;
+import com.github.javaparser.ast.body.MethodDeclaration;
+import com.github.javaparser.ast.body.Parameter;
+import com.github.javaparser.ast.body.TypeDeclaration;
 import com.github.javaparser.ast.expr.Expression;
 import com.github.javaparser.ast.expr.MethodCallExpr;
 import com.github.javaparser.ast.nodeTypes.NodeWithArguments;
+import com.github.javaparser.ast.nodeTypes.NodeWithParameters;
 import com.github.javaparser.resolution.MethodAmbiguityException;
 import com.github.javaparser.resolution.Resolvable;
 import com.github.javaparser.resolution.UnsolvedSymbolException;
 import com.github.javaparser.resolution.declarations.ResolvedAnnotationDeclaration;
 import com.github.javaparser.resolution.declarations.ResolvedAnnotationMemberDeclaration;
 import com.github.javaparser.resolution.declarations.ResolvedMethodDeclaration;
+import com.github.javaparser.resolution.declarations.ResolvedParameterDeclaration;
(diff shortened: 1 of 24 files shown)


src/main/java/org/checkerframework/specimin/Resolver.java
```diff
@@ -265,6 +279,75 @@ public static void setFqnToCompilationUnitMap(
...
+    }
+
+    if (methodCallExpr.getArguments().isNonEmpty()) {
+      return null;
+    }
+
+    // JavaParser bug: cannot resolve MethodCallExpr if it represents a record member
+
+    // We have to use our own implementation, ResolvedRecordMemberDeclaration, since JavaParser
+    // throws an IllegalStateException anytime we try to resolve the original parameter.
```
kelloggm, line 302: Should we upstream this fix into JavaParser?

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
```diff
@@ -825,32 +826,21 @@ private void handleNameExpr(NameExpr nameExpr, List<UnsolvedSymbolAlternates<?>>
   private void handleMethodCallExpr(
       MethodCallExpr methodCall, List<UnsolvedSymbolAlternates<?>> result) {
     ResolvedMethodDeclaration resolvedMethodDeclaration = Resolver.resolve(methodCall);
-    if (resolvedMethodDeclaration != null) {
-      if (JavaParserUtil.tryResolveNodeIfInAnonymousClass(methodCall) != null) {
-        return;
```
kelloggm, line 830: I don't understand why it's safe to remove this logic or the similar logic around line 3200. I don't see similar special cases for e.g., method calls in anonymous classes (as here) somewhere else in this PR, so it seems like you really are removing this logic. Is that because it was never actually being triggered here? Or is there some other reason?

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
theron-wang: This was code I forgot to remove when adding in `Resolver#resolve` from a few PRs ago. Since `Resolver#resolve` already calls this code, I figured it would be better to remove it here.

kelloggm: Ok, that's a good reason. In general, a change like that should be in a separate PR with a brief explanation that links back to the PR that made the change that makes it safe, but I'm okay with just merging it as part of this one for expediency.

kelloggm (approved):

src/main/java/org/checkerframework/specimin/Resolver.java
theron-wang: I'm not sure if we can, since this is more of a workaround than an actual fix. JavaParser technically already handles record members, but it throws an unrelated IllegalStateException (symbol resolution not configured) so I think the behavior that's fixed here is a side-effect of another JavaParser bug.

PR #462, opened 2026-07-15
Add must implement method support for "adjacent" classes

theron-wang:
Fixes:
* Method preservation for "adjacent" classes, where a used type implementing an interface with a known method that must be preserved also has to preserve that method, even if that method is never actually called by that specific used type, but is called by another used type that implements the same interface (see MustImplementMethodsAdjacentClassTest for an example)

Once #461 merges and this PR is reviewed, set base to branch main and let the CI run again.

(I based this off of record-classes instead of main because I figured it would avoid merge conflicts down the line once this is merged)


kelloggm (approved):
LGTM

PR #468, opened 2026-07-21
Add test for #402

kelloggm:
This PR supersedes #410. The new implementation gets this case correct, so this PR just adds it to the regression test suite.


theron-wang (approved):
Just the CodeRabbit comment, all else LGTM

PR #469, opened 2026-07-21
Fix crash on unsolved argument to JDK constructor

kelloggm:
Fixes the crash in #405. The actual test case for #405 still doesn't compile, but I think that's a separate problem. I've manually confirmed that the test case fails with the appropriate crash without the fix commit.

This PR is a further experiment in AI-assisted software engineering. All of the code in this PR was initially authored by Claude, but I've reviewed it and it seems reasonable to me. Here is its change summary:

In UnsolvedSymbolGenerator.java, a constructor call to a fully-known class (e.g. a JDK type like IllegalStateException) whose argument type was unsolvable was being routed down the "synthetic constructor" path. That path assumes the constructor's declaring type is one of Specimin's synthetic symbols, so it looked up a scope that doesn't exist and crashed with Scope not created when it should've been.

Two guards were added in the unresolvable-args branch:
1. isKnownConstructorDeclaringType(node) — if the constructor's declaring type resolves (JDK/known class), there is no synthetic constructor to generate, so return EMPTY.
2. If the synthetic constructor's alternates can't be located (because argument types were refined after the constructor was first generated), return EMPTY gracefully instead of throwing.

The new test (JdkCtorUnsolvedArgTest + jdkctorunsolvedarg/)

A minimal reproduction distilled from the real crash: a target method that throws new IllegalStateException("bad value: " + field) where field has an unsolvable type. This exercises exactly the fixed path.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
+++ b/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
@@ -1042,6 +1042,9 @@ public class JavaParserUtil {
   /**
    * Checks if the given name is in ALL_CAPS, as is the standard convention for constants. Heuristic
    * for detecting likely enum constants.
+   *
+   * @param simpleName the name to check
+   * @return true if the name matches the standard Java convention for constants, false otherwise
    */
   public static boolean isProbablyAConstant(String simpleName) {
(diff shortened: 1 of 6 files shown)


theron-wang (approved):
LGTM

PR #471, opened 2026-07-24
Switch Checker Framework to run during CI instead of on all builds

kelloggm:
I'm getting frustrated with long build times, and I've been running local tests with the CF off by default for awhile.


theron-wang (approved):
LGTM -- I have also been disabling this locally.

Issue #404, opened 2025-03-12
Ambiguous Method Call crash


M3CHR0M4NC3R:
I uncovered another crash in specimin; here is the stack trace. I created the "wrappedNode" branch with a reproducer test of the same name.
```
Exception in thread "main" com.github.javaparser.resolution.MethodAmbiguityException: Ambiguous method call: cannot find a most applicable method: JavaParserMethodDeclaration{wrappedNode=/**
* Convenience method for reading all entries accessible via
* this iterator
*
* @return List of entries read (same as passed-in argument)
*/
public <L extends List<? super T>> L readAll(L resultList) throws JacksonException {
throw new java.lang.Error();
}, typeSolver=com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver@156b88f5}, JavaParserMethodDeclaration{wrappedNode=/**
* Convenience method for reading all entries accessible via
* this iterator
*/
public <C extends Collection<? super T>> C readAll(C results) throws JacksonException {
throw new java.lang.Error();
}, typeSolver=com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver@156b88f5}
at com.github.javaparser.resolution.logic.MethodResolutionLogic.findMostApplicable(MethodResolutionLogic.java:706)
at com.github.javaparser.resolution.logic.MethodResolutionLogic.findMostApplicable(MethodResolutionLogic.java:612)
at com.github.javaparser.symbolsolver.javaparsermodel.contexts.JavaParserTypeDeclarationAdapter.solveMethod(JavaParserTypeDeclarationAdapter.java:303)
at com.github.javaparser.symbolsolver.javaparsermodel.contexts.ClassOrInterfaceDeclarationContext.solveMethod(ClassOrInterfaceDeclarationContext.java:106)
at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.solveMethod(JavaParserClassDeclaration.java:336)
at com.github.javaparser.resolution.logic.MethodResolutionLogic.solveMethodInType(MethodResolutionLogic.java:909)
at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethod(MethodCallExprContext.java:170)
at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:296)
[...]

PR #472, opened 2026-07-28
Test case for issue 404

kelloggm:
Crash in #404 is no longer present after the refactor; this PR adds a regression test.


theron-wang (approved):
LGTM, though I’m not sure if “wrappedNode” is a good name for this test

kelloggm:
> LGTM, though I’m not sure if “wrappedNode” is a good name for this test

agreed, but I didn't have any better ideas (and that's what the original reporter named it)

PR #474, opened 2026-07-28
When a synthetic exception is used in such a way that the compiler will require it to be unchecked, make it extend Error instead of Exception

kelloggm:
The important difference in the test case's results here is that without this change, `MathIllegalArgumentException` would extend `java.lang.Exception` instead of `java.lang.Error`. However, the input file makes it clear that the exception must be unchecked for the code to compile: `withNanStrategy` calls the (throwing!) constructor without a catch block and without declaring that it also throws.

This change should strictly improve Specimin's "compilable output" performance.

All code changes in this PR were generated by Claude and reviewed by me. (And I caught a bug in its design, so doing so was worthwhile.) The test case was on the same branch as the `wrappedNode` test for #404, but doesn't seem to have its own issue as far as I can tell.


src/test/java/org/checkerframework/specimin/WithNaNStrategyTest.java
```diff
@@ -0,0 +1,15 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.jupiter.api.Test;
+
+/** TODO */
```
theron-wang: Does this need to be un-TODOed?

theron-wang (changes requested):
LGTM, but there seems to be a TODO comment in the test case that maybe you forgot to add?

kelloggm:
@theron-wang please approve this so that it will auto-merge when CI finishes.

theron-wang (approved):

PR #475, opened 2026-07-28
If a class extends an unmodifiable class (e.g., from the JDK) without a zero-argument constructor, we must preserve a compatible constructor or the output won't be compilable

kelloggm:
Test cases from the `entryreader` branch (#393).

### DIFF
--- a/src/test/resources/nozeroargctor/input/com/example/SomeOtherClass.java
+++ b/src/test/resources/nozeroargctor/input/com/example/SomeOtherClass.java
@@ -1,0 +1,7 @@
+package com.example;
+
+class SomeOtherClass {
+    public SomeOtherClass(Object object) {
+
+    }
+}
--- a/src/test/resources/nozeroargctorint/input/com/example/Simple.java
+++ b/src/test/resources/nozeroargctorint/input/com/example/Simple.java
@@ -1,0 +1,13 @@
+package com.example;
+
+class Simple extends SomeOtherClass {
+    // Target method.
+    void bar() {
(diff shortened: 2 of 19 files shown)


src/test/resources/nozeroargctorint/input/com/example/Simple.java
```diff
@@ -0,0 +1,13 @@
+package com.example;
+
+class Simple extends SomeOtherClass {
+    // Target method.
+    void bar() {
+
+    }
+
+    // This needs to be preserved, because otherwise the class won't compile.
```
theron-wang, line 9: This isn't true, since the target method doesn't contain a call to SomeOtherClass's constructor, so that constructor isn't included in the output anyway (so this constructor can also be safely removed). I think you meant to include a call to new SomeOtherClass(0); in bar()?

src/test/resources/nozeroargctor/input/com/example/SomeOtherClass.java
```diff
@@ -0,0 +1,7 @@
+package com.example;
+
+class SomeOtherClass {
+    public SomeOtherClass(Object object) {
```
theron-wang, line 4: This is similar to my other comment for NoZeroArgCtorInt; see that

theron-wang (changes requested):
Some comments about the new test cases, and I think CodeRabbit's comment on Slicer.java is worth looking at.

theron-wang (approved):
LGTM

Issue #386, opened 2025-01-15
Failing to include a second static field in a synthetic type, causing compilation failure


kelloggm:
Test case is called `StaticImportField2Test` on branch `twostaticfields`.

Test input:
```
package com.example;

import org.first.Foo;

class Simple {
    int test() {
        return Foo.COUNT.AMOUNT;
    }
}
```

The problem is that the synthetic class created for `COUNT` doesn't contain an `AMOUNT` field.

kelloggm:
Real example:
```
./gradlew run -PskipCheckerFramework --args='--outputDirectory "/Users/mjk76/Downloads/Archive/tmux-out" --root "/Users/mjk76/Downloads/Archive/termux-app/termux-shared/src/main/java" --\
targetFile "com/termux/shared/file/filesystem/FileAttributes.java" --targetMethod "com.termux.shared.file.filesystem.FileAttributes#loadFromStructStat(StructStat)"'
```

kelloggm:
Confirmed this now works after the refactor.

PR #476, opened 2026-07-28
Test case for #386

kelloggm:
I do find it odd that the test case doesn't actually contain a static import, despite the name. Should we rename it, leave it as it is, or just delete it (i.e., leave this PR unmerged)? I am not sure.

I tested the actual example in #386 (i.e., the tmux target that's listed in the comment) and it works just fine post-refactor.


theron-wang:
I think we should rename it. I'm pretty sure we do not have a test case like this anyway, so it should be helpful to keep.

src/test/java/org/checkerframework/specimin/StaticImportField2Test.java
```diff
@@ -0,0 +1,15 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.jupiter.api.Test;
+
+/** This test checks if Specimin can handle an unsolved field two layers deep in a static import. */
```
theron-wang: If we rename it, this javadoc should be changed as well

theron-wang (approved):
Otherwise LGTM

PR #477, opened 2026-07-28
Add support for inferring that methods throw an exception based on their use in a try clause

### LINKED ISSUE #385: Failing to preserve thrown checked exception when the target method catches the exception
Test case is called `MustThrowTest`, on branch `mustthrow`.

Test input:
```
package com.example;

import org.example.Foo;
import org.example.CustomException;

class Simple {
    // Target method.
    void bar() {
        try {
            Foo.stuff();
        } catch (CustomException custom) {
            // do nothing
        }
    }
}
```

The generated `Foo.java` doesn't add `throws CustomException` to the signature of the `Foo.stuff()` method, which causes a compilation error when the minimized code is compiled.

kelloggm:
Fixes #385. I'm not thrilled with the fact that it's possible for this code to arbitrarily assign the exception to the first in a long list of method calls, but I don't think that can be avoided in the worst case. It still feels like a dangerous ambiguity to leave around, though. I _think_ I've added enough support here so that when we do implement a "generate all alternates" mode, it'll be possible to generate all combinations of choices of where the exception should be thrown.

I'm not sure if there is another method that implements the same functionality as `getSelfAndSupertypeFqns`, but I wouldn't be surprised if there is. The closest I could find is `FullyQualifiedNameGenerator#getFQNsOfAllUnresolvableParents`, but that's not quite what I needed. Should this method be in that class, even if it's not a duplicate of anything that's there?

This change should monotonically increase the fraction of Specimin's outputs that compile.


src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
```diff
@@ -3583,6 +3600,294 @@ private boolean isMethodABuiltInThrowableMethod(
...
+   * @param resolved the resolved type, or null if the type is not solvable
+   * @param fqns the potential fully-qualified names of the type
+   * @return the fully-qualified names of the type and of its supertypes
+   */
+  private Set<String> getSelfAndSupertypeFqns(@Nullable ResolvedType resolved, Set<String> fqns) {
+    Set<String> result = new LinkedHashSet<>(fqns);
+
+    if (resolved != null && resolved.isReferenceType()) {
+      ResolvedReferenceType referenceType = resolved.asReferenceType();
+      result.add(referenceType.getQualifiedName());
```
theron-wang: JavaParserUtil contains two methods that may be useful here: `getAllSolvableAncestors` and `getAllJDKAncestors`, since the former handles cases where getAllAncestors may throw. They take in `TypeDeclaration<?>`s as input, but that should be as easy as calling `tryFindAttachedNode` or `findAttachedNode` on referenceType.getTypeDeclaration().

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
```diff
@@ -3583,6 +3600,294 @@ private boolean isMethodABuiltInThrowableMethod(
...
+          worklist.add(jdkSuperclass);
+        }
+        continue;
+      }
+
+      if (!(generatedSymbols.get(fqn) instanceof UnsolvedClassOrInterfaceAlternates type)) {
+        continue;
+      }
+
+      MemberType extended = type.getAlternates().get(0).getExtendedType();
```
theron-wang: It's been a while since I wrote the extending logic, but if I remember correctly, this call to getExtendedType() will always be null. This is because `UnsolvedSymbolGenerator#createAlternatesBasedOnSuperTypeRelationships` sets the alternates' extended types, but this is only called by `UnsolvedSymbolGenerator#generateAllAlternatesBasedOnSuperTypeRelationships` after `addInformation` is complete (but `getSelfAndSupertypeFqns` will likely only be called only during the `addInformation` stage).

A safer thing to do may be to write some method in `UnsolvedClassOrInterfaceAlternates` and get all the keys where the values are EXTENDS from `UnsolvedClassOrInterfaceAlternates#superTypeRelationships`.

theron-wang (changes requested):
Overall LGTM, just two comments

theron-wang:
> I'm not sure if there is another method that implements the same functionality as `getSelfAndSupertypeFqns`, but I wouldn't be surprised if there is. The closest I could find is `FullyQualifiedNameGenerator#getFQNsOfAllUnresolvableParents`, but that's not quite what I needed. Should this method be in that class, even if it's not a duplicate of anything that's there?

I left a comment about this, but you're right that there is not. I do agree that this method should be moved to FullyQualifiedNameGenerator, however.

kelloggm:
> I do agree that this method should be moved to FullyQualifiedNameGenerator, however.

I didn't do this because `getSelfAndSupertypeFqns` does use the `generatedSymbols` field of `UnsolvedSymbolGenerator` to skip non-synthetic types. While we could pass that in, I feel like that use does make this an unsolved symbol generator thing. I've updated the documentation to explain it, though.

kelloggm:
> > I do agree that this method should be moved to FullyQualifiedNameGenerator, however.
> 
> I didn't do this because `getSelfAndSupertypeFqns` does use the `generatedSymbols` field of `UnsolvedSymbolGenerator` to skip non-synthetic types. While we could pass that in, I feel like that use does make this an unsolved symbol generator thing. I've updated the documentation to explain it, though.

I posted this too soon, before reading your other comments carefully. I am going to end up deleting that part based on something else you wrote, so I will end up moving it.

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
kelloggm: So I tested whether this logic was actually doing anything, and it's not. So, I decided to just simplify this code. I think the fix you're suggesting here is _possible_, but I don't think it's _necessary_, because I think `getAllJDKAncestors` makes it unnecessary. So, I decided it was better to just remove it, and the tests still pass, so I'm happy with that.

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
```diff
@@ -2368,6 +2368,34 @@ private static FullyQualifiedNameSet getSimpleNameOfFunctionalInterface(
...
+  /**
+   * Computes the fully-qualified names of a type together with those of all of its supertypes, on a
+   * best-effort basis. Either the resolved type or the set of potential fully-qualified names may
+   * be used, depending on whether the type is solvable.
+   *
+   * @param resolved the resolved type, or null if the type is not solvable
+   * @param fqns the potential fully-qualified names of the type
+   * @return the fully-qualified names of the type and of its supertypes
+   */
+  public static Set<String> getFQNsOfSelfAndSupertypes(
```
theron-wang, line 2380: since this now only calls `getAllJDKAncestors` perhaps this method should be renamed to something like getFQNsOfSelfAndJdkSupertypes (and the javadoc too, accordingly)?

theron-wang (approved):
LGTM, just one nitpick

src/main/java/org/checkerframework/specimin/unsolved/FullyQualifiedNameGenerator.java
kelloggm: > perhaps this method should be renamed to something like getFQNsOfSelfAndJdkSupertypes

I think just "supertypes" is more clear to indicate all of the supertypes, whether they are in the JDK or not. I'll add a note to the javadoc, though, that JDK supertypes are included.

Issue #480, opened 2026-08-04
Bad interaction between casting and context causes non-compilable input


kelloggm:
Two new failing test cases on branch `castandcontext`.

Specimin reconciles multiple use sites correctly in general; it is specifically adding a cast site that degrades the result. Adding any cast, even (Object), collapses the return type to Object and breaks the assignment. As a result, the output is not compilable.

There is reconciliation logic, and Object is a defensible least-upper-bound of the two candidate types. The bug is subtler: a cast site shouldn't contribute a type to be LUB'd at all, because a cast absorbs any operand type. It should be treated as unconstraining. The addSuperType(Set.of(lub)) calls around UnsolvedSymbolGenerator.java:3507 are a likely place for the defect that causes the problem, though I haven't confirmed that's the path taken. Any debugging effort should probably start there.

PR #481, opened 2026-08-04
Ensure that Specimin produces a compilable output when an expression whose type cannot be solved appears as the operand of a cast expression

kelloggm:
Note that this is orthogonal to #480: I checked both before and after this fix and #480 exists on both sides.

This is a partial fix for the compilation failure of the test case from #403 (more hopefully coming!).


theron-wang (approved):
LGTM

PR #482, opened 2026-08-04
A synthetic class assigned to a variable whose synthetic type is used with a wildcard type argument should still be a subtype of that type

kelloggm:
Previously, this caused non-compilable output because the `extends` clause ended up empty, which isn't correct.

Discovered as part of my investigation into why the example in #403 didn't compile.


theron-wang (approved):
LGTM

PR #483, opened 2026-08-04
Fix incorrect logic when a wildcard type has a lower bound

kelloggm:
Discovered while working on #482. One of the tests in this PR (`WildcardSupertypeBoundedTest.java`) was supposed to be a test for #482, but it also required us to fix this bug. The other test in this PR is an actual regression test for this bug.


src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
```diff
@@ -4080,9 +4080,11 @@ private void handleLHSAndRHSRelationship(
 
           // ? extends with ? extends; there is no ? extends with ? super
           if (isUpperBound) {
+            // Foo<? extends Bound> = Foo<A> tells us that A is a subtype of Bound.
             handleLHSAndRHSRelationship(Set.of(bound), rhsTypeParameters, () -> null);
           } else {
-            handleLHSAndRHSRelationship(rhsTypeParameters, rhsTypes, () -> null);
+            // Foo<? super Bound> = Foo<A> tells us that Bound is a subtype of A.
```
theron-wang, line 4086: Could you add a test case where the RHS also has type arguments to demonstrate this?

theron-wang (changes requested):
Overall LGTM, just a request for another test case

theron-wang (approved):
LGTM

PR #485, opened 2026-08-05
Improve compilability of input programs involving casts

### LINKED ISSUE #480: Bad interaction between casting and context causes non-compilable input

kelloggm:
Fixes #480 and extends the fix in #481, which was in retrospect woefully incomplete. The key to this fix is adding a cast case to `addInformation` that shares code with the logic for `instanceof` for the case when the input program is casting to a type that Specimin will generate, because the JLS rule for their legality is identical; the rest of the changes are mostly just enabling that change.

There is a downside to this change: it results in some extra synthetic types being kept around (e.g., in `Issue403Test`). I'm okay with the mild loss of precision in exchange for reducing the number of non-compiling outputs.


theron-wang (approved):
LGTM

Just one comment: `Slicer` traverses the statements of a method from bottom to top, so the output will _probably_ be different in the "castandcontext" tests if you swap `Bar b = f.get()`; with the line containing the cast. I think Specimin would set the return type to `Bar` in that case and avoid `GetReturnType` altogether, but I'm not 100% sure; it may be worth a test case but it's not a big deal.

kelloggm:
> Just one comment: Slicer traverses the statements of a method from bottom to top, so the output will probably be different in the "castandcontext" tests if you swap Bar b = f.get(); with the line containing the cast. I think Specimin would set the return type to Bar in that case and avoid GetReturnType altogether, but I'm not 100% sure; it may be worth a test case but it's not a big deal.

Yeah, that ambiguity is the reason that I wanted to use this version of the test case and not the one that you mention here - it's important from my perspective that we both document the limitation (so we aren't surprised by it later) and also write down why it happens. Ideally, in the future we can remove the ordering dependency between constraints, but it's not a huge deal.

PR #486, opened 2026-08-05
Use of a variable in a string concatenation does not necessarily mean that its type is compatible with String

kelloggm:
If at least one argument to a `+` operator is a String, then the other argument's type is unconstrained: the compiler introduces an implicit call to `toString()` (JLS 15.18.1: "If only one operand expression is of type String, then string conversion (§5.1.11) is performed on the other operand to produce a string at run time.").

This required a change to the expected output of one existing test, which depended on Specimin's old behavior of constraining _both_ arguments of a string concatenation to be `java.lang.String`. But, doing that causes compilability issues if the variable is used in other contexts (which is often the case in real code).


theron-wang (approved):
LGTM

PR #501, opened 2026-08-12
Fix logic for deciding whether a lambda is void- or value-compatible

kelloggm:
This is one of the blockers of Specimin's output on #442 compiling.


kelloggm:
Ready to merge once CI passes.

PR #503, opened 2026-08-13
Improve conformance to JLS type variable substitution rules

kelloggm:
The non-compiling output from #442 showed that Specimin's rules for type variables were incomplete, and subsequent testing (most of which led to test cases in this PR) showed that non-compilable output was common. The underlying problem was that Specimin was treating type variable names just like type names, which is not correct in general; there were a few cases that were handled, but the majority of type variable substitution cases were not covered. The implementation in this PR is a lot more thorough, and should in general prevent type variables from "leaking" into calling contexts of various kinds.


kelloggm:
This is ready to merge when it is green, unless coderabbit finds anything else we need to address.

PR #512, opened 2026-08-18
NullType should not constrain the type of a variable

### LINKED ISSUE #511: Specimin crashes when an unresolved receiver is initialized to `null`
When an unresolved receiver (for instance caused by a dependency outside the root directory) is assigned a null value, and then later dereferenced, specimin crashes when attempting to identify the type of the receiver for the dereference.

Tested on `commit 697c0c1`

### Source code
`root/src/main/java/org/example/Target.java`:
```java
package org.example;

class Target {
    void target() {
        Missing value = null;
        value.unknownMethod();
    }
}
```

`root/dependency-src/main/java/org/example/Missing.java`:
```java
package org.example;

public class Missing {
    public void unknownMethod() {}
}
```
Invoking specimin causes a crash:
```bash
./gradlew run --args='--root "/home/cenaras/phd/riverside-project/agent-false-positive-playground/specimin-nulltype-reproducer/src/main/java/" --targetFile org/example/Target.java --targetMethod org.example.Target#target() --outputDirectory /home/cenaras/phd/riverside-project/agent-false-positive-playground/specimin-nulltype-reproducer/build/specimin-out'

> Task :run FAILED
Exception in thread "main" java.lang.UnsupportedOperationException: com.github.javaparser.resolution.model.typesystem.NullType@75db5df9 is not a Reference Type
        at com.github.javaparser.resolution.types.ResolvedType.asReferenceType(ResolvedType.java:140)
        at org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator.matchMethodReturnTypesToKnownChildClasses(UnsolvedSymbolGenerator.java:3631)
        at org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator.addInformation(UnsolvedSymbolGenerator.java:2718)
        at org.checkerframework.specimin.Slicer.buildSlice(Slicer.java:188)
        at org.checkerframework.specimin.Slicer.slice(Slicer.java:145)
        at org.checkerframework.specimin.SpeciminRunner.performMinimizationImpl(SpeciminRunner.java:302)
        at org.checkerframework.specimin.SpeciminRunner.performMinimization(SpeciminRunner.java:184)
        at org.checkerframework.specimin.SpeciminRunner.main(SpeciminRunner.java:106)
```

### Likely cause
Codex reports the following likely cause:
```
`matchMethodReturnTypesToKnownChildClasses` collects candidate types for the receiver of an
unresolved method call. The `null` initializer contributes JavaParser's `NullType` to that set.
The method then calls `asReferenceType()` on every candidate without first checking
`type.isReferenceType()`.
```

kelloggm:
Fixes #511

### DIFF
--- a/src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
+++ b/src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
@@ -3622,6 +3622,19 @@ public class UnsolvedSymbolGenerator {
   }
 
   /**
+   * Returns true if the type of a value assigned to a variable tells us anything about that
+   * variable's type. The null type does not: {@code null} is assignable to every reference type
+   * (JLS 4.1), so a {@code null} initializer or assignment is compatible with any declared type.
+   *
+   * @param type the resolved type of an initializer or of an assigned value, or null if it could
+   *     not be resolved
+   * @return true if type constrains the assigned variable's type
+   */
+  private static boolean constrainsVariableType(@Nullable ResolvedType type) {
+    return type != null && !type.isNull();
+  }
+
+  /**
    * Given a method call expression, try to match its return types to known child classes if this
    * method declaration behind the call matches all of these requirements:
    *
@@ -3759,7 +3772,7 @@ public class UnsolvedSymbolGenerator {
             ResolvedType resolvedType =
                 Resolver.calculateResolvedType(varDecl.getInitializer().get());
 
-            if (resolvedType == null) {
+            if (!constrainsVariableType(resolvedType)) {
               continue;
             }
 
@@ -3776,7 +3789,7 @@ public class UnsolvedSymbolGenerator {
             && assignExpr.getTarget().toString().equals(scope.toString())) {
           ResolvedType resolvedType = Resolver.calculateResolvedType(assignExpr.getValue());
 
-          if (resolvedType != null) {
+          if (constrainsVariableType(resolvedType)) {
             potentialTypes.add(resolvedType);
           }
         }
@@ -3792,8 +3805,8 @@ public class UnsolvedSymbolGenerator {
         // Check to see if any of these contain the same method signature; if so, we can
         // update the return type of the current generated one to match it
 
-        // Must be a reference type: if it were not, the method would be solvable, which we
-        // checked already
+        // Must be a reference type: the null type was filtered out above, and for any other
+        // non-reference type the method would be solvable, which we checked already
         ResolvedReferenceType refType = type.asReferenceType();
 
         // The type must also be a user-defined class, not a built-in Java class. This means we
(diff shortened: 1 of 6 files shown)


src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
```diff
@@ -3776,7 +3789,7 @@ private void matchMethodReturnTypesToKnownChildClasses(MethodCallExpr methodCall
             && assignExpr.getTarget().toString().equals(scope.toString())) {
           ResolvedType resolvedType = Resolver.calculateResolvedType(assignExpr.getValue());
 
-          if (resolvedType != null) {
+          if (constrainsVariableType(resolvedType)) {
```
Cenaras, line 3795: `resolvedType != null` stips the `@Nullable` type off of `resolvedType`, whereas `constrainsVariableType(arg)` doesn't give any guarantees from the type system alone (no `EnsuresNonNull(#1)` annotation).

Cenaras (commented):
`:compileJava` fails for me:
```
> Task :compileJava FAILED
/home/cenaras/phd/riverside-project/specimin/src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java:3779: error: [argument] incompatible argument for parameter arg0 of Set.add.
            potentialTypes.add(resolvedType);
                               ^
  found   : @Initialized @Nullable ResolvedType
  required: @Initialized @NonNull ResolvedType
/home/cenaras/phd/riverside-project/specimin/src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java:3793: error: [argument] incompatible argument for parameter arg0 of Set.add.
            potentialTypes.add(resolvedType);
                               ^
  found   : @Initialized @Nullable ResolvedType
  required: @Initialized @NonNull ResolvedType
2 errors
```

Likely because of the `resolvedType != null` check being replaced with a method call without guarantees about the argument

src/main/java/org/checkerframework/specimin/unsolved/UnsolvedSymbolGenerator.java
kelloggm: Yes, sorry, this was my fault for not running the CF locally before I pushed.

kelloggm: Should be fixed now, but if you want to test it locally on your actual example you can pass `-PskipCheckerFramework` to gradle to skip the typechecker during the build.

kelloggm: (which will also be much faster)

Cenaras (approved):
Fixes the issue for my real-world benchmark as well now.

Issue #520, opened 2026-08-24
RuntimeException: Field scope types are not yet created


Cenaras:
### Input
`reproducer/input/retrofit2/RequestFactory.java`
```java
package retrofit2;

import javax.annotation.Nullable;

final class RequestFactory {
  @Nullable private final ParameterHandler<?>[] parameterHandlers;

  RequestFactory(Builder builder) {
    parameterHandlers = builder.parameterHandlers;
  }

  void create(Object[] args) {
    @SuppressWarnings("unchecked")
    ParameterHandler<Object>[] handlers = (ParameterHandler<Object>[]) parameterHandlers;
    if (args.length != handlers.length) {
      throw new IllegalArgumentException();
    }
  }

  static final class Builder {
    @Nullable ParameterHandler<?>[] parameterHandlers;

    RequestFactory build() {
      parameterHandlers = new ParameterHandler<?>[1];
      return new RequestFactory(this);
    }
  }
}
```

`reproducer/input/javax/annotation/Nullable.java`
```java
package javax.annotation;

public @interface Nullable {}
```

### Executed Command And Crash
```sh
./gradlew run --args="--root path/to/reproducer/retrofit-crash/input --targetFile retrofit2/RequestFactory.java --targetMethod 'retrofit2.RequestFactory#create(Object[])' --outputDirectory /tmp/retrofit-specimin-reproducer"

> Task :run FAILED
Exception in thread "main" java.lang.RuntimeException: Field scope types are not yet created; FQNs: [retrofit2.ParameterHandler[]]
        at org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator.handleFieldAccessExpr(UnsolvedSymbolGenerator.java:670)
        at org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator.inferContextImpl(UnsolvedSymbolGenerator.java:201)
        at org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator.inferContext(UnsolvedSymbolGenerator.java:158)
        at org.checkerframework.specimin.Slicer.handleElement(Slicer.java:275)
        at org.checkerframework.specimin.Slicer.buildSlice(Slicer.java:180)
        at org.checkerframework.specimin.Slicer.slice(Slicer.java:145)
        at org.checkerframework.specimin.SpeciminRunner.performMinimizationImpl(SpeciminRunner.java:302)
[...]

PR #529, opened 2026-08-24
Don't try to generate array types

kelloggm:
Should fix the crash in #520, and also several similar crashes related to how array names are handled (see the test cases).


Cenaras (approved):
This fixes the error in the reproducer. On the real benchmark specimin crashes with: `erasedFqns cannot contain a wildcard; use parameter wildcard instead`. I believe this is fixed in https://github.com/njit-jerse/specimin/pull/531

Issue #522, opened 2026-08-24
IllegalArgumentException: erasedFqns cannot contain a wildcard; use parameter wildcard instead


Cenaras:
### Input command
```sh
./gradlew run --args="--root /home/cenaras/phd/riverside-project/nullness-benchmarks/retrofit/retrofit/src/main/java --targetFile retrofit2/RequestFactory.java --targetMethod 'retrofit2.RequestFactory.Builder#build()' --targetMethod 'retrofit2.RequestFactory#RequestFactory(Builder)' --targetField 'retrofit2.RequestFactory#httpMethod' --outputDirectory /tmp/retrofit-error-10-specimin-reproducer"

> Task :run FAILED
Exception in thread "main" java.lang.IllegalArgumentException: erasedFqns cannot contain a wildcard; use parameter wildcard instead: [retrofit2.ParameterHandler<?>[]]
        at org.checkerframework.specimin.unsolved.FullyQualifiedNameSet.<init>(FullyQualifiedNameSet.java:39)
        at org.checkerframework.specimin.unsolved.FullyQualifiedNameSet.<init>(FullyQualifiedNameSet.java:68)
        at org.checkerframework.specimin.unsolved.FullyQualifiedNameSet.<init>(FullyQualifiedNameSet.java:88)
        at org.checkerframework.specimin.unsolved.FullyQualifiedNameSet.<init>(FullyQualifiedNameSet.java:97)
        at org.checkerframework.specimin.unsolved.FullyQualifiedNameGenerator.getFQNsForResolvedType(FullyQualifiedNameGenerator.java:1345)
        at org.checkerframework.specimin.unsolved.FullyQualifiedNameGenerator.getFQNsForExpressionTypeImpl(FullyQualifiedNameGenerator.java:519)
        at org.checkerframework.specimin.unsolved.FullyQualifiedNameGenerator.getFQNsForExpressionType(FullyQualifiedNameGenerator.java:446)
        at org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator.addInformation(UnsolvedSymbolGenerator.java:2949)
        at org.checkerframework.specimin.Slicer.buildSlice(Slicer.java:188)
        at org.checkerframework.specimin.Slicer.slice(Slicer.java:145)
        at org.checkerframework.specimin.SpeciminRunner.performMinimizationImpl(SpeciminRunner.java:302)
[...]

PR #531, opened 2026-08-24
Make `FullyQualifiedNameGenerator#getFQNsForResolvedType` exhaustive over all possible type-shapes

kelloggm:
Should fix the crash in #522, which hit the new array type branch. Added the rest of the cases (based on `ResolvedType`'s subtypes) while I was here; this set of conditionals _should_ have been exhaustive in the first place.


Cenaras (approved):
Fixes the crash on the real benchmark. Minified output compiles as well.

Issue #490, opened 2026-08-07
Annotation arguments must be non-null compile-time constants for minified code


Cenaras:
### The issue
Since fields are initialized to `null` during minification, annotations that are passed such a field value causes compilation failure, as annotations expect compile-time constants and do not permit null values.

### Input
`target/src/main/java/org/example/Target.java`
```java
package org.example;

import external.Header;

public class Target {
  private static final String HEADER_NAME = "X-Header";

  @Header(HEADER_NAME)
  public void target() {}
}
```

`dependency/src/main/java/external/Header.java`
```java
package external;

public @interface header {
  string value();
}
```
### Command
I invoke specimin on `--targetFile org/example/Target.java --targetMethod org.example.Target#target()`
```bash
SPECIMIN_ROOT=/home/cenaras/phd/riverside-project/specimin
SPECIMIN_OUTPUT=/home/cenaras/phd/riverside-project/nullness-benchmarks/eureka/specimin-reproducers/error-18/null-annotation-constant/build/specimin-out
TARGET_ROOT=/home/cenaras/phd/riverside-project/nullness-benchmarks/eureka/specimin-reproducers/error-18/null-annotation-constant/target/src/main/java
GENERATED_CLASSES=/home/cenaras/phd/riverside-project/nullness-benchmarks/eureka/specimin-reproducers/error-18/null-annotation-constant/build/generated-classes


SPECIMIN_ARGS="--outputDirectory $SPECIMIN_OUTPUT --root $TARGET_ROOT --targetFile org/example/Target.java --targetMethod org.example.Target#target()"
(
  cd "$SPECIMIN_ROOT"
  ./gradlew run --args="$SPECIMIN_ARGS"
)

echo "Compiling the generated output (expected to fail: annotation value is not constant)"
javac -d "$GENERATED_CLASSES" $(find "$SPECIMIN_OUTPUT" -name '*.java')
```
And get the following output:
```bash
BUILD SUCCESSFUL in 939ms
10 actionable tasks: 1 executed, 9 up-to-date
Consider enabling configuration cache to speed up this build: https://docs.gradle.org/9.6.1/userguide/configuration_cache_enabling.html
Compiling the generated output (expected to fail: annotation value is not constant)
/home/cenaras/phd/riverside-project/nullness-benchmarks/eureka/specimin-reproducers/error-18/null-annotation-constant/build/specimin-out/org/example/Target.java:9: error: element value must be a constant expression
  @Header(HEADER_NAME)
[...]

PR #534, opened 2026-08-25
Synthetic annotation arguments need to be special-cased when they appear by name, not just when they appear as field accesses

### LINKED ISSUE #524: Specimin produces non-compiling output
### Input

Directory: `input/common/src/main/java/`

`com/netflix/conductor/common/metadata/tasks/TaskType.java`:
```java
package com.netflix.conductor.common.metadata.tasks;

public enum TaskType {
  INLINE;

  public static final String TASK_TYPE_INLINE = "INLINE";
}
```

`org/springframework/stereotype/Component.java`:
```java
package org.springframework.stereotype;

import java.lang.annotation.ElementType;
import java.lang.annotation.Target;

@Target(ElementType.TYPE)
public @interface Component {
  String value();
}
```

File `input/core/src/main/java/com/nextflix/conductor/core/execution/tasks/Inline.java`:
```java
package com.netflix.conductor.core.execution.tasks;

import static com.netflix.conductor.common.metadata.tasks.TaskType.TASK_TYPE_INLINE;

import org.springframework.stereotype.Component;

@Component(TASK_TYPE_INLINE)
public class Inline {
  public void checkEvaluatorType(String evaluatorType) {
    if (evaluatorType == null) {
      throw new IllegalArgumentException();
    }
  }
}
```

### Reproducing the error:
`javac` compiles:
```sh
╰─➤  javac $(find . -name "*.java") && echo $?
0
```

Running `specimin`:
```sh
./gradlew run --args="--root /home/cenaras/phd/riverside-project/agent-false-positive-playground/specimin-reproducer/conductor-error-15/input/core/src/main/java --targetFile com/netflix/conductor/core/execution/tasks/Inline.java --targetMethod 'com.netflix.conductor.core.execution.tasks.Inline#checkEvaluatorType(String)' --outputDirectory /tmp/conductor-error-15-specimin"
```
Compiling the generated output fails:
```sh
javac -d /tmp/conductor-error-15-generated-classes $(find /tmp/conductor-error-15-specimin -name '*.java' | sort)
/tmp/conductor-error-15-specimin/com/netflix/conductor/core/execution/tasks/Inline.java:7: error: annotation value not of an allowable type
@Component(TASK_TYPE_INLINE)
           ^
/tmp/conductor-error-15-specimin/org/springframework/stereotype/Component.java:5: error: invalid type for annotation interface element
    public com.netflix.conductor.common.metadata.tasks.ComNetflixConductorCommonMetadataTasksTaskTypeTASK_TYPE_INLINESyntheticType value();
                                                      ^
2 errors
```

The `@ComponentTask(TASK_TYPE_INLINE)` constant is transformed from a `String` into a synthetic type.

This seems related to https://github.com/njit-jerse/specimin/issues/490 but the constant string here lives in another module outside the root scope and thus escapes the fix in that PR. Since synthetic types are used, compilation fails

kelloggm:
Fixes #524. Also fixes an adjacent, closely-related problem: synthetic annotation arguments of type String need to be initialized to a constant value (since annotation argument are a constant context; see #490 and #492), which `null` is not. So, add a case for choosing the initializer value that returns `""`.

This PR also includes some refactoring to the code added by #492 to reduce duplication.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaLangUtils.java
+++ b/src/main/java/org/checkerframework/specimin/JavaLangUtils.java
@@ -360,6 +360,20 @@ public final class JavaLangUtils {
   }
 
   /**
+   * Returns true if a variable of the named type can be a <em>constant variable</em> (JLS 4.12.4),
+   * which requires the type to be a primitive type or {@code String}. No other type can be a
+   * constant variable, so a variable of any other type is never usable in a constant context, such
+   * as an annotation argument or a {@code case} label.
+   *
+   * @param type a primitive type name, or a fully-qualified class name. A simple class name is not
+   *     accepted, for the reason given in {@link #isJavaLangString(String)}.
+   * @return true if a constant variable can have this type
+   */
+  public static boolean isConstantVariableType(String type) {
+    return isPrimitive(type) || isJavaLangString(type);
+  }
+
+  /**
    * Converts a primitive to its boxed type (i.e. int --> Integer)
    *
    * @param primitive the primitive type (int, boolean, char, etc.)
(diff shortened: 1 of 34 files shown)


kelloggm:
@coderabbit review

Cenaras (approved):
Fixes the issue on my real benchmark as well

kelloggm:
ready to merge when CI passes

Issue #525, opened 2026-08-24
`java.lang.Exception` is synthesized in enclosing package


Cenaras:
Specimin fails to produce compiling output for the following. 

Method outside provided root that uses `java.lang.Exception` becomes synthesized with `example.Exception` instead of retaining the `java.lang.Exception`.

### Input
 `java-lang-exception/input/add/example/Target.java`:
```java
package example;

import library.Listener;

public final class Target {
  public void target() {
    Listener listener =
        new Listener() {
          @Override
          public void started() throws Exception {}
        };
    listener.toString();
  }
}
```
`java-lang-exception/input/dependency/library/Listener.java`:
```java
package library;

public class Listener {
  public void started() throws Exception {}
}
```

### Reproducing the error
Run the following
```sh
javac -d /tmp/specimin-exception-input \
  $(find /path/to/java-lang-exception/input -name '*.java' | sort)

./gradlew run --args="\
--root /path/to/java-lang-exception/input/app \
--targetFile example/Target.java \
--targetMethod 'example.Target#target()' \
--outputDirectory /tmp/specimin-exception-output" --quiet

javac -d /tmp/specimin-exception-output-classes \
  $(find /tmp/specimin-exception-output -name '*.java' | sort)
```
The first `javac` invocation should succeed. Then after invoking specimin, rerun `javac` to see the error. Expected error is:
```sh
╰─➤  javac -d /tmp/specimin-exception-output-classes \
  $(find /tmp/specimin-exception-output -name '*.java' | sort)

/tmp/specimin-exception-output/library/Listener.java:12: error: cannot find symbol
    public void started() throws example.Exception {
                                        ^
  symbol:   class Exception
  location: package example
1 error
```

kelloggm:
Closed by #535.

PR #535, opened 2026-08-26
Resolve type names in anonymous classes and include implicitly-declared java.lang types in list of possible FQNs for a simple name

kelloggm:
Either of these fixes should, on its own, prevent #525. I implemented both defensively, since I view both as bugs:
* the first fix (in `Resolver.java`) extends a workaround for a JavaParser limitation to types, in addition to expressions. In an anonymous class with an unsolved superclass, JavaParser can't resolve anything. We hoisted expressions out of the anonymous class to check if they're resolvable there; this fix does the same for types. We can't do this for all nodes because not all nodes have the same meaning regardless of source position.
* the second fix (in `FullyQualifiedNameGenerator.java`) explicitly models the possibility that a simple name may refer to a java.lang class. This was just an oversight on our part when writing that logic - there's no reason we shouldn't have already considered the possibility.

I tested both fixes independently and either one would have prevented the miscompilation in #525. Since they're both safe, I see no reason not to merge both.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/Resolver.java
+++ b/src/main/java/org/checkerframework/specimin/Resolver.java
@@ -11,6 +11,7 @@ import com.github.javaparser.ast.expr.MethodCallExpr;
 import com.github.javaparser.ast.expr.MethodReferenceExpr;
 import com.github.javaparser.ast.nodeTypes.NodeWithArguments;
 import com.github.javaparser.ast.nodeTypes.NodeWithParameters;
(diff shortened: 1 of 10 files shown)


kelloggm:
(in one PR rather than two because they share the same test cases - with the first fix in, there's not any way to exercise the second that I can find for now)

Cenaras (approved):
resolved the issue -- javac now produces compiling output

PR #536, opened 2026-08-26
Better model inner class constructors

### LINKED ISSUE #526: Invalid constructor for an unresovled nested class
### Input
`nested-constructor/input/app/example/Target.java`:
```java
package example;

import library.Outer;

public final class Target {
  private Outer.Nested nested;

  public void target() {
    nested = new Outer.Nested();
  }
}
```

`nested-constructor/input/dependency/library/Outer.java`:
```java
package library;

public final class Outer {
  public static final class Nested {
    public Nested() {}
  }
}
```

### Reproducing the error
```sh
javac -d /tmp/specimin-nested-input \
  $(find /path/to/nested-constructor/input -name '*.java' | sort)

./gradlew run --args="\
--root /path/to/nested-constructor/input/app \
--targetFile example/Target.java \
--targetMethod 'example.Target#target()' \
--outputDirectory /tmp/specimin-nested-output" --quiet

javac -d /tmp/specimin-nested-output-classes \
  $(find /tmp/specimin-nested-output -name '*.java' | sort)
```

The second `javac` command (on the minified code) fails with the following error:
```sh
╰─➤  javac -d /tmp/specimin-nested-output-classes \
  $(find /tmp/specimin-nested-output -name '*.java' | sort)

/tmp/specimin-nested-output/library/Outer.java:5: error: <identifier> expected
    public Outer.Nested() {
                       ^
1 error
```

The constructor name must be a simple name `Nested`, not `Outer.Nested`

kelloggm:
Fixes #526.

Two major parts:
* better model how names are represented via the new `QualifiedTypeName` class. This class answers questions about names that can be precisely answered without a classpath, like "what's the first identifier" or "what's the simple name". This is most of what's needed for fixing #526, but we still need a way to distinguish methods and constructors. A simpler version of this PR (in my initial draft) added an `isConstructor` flag to `UnsolvedMethod`, but that's a bad way to do it: constructors and methods aren't quite the same, and that ought to be represented in the type system. That motivates:
* better model unsolved constructors vs methods. This is a pretty significant refactor that splits the old `UnsolvedMethod` and its related classes into three: a shared `UnsolvedCallable` (with most of the old `UnsolvedMethod` logic) and then its subclasses `UnsolvedMethod` and `UnsolvedConstructor`. This change makes the incorrect behavior in #526 no longer representable at all, and is the right thing to do in general - it should make a whole class of similar bugs unrepresentable (assuming I got it right).

Especially because of the second part of the change, this PR feels pretty high-risk to me. So, I'm going to run a few small experiments re:compilability before merging. New tests besides the regression test for #526 mostly cover things that I had to reason through during the second refactor that the existing suite didn't cover.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
+++ b/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
@@ -3178,7 +3178,7 @@ public class JavaParserUtil {
 
     ClassOrInterfaceType asClassOrInterface = scope.asTypeExpr().getType().asClassOrInterfaceType();
 
-    if (asClassOrInterface.getScope().isPresent()
+    if (QualifiedTypeName.of(asClassOrInterface).isQualified()
         || asClassOrInterface.getTypeArguments().isPresent()) {
       // A qualified name or an explicitly parameterized name is unambiguously a type.
       return null;
--- a/src/main/java/org/checkerframework/specimin/QualifiedTypeName.java
+++ b/src/main/java/org/checkerframework/specimin/QualifiedTypeName.java
@@ -12,22 +12,23 @@ import org.checkerframework.checker.signature.qual.ClassGetSimpleName;
 /**
  * A type name decomposed into its identifiers: {@code library.Outer.Nested} becomes {@code
  * [library, Outer, Nested]}. Specifically, this class can answer questions about:
- *  <ul>
- *    <li>the <b>first identifier</b> is the one a single-type-import must be matched against (JLS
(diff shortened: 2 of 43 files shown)


kelloggm:
I ran a small experiment comparing the results of Specimin with this patch and without it (i.e., the current main) using 1,688 randomly-selected methods, constructors, and fields from checkstyle, commons-io, commons-lang, gson, jsoup, junit4, and retrofit; the only regression that I observed was caused by a mistake in my experimental infrastructure (failed to delete a file between different runs), so I'm reasonably confident that this refactoring is safe.

@Cenaras I'm going to merge this, but I'd still appreciate confirmation from you that #526 is fixed by this patch!

PR #545, opened 2026-09-02
Reconcile guesses for annotation arguments with known types

### LINKED ISSUE #538: Specimin generates conflicting annotation value types
Currently specimin generates conflicting annotation value types when one type is resolvable from the root while the other one isn't.

### Input
`error/input/core/src/main/java/org.example/ApplicationsResource.java`:
```java
package org.example;

import external.EurekaAccept;
import external.HeaderParam;

public class ApplicationsResource {
  private static final String HEADER_ACCEPT = "Accept";

  public void getContainers(
      @HeaderParam(HEADER_ACCEPT) String acceptHeader,
      @HeaderParam(EurekaAccept.HTTP_X_EUREKA_ACCEPT) String eurekaAccept) {}
}
```

`error/input/dependency/src/main/java/external/EurekaAccept.java:`
```java
package external;

public final class EurekaAccept {
  public static final String HTTP_X_EUREKA_ACCEPT = "X-Eureka-Accept";

  private EurekaAccept() {}
}
```
`error/input/dependency/src/main/java/external/HeaderParam.java:`
```java
package external;

public @interface HeaderParam {
  String value();
}
```

The source code compiles with `javac` and specimin runs successfully:
```sh
javac -d /tmp/eureka-annotation-input-classes \
  $(find /path/to/conflicting-annotation-value-types/input -name '*.java' | sort)

./gradlew run --args="\
--root /path/to/conflicting-annotation-value-types/input/core/src/main/java \
--targetFile org/example/ApplicationsResource.java \
--targetMethod 'org.example.ApplicationsResource#getContainers(String,String)' \
--outputDirectory /tmp/eureka-annotation-specimin" --quiet
```

However, the generated output does not compile:
```sh
╰─➤  javac -d /tmp/specimin-annotation-constant-output-classes \
  $(find /tmp/eureka-annotation-specimin -name '*.java' | sort) 

/tmp/eureka-annotation-specimin/external/HeaderParam.java:7: error: method value() is already defined in class HeaderParam
    public java.lang.String value();
                            ^
/tmp/eureka-annotation-specimin/org/example/ApplicationsResource.java:11: error: incompatible types: String cannot be converted to EurekaAccept
      @HeaderParam(HEADER_ACCEPT) String acceptHeader,
                   ^
2 errors
```

### The generated specimin code:
The error lies in how the annotation parameters are typed. One is typed as a `String` constant (`HEADER_ACCEPT`), whereas the other gets assigned a synthetic type as it lies outside the root (`EurekaAccept.HTTP_X_EUREKA_ACCEPT`).

This generates two conflicting types for the `HeaderParam` annotation, and two `values()` methods:
```java
package external;
@java.lang.annotation.Target({ java.lang.annotation.ElementType.TYPE_USE, java.lang.annotation.ElementType.TYPE, java.lang.annotation.ElementType.FIELD, java.lang.annotation.ElementType.METHOD, java.lang.annotation.ElementType.PARAMETER, java.lang.annotation.ElementType.CONSTRUCTOR, java.lang.annotation.ElementType.LOCAL_VARIABLE, java.lang.annotation.ElementType.ANNOTATION_TYPE, java.lang.annotation.ElementType.PACKAGE, java.lang.annotation.ElementType.TYPE_PARAMETER})
public @interface HeaderParam {

    public external.EurekaAccept value();

    public java.lang.String value();
}
```
Causing compilation to fail

kelloggm:
Fixes #538. Previously, if Specimin saw an annotation argument that wasn't otherwise constraining (e.g., a name), it would assume it was an enum constant, causing context constraints later on (e.g., that it's a string) to be ignored. With this change, under both possible orderings of "constraining" and "non-constraining" contexts, Specimin will always choose the "constraining" context.


kelloggm:
@Cenaras can you check that this does resolve #538 in situ? I'm going to merge this and close that issue, but please do reopen it if there is something else in the real case that this fix doesn't capture.

PR #546, opened 2026-09-02
Make `Resolver#resolve`'s return type accurately reflect when it doesn't return a compatible instance, and add appropriate checks where it doesn't

### LINKED ISSUE #540: ClassCastException: ReferenceImplType cannot be cast to ResolvedValueDeclaration
### Input
`static-scope-resolved-as-type/input/core/src/main/java/org/example/AtlasData.java`:
```java
package org.example;

import external.Helper;

public class AtlasData {

  public interface Field<T> {
    void parse(T object);
  }

  public static class Page {
    public int width;
    public String name;
  }

  public void load(Page page) {
    final String[] entry = new String[5];
    Field<Page> field =
        new Field<Page>() {
          public void parse(Page page) {
            // The scope `Integer` is a NameExpr denoting a type, not a value.
            page.width = Integer.parseInt(entry[1]);
            // An unsolved call, so a stub is generated and addInformation runs.
            page.name = Helper.normalize(entry[0]);
          }
        };
    field.parse(page);
  }
}
```
`static-scope-resolved-as-type/input/dependency/src/main/java/external/Helper.java`:
```java
package external;

public class Helper {
  public static String normalize(String value) {
    return value.trim();
  }
}
```

### Command
`javac` runs:
```sh
javac -d /tmp/static-scope-input-classes \
  $(find /path/to/static-scope-resolved-as-type/input -name '*.java' | sort)
```

Running specimin crashes:
```sh
./gradlew run --args="\
--root /path/to/static-scope-resolved-as-type/input/core/src/main/java \
--targetFile org/example/AtlasData.java \
--targetMethod 'org.example.AtlasData#load(Page)' \
--outputDirectory /tmp/static-scope-specimin" --quiet

Exception in thread "main" java.lang.ClassCastException: class com.github.javaparser.resolution.model.typesystem.ReferenceTypeImpl cannot be cast to class com.github.javaparser.resolution.declarations.ResolvedValueDeclaration (com.github.javaparser.resolution.model.typesystem.ReferenceTypeImpl and com.github.javaparser.resolution.declarations.ResolvedValueDeclaration are in unnamed module of loader 'app')
        at org.checkerframework.specimin.unsolved.FullyQualifiedNameGenerator.getFQNsForExpressionLocationImpl(FullyQualifiedNameGenerator.java:340)
        at org.checkerframework.specimin.unsolved.FullyQualifiedNameGenerator.getFQNsForExpressionLocation(FullyQualifiedNameGenerator.java:199)
        at org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator.findGeneratedMethodFromMethodCall(UnsolvedSymbolGenerator.java:4712)
        at org.checkerframework.specimin.unsolved.UnsolvedSymbolGenerator.addInformation(UnsolvedSymbolGenerator.java:3211)
        at org.checkerframework.specimin.Slicer.buildSlice(Slicer.java:196)
        at org.checkerframework.specimin.Slicer.slice(Slicer.java:145)
        at org.checkerframework.specimin.SpeciminRunner.performMinimizationImpl(SpeciminRunner.java:302)
        at org.checkerframework.specimin.SpeciminRunner.performMinimization(SpeciminRunner.java:184)
        at org.checkerframework.specimin.SpeciminRunner.main(SpeciminRunner.java:106)

```

kelloggm:
Fixes #540.

`Resolver#resolve` type is `Resolvable<T> -> T` in almost all cases, but there a few exceptions - for example, the implicit constructor on a record resolves to the record itself, which is not in fact a constructor. This change adjusts how `Resolver#resolve` is typed so that for _most_ calls, when it is guaranteed to return a `T`, it still does - but when it doesn't we've overloaded it so that the best, most-specific return type is used. That required 1) adding some `instanceof` checks at call sites where a `T` isn't guaranteed, and 2) adding a regression test that checks that the overload list is correct. #540 was a crash caused by one of the non-`T` cases that just did a cast without a check.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
+++ b/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
@@ -960,9 +960,11 @@ public class JavaParserUtil {
         return resolved instanceof ReflectionFieldDeclaration;
       }
     } else if (expr.isMethodCallExpr()) {
-      ResolvedMethodDeclaration resolvedScopeMethod = Resolver.resolve(expr.asMethodCallExpr());
-
-      if (resolvedScopeMethod == null) {
+      // A call on an annotation member resolves to that member rather than to a method (JLS
+      // 9.6.1), which has no declared type to read here; treat it as unhandled, as for a call
+      // that does not resolve at all.
+      if (!(Resolver.resolve(expr.asMethodCallExpr())
+          instanceof ResolvedMethodDeclaration resolvedScopeMethod)) {
         return false;
       }
 
@@ -1689,9 +1691,9 @@ public class JavaParserUtil {
 
     TypeDeclaration<?> enclosingClass;
 
-    ResolvedEnumConstantDeclaration resolvedDecl = Resolver.resolve(enumConstant);
-
-    if (resolvedDecl == null) {
+    // An enum constant whose arguments are unresolvable resolves to the enum's constructor
+    // instead (see Resolver#WIDENED_NODE_KINDS), which is not a constant to read a type from.
+    if (!(Resolver.resolve(enumConstant) instanceof ResolvedEnumConstantDeclaration resolvedDecl)) {
       return List.of();
     }
     ResolvedType type = resolvedDecl.getType();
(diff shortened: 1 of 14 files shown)


kelloggm:
@Cenaras can you confirm that this change fixed #540?

Cenaras:
Yes, the MWE is fixed with this, and running specimin on the real benchmark now doesn't cause a crash :+1:

Issue #543, opened 2026-09-02
Synthesizes return type does not respect boolean return


Cenaras:
### Input
`synthetic-return-type-ignores-use-site/input/core/src/main/java/org/example/ApplicationContextFactory.java`
```java
package org.example;

import external.AotDetector;

public class ApplicationContextFactory {

  public String createDefaultApplicationContext() {
    if (!AotDetector.useGeneratedArtifacts()) {
      return "annotation-config";
    }
    return "generic";
  }
}
```
`synthetic-return-type-ignores-use-site/input/dependency/src/main/java/external/AotDetector.java`:
```java
package external;
public abstract class AotDetector {
  public static boolean useGeneratedArtifacts() {
    return false;
  }
}
```

### Commands
`javac` passes.  Specimin runs successfully, but the produced code does not compile:
```sh
./gradlew run --args="\
--root /path/to/synthetic-return-type-ignores-use-site/input/core/src/main/java \
--targetFile org/example/ApplicationContextFactory.java \
--targetMethod 'org.example.ApplicationContextFactory#createDefaultApplicationContext()' \
--outputDirectory /tmp/synthetic-return-specimin" --quiet

╰─➤  javac -d /tmp/synthetic-return-output-classes \
  $(find /tmp/synthetic-return-specimin -name '*.java' | sort)

/tmp/synthetic-return-specimin/org/example/ApplicationContextFactory.java:8: error: bad operand type ExternalAotDetectorUseGeneratedArtifactsReturnType for unary operator '!'
    if (!AotDetector.useGeneratedArtifacts()) {
        ^
1 error
```

In particular, the synthesized dependency `useGeneratedArtifacts` is declared as:
```java
public static external.ExternalAotDetectorUseGeneratedArtifactsReturnType useGeneratedArtifacts() {
    throw new java.lang.Error();
}
```

However its use-site expects a boolean: `if (!AotDetector.useGeneratedArtifacts()) { ... }`

kelloggm:
#551 fixed this.

PR #551, opened 2026-09-09
Unary operators constrain their operands

kelloggm:
Fixes issue #543. Specimin just completely lacked handling of unary operators before this - I guess we missed it in our prototype, and it has never come up (somehow).

Incidentally, I fixed a bug related to binary operators (which I discovered while factoring the shared logic between unary and binary operators into a helper): when the two halves of a binary operator were structurally equal (e.g., `x + x`), Specimin would crash. The second regression test in this PR covers that case.


Cenaras (approved):
Resolved the issue

kelloggm:
this is ok to merge once CI is passing

PR #553, opened 2026-09-10
Use "approximate" signatures for abstract methods with unsolved parameters that need to be preserved

kelloggm:
Should fix #547.

The problem was that our reasoning for "must-implement" methods made an incorrect assumption: it assumed that we could resolve the complete signature of such a method (probably because IIRC the first time we encountered the need for must-implement methods was when dealing with JDK interfaces). #547 was caused by a must-implement method with an unsolved parameter type. We actually don't _need_ the complete signature - just the method name and arity - and so we can use a notion of "approximate" signatures (with just the name and arity) instead, which is what this PR does. To accomplish that, we add a set of `getApproximateSignature` methods parallel to the existing `getSignature` methods in `JavaParserUtil`.

PR #556, opened 2026-09-10
Audit for uses of non-deterministicly-ordered-output JavaParser methods in ways that could impact Specimin's output

kelloggm:
In #553, one sub-problem that I had to solve was that we were iterating over a JavaParser method (`ResolvedReferenceTypeDeclaration#getDeclaredMethods`) that returns its results in non-deterministic order - in fact, JavaParser returns different orders each time this method runs (it re-hashes the results and order them by hash). This non-determinism was leaking into Specimin's output.

This PR is the result of an audit of Specimin's use of that method and similar APIs. I found one genuine bug (see the new test case) that can be triggered today, and a number of latent possible bugs (none of which I managed to trigger). The genuine bug required more than just iterating in deterministic order to fix, because we were returning the _first_ match for overloaded methods (in the non-deterministically ordered set...), rather than the most-specific.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
+++ b/src/main/java/org/checkerframework/specimin/JavaParserUtil.java
@@ -2874,9 +2874,8 @@ public class JavaParserUtil {
         }
 
         // An approximately-applicable candidate is one the argument is not actually assignable to.
-        // javac can only have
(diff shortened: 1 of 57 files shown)


kelloggm:
I think this is ready to merge once all tests pass

PR #557, opened 2026-09-11
Static block patch

mushfiqiit:
Previously, if Specimin reduced a static final field with no declaration-site initializer (a "blank final" assigned in a static block instead), it had no way to find that assignment as a dependency. The field would fall through to Slicer's "empty final field" repair, which invents a default value (null, 0, etc.) — silently changing the field's actual value and, for fields whose value matters, breaking behavior in the reduced program.

This patch makes StandardTypeRuleDependencyMap search the declaring class's static initializer blocks for the one that assigns the target field, and follow it as a dependency like any other reachable code, so the real assignment (and whatever it calls) is preserved instead of being replaced with a placeholder. The search is precise: it looks for the specific block that assigns the field in question, so an unrelated field assigned in a separate static block in the same class is left alone and still gets pruned normally if nothing else needs it.

Includes two regression tests adapted from real-world code (greenrobot/EventBus's AndroidComponents):

AndroidComponentsStaticBlockTest — a field assigned in a static block, conditionally, via calls to another class.
PlatformServicesStaticBlockTest — same pattern, plus a second field assigned in its own separate static block, checking that it's correctly excluded rather than swept in along with the target.

### DIFF
--- a/src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
+++ b/src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
@@ -9,10 +9,12 @@ import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.ConstructorDeclaration;
 import com.github.javaparser.ast.body.EnumConstantDeclaration;
 import com.github.javaparser.ast.body.FieldDeclaration;
+import com.github.javaparser.ast.body.InitializerDeclaration;
 import com.github.javaparser.ast.body.MethodDeclaration;
 import com.github.javaparser.ast.body.TypeDeclaration;
 import com.github.javaparser.ast.body.VariableDeclarator;
 import com.github.javaparser.ast.expr.AnnotationExpr;
+import com.github.javaparser.ast.expr.AssignExpr;
 import com.github.javaparser.ast.expr.Expression;
 import com.github.javaparser.ast.expr.ObjectCreationExpr;
 import com.github.javaparser.ast.nodeTypes.NodeWithAnnotations;
@@ -36,6 +38,7 @@ import com.github.javaparser.resolution.declarations.ResolvedMethodDeclaration;
 import com.github.javaparser.resolution.declarations.ResolvedMethodLikeDeclaration;
 import com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration;
 import com.github.javaparser.resolution.declarations.ResolvedTypeParameterDeclaration;
+import com.github.javaparser.resolution.declarations.ResolvedValueDeclaration;
 import com.github.javaparser.resolution.types.ResolvedReferenceType;
 import com.github.javaparser.resolution.types.ResolvedType;
 import com.github.javaparser.symbolsolver.javaparsermodel.declarations.DefaultConstructorDeclaration;
@@ -531,6 +534,45 @@ public class StandardTypeRuleDependencyMap implements TypeRuleDependencyMap {
           && !variableDeclarator.getType().isPrimitiveType()
           && JavaParserUtil.isInConstantContext(node)) {
         elements.add(variableDeclarator.getInitializer().get());
+      }
+
+      // If this static final field has no declaration-site initializer, it must be assigned
+      // somewhere else -- most commonly a static initializer block. Follow that assignment as a
+      // dependency, the same way any other reachable code is followed, instead of leaving it for
+      // Slicer's "empty final field" repair to invent a default value for. That repair still
+      // exists as a fallback for a field this loop can't find an assignment for.
+      if (variableDeclarator.getInitializer().isEmpty() && field.isStatic()) {
+        for (BodyDeclaration<?> member : type.getMembers()) {
+          if (!(member instanceof InitializerDeclaration initializer) || !initializer.isStatic()) {
+            continue;
+          }
+
+          boolean assignsThisField =
+              initializer.getBody().findAll(AssignExpr.class).stream()
+                  .anyMatch(
+                      assignExpr -> {
+                        ResolvedValueDeclaration target = null;
+                        if (assignExpr.getTarget().isFieldAccessExpr()) {
+                          target = Resolver.resolve(assignExpr.getTarget().asFieldAccessExpr());
+                        } else if (assignExpr.getTarget().isNameExpr()) {
+                          target = Resolver.resolve(assignExpr.getTarget().asNameExpr());
+                        }
+
+                        return target != null
+                            && target.isField()
+                            && target
+                                .asField()
+                                .declaringType()
+                                .getQualifiedName()
+                                .equals(resolvedFieldDeclaration.declaringType().getQualifiedName())
+                            && target.getName().equals(resolvedFieldDeclaration.getName());
+                      });
+
+          if (assignsThisField) {
+            elements.add(initializer);
+            break;
+          }
+        }
       }
     }
 
--- a/src/test/java/org/checkerframework/specimin/PlatformServicesStaticBlockTest.java
+++ b/src/test/java/org/checkerframework/specimin/PlatformServicesStaticBlockTest.java
@@ -1,0 +1,24 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Like {@link AndroidComponentsStaticBlockTest}, but with a second {@code static final} field
+ * ({@code bootDiagnostics}) that is assigned in its own, separate static initializer block. Since
+ * that field is never the target and nothing reachable from the target reads it, neither the
+ * field, its static block, nor the detector method it alone depends on should survive slicing.
+ * This checks that following a static initializer block as a dependency is precise: it must find
+ * the specific block that assigns the target field, not sweep in every static block in the class.
+ */
+public class PlatformServicesStaticBlockTest {
+  @Test
+  public void runTest() throws IOException {
+    SpeciminTestExecutor.runTestWithoutJarPaths(
+        "platformservicesstaticblock",
+        new String[] {
+          "org/example/runtime/PlatformServices.java", "org/example/runtime/PlatformDetector.java"
+        },
+        new String[] {"org.example.runtime.PlatformServices#instance"});
+  }
+}
--- a/src/test/java/org/checkerframework/specimin/AndroidComponentsStaticBlockTest.java
+++ b/src/test/java/org/checkerframework/specimin/AndroidComponentsStaticBlockTest.java
@@ -1,0 +1,24 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Regression test adapted from greenrobot/EventBus's {@code AndroidComponents}. The target field,
+ * {@code implementation}, is a {@code static final} field with no declaration-site initializer
+ * that is instead assigned in a static initializer block, conditionally, based on a call to
+ * another class in the same package. This exercises following a static initializer block as a
+ * dependency instead of letting Specimin's "empty final field" repair invent a default value.
+ */
+public class AndroidComponentsStaticBlockTest {
+  @Test
+  public void runTest() throws IOException {
+    SpeciminTestExecutor.runTestWithoutJarPaths(
+        "androidcomponentsstaticblock",
+        new String[] {
+          "org/greenrobot/eventbus/android/AndroidComponents.java",
+          "org/greenrobot/eventbus/android/AndroidDependenciesDetector.java"
+        },
+        new String[] {"org.greenrobot.eventbus.android.AndroidComponents#implementation"});
+  }
+}
--- a/src/test/resources/androidcomponentsstaticblock/expected/org/greenrobot/eventbus/android/AndroidComponents.java
+++ b/src/test/resources/androidcomponentsstaticblock/expected/org/greenrobot/eventbus/android/AndroidComponents.java
@@ -1,0 +1,12 @@
+package org.greenrobot.eventbus.android;
+
+public abstract class AndroidComponents {
+
+    private static final AndroidComponents implementation;
+
+    static {
+        implementation = AndroidDependenciesDetector.isAndroidSDKAvailable()
+            ? AndroidDependenciesDetector.instantiateAndroidComponents()
+            : null;
+    }
+}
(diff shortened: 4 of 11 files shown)


src/test/java/org/checkerframework/specimin/PlatformServicesStaticBlockTest.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Like {@link AndroidComponentsStaticBlockTest}, but with a second {@code static final} field
+ * ({@code bootDiagnostics}) that is assigned in its own, separate static initializer block. Since
+ * that field is never the target and nothing reachable from the target reads it, neither the
+ * field, its static block, nor the detector method it alone depends on should survive slicing.
+ * This checks that following a static initializer block as a dependency is precise: it must find
+ * the specific block that assigns the target field, not sweep in every static block in the class.
+ */
+public class PlatformServicesStaticBlockTest {
```
kelloggm, line 14: The name of this test is confusing, since it refers to _where the test came from_ rather than _what it tests_. Please pick a descriptive name that indicates what the test is supposed to be testing (for both of the new tests). For this one, maybe something like "staticblockunrelatedfieldtest"?

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
```diff
@@ -532,6 +535,45 @@ public List<Node> getRelevantElements(Object resolved, Node node) {
           && JavaParserUtil.isInConstantContext(node)) {
         elements.add(variableDeclarator.getInitializer().get());
       }
+
+      // If this static final field has no declaration-site initializer, it must be assigned
+      // somewhere else -- most commonly a static initializer block. Follow that assignment as a
+      // dependency, the same way any other reachable code is followed, instead of leaving it for
+      // Slicer's "empty final field" repair to invent a default value for. That repair still
+      // exists as a fallback for a field this loop can't find an assignment for.
+      if (variableDeclarator.getInitializer().isEmpty() && field.isStatic()) {
```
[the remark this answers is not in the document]
kelloggm: I agree with CodeRabbit here: there is nothing at this location that constrains this to only final static fields. Add a regression test that shows that this fix is wrong before you make a change, though.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
```diff
@@ -532,6 +535,45 @@ public List<Node> getRelevantElements(Object resolved, Node node) {
...
+        for (BodyDeclaration<?> member : type.getMembers()) {
```
kelloggm, line 545: The logic in this block is not gated on either:
* the field itself being a target - this should only fire when an initializer is otherwise being preserved, not any time that a field is read
* the NullAway modularity model

Both of these are surprises, and should be justified in the PR description if they are really necessary.

For the former, I'm surprised this doesn't break any existing test cases. Write a test case where this logic fires for a field that is only read within a target method, but is not itself a target. Or, make an argument in response to this comment about why that is impossible (perhaps I've mis-remembered what the precondition for this part of the logic is).

For the latter, you need to better motivate the problem. The PR description should explain what's wrong with Specimin's current behavior, since there's no associated issue that this PR is fixing. How do you know that this behavior is wrong for plain javac and Checker Framework checkers, in addition to NullAway? As far as I know from our discussions about this, you've only tested it with NullAway, and it might be something NullAway-specific.

kelloggm (changes requested):
You should also address or dismiss the feedback from CodeRabbit (it's not always right, of course, but because it sometimes comes up with useful feedback it is worthwhile to consider each of its comments).

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
mushfiqiit: I have added the flag to make it execute of NullAway only.

src/test/java/org/checkerframework/specimin/PlatformServicesStaticBlockTest.java
mushfiqiit: I removed the earlier test file and added new test files.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
```diff
@@ -541,7 +548,18 @@ public List<Node> getRelevantElements(Object resolved, Node node) {
       // dependency, the same way any other reachable code is followed, instead of leaving it for
       // Slicer's "empty final field" repair to invent a default value for. That repair still
       // exists as a fallback for a field this loop can't find an assignment for.
-      if (variableDeclarator.getInitializer().isEmpty() && field.isStatic()) {
+      //
+      // Only do this under modularity models that actually observe the difference (see
```
kelloggm: This comment is too verbose. I think you could cut the whole thing.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
```diff
@@ -541,7 +548,18 @@ public List<Node> getRelevantElements(Object resolved, Node node) {
...
+      // ModularityModel#preserveStaticInitializerAssignments for why): the javac/Checker Framework
+      // baseline is unaffected either way, since a repaired default is just as compilable as the
+      // real assignment, and only NullAway's initialization-flow analysis can tell them apart.
+      // This is not gated on the field being a --targetField: any reachable read of the field --
+      // whether it's the target itself or just read from within a target method -- needs the same
+      // real value, for the same reason.
```
kelloggm: I don't find this argument convincing. Specimin doesn't have to preserve the warnings (or lack thereof) on a field that's not a target, since only within targets does it make that guarantee. So, there's no reason for this logic to be used when the field in question is only going to be included in the slice because of a read, in my opinion.

If you really think this is necessary, please provide a test case that shows why.

src/test/java/org/checkerframework/specimin/CheckerFrameworkStaticInitializersNotPreservedTest.java
```diff
@@ -0,0 +1,27 @@
+package org.checkerframework.specimin;
+
+import java.io.IOException;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Counterpart to {@link NullAwayIgnoresUnrelatedStaticInitializerTest} that uses the exact same
```
kelloggm: I'm not convinced that this test is necessary - I think it's covering the same logical partition as `CheckerFrameworkStaticInitializerAssignmentNotPreservedTest` ("with the CF modularity model, the logic isn't used").

src/test/java/org/checkerframework/specimin/NullAwayStaticInitializerAssignmentPreservedTest.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.io.IOException;
+import org.junit.jupiter.api.Test;
+
+/**
+ * This test checks that when running with the NullAway modularity model, a {@code static final}
+ * field with no declaration-site initializer -- assigned instead in a static initializer block,
+ * conditionally, via calls to another class -- has that block followed as a dependency, instead of
+ * Specimin's "empty final field" repair inventing a default value for it. See {@link
+ * org.checkerframework.specimin.modularity.ModularityModel#preserveStaticInitializerAssignments}
```
kelloggm, line 11: Avoid direct links into the code in test descriptions, which shouldn't need to be coupled to the code itself. Instead, explain the test from first principles rather than trying to justify it based on the particulars of the implementation.

kelloggm: (In other words, tests should be understandable stand-alone)

kelloggm (changes requested):

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
mushfiqiit: Removed.

src/test/java/org/checkerframework/specimin/CheckerFrameworkStaticInitializersNotPreservedTest.java
mushfiqiit: The goal of this test is to show that non-target field initialisations do not get included during the slicing.

src/test/java/org/checkerframework/specimin/NullAwayStaticInitializerAssignmentPreservedTest.java
mushfiqiit: I updated it.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
mushfiqiit: I added a testcase "NullAwayFieldReadByTargetMethodPreservedTest.java". I executed NullAway on the Specimin output without the patch applied, where the nonnull field 'greeting' gets initialised to null. It adds a new warning on the output while the input program does not contain that warning upon NullAway checking.

src/test/java/org/checkerframework/specimin/CheckerFrameworkStaticInitializersNotPreservedTest.java
kelloggm: That makes sense to me for the NullAway version of this test, but for the CF/javac version I don't think it makes a difference: initializers shouldn't be preserved, whether they're targets or not. So, maybe we want this test but not `CheckerFrameworkStaticInitializerAssignmentNotPreservedTest`? I do not think we need both.

src/test/java/org/checkerframework/specimin/NullAwayFieldReadByTargetMethodPreservedTest.java
```diff
@@ -0,0 +1,30 @@
...
+ * program would regress: {@code greeting} would fall through to Slicer's ordinary repair and
+ * become {@code private static final String greeting = null;}, while the target method {@code
+ * greet()} -- which Specimin does promise to preserve faithfully -- would still read {@code
+ * return greeting.toUpperCase();} verbatim. Running NullAway on that reduced program would then
+ * report a nullness warning on that dereference, inside the target itself, that does not exist in
+ * the original program, where {@code greeting} is always assigned a real value. The field is not
+ * the target, but the dereference site is: it is a statement of {@code greet()}'s own body, so its
+ * correctness is exactly what Specimin's fidelity guarantee for the target covers.
+ */
+public class NullAwayFieldReadByTargetMethodPreservedTest {
```
kelloggm: I don't think this test's reasoning holds, because NullAway doesn't behave the way the Javadoc says. When NullAway checks a dereference of a field, it uses the field's declared type. It doesn't look at the initializer. I checked this with NullAway 0.14.1 (Error Prone 2.42.0, `AnnotatedPackages=org.example`) on small variants of this fixture:

| Program | NullAway result |
|---|---|
| The input as written | no errors |
| The output the Javadoc predicts: `private static final String greeting = null;`, with `greet()` unchanged | **one error, on the field declaration**: `assigning @Nullable expression to @NonNull field`. No error in `greet()`. |
| The same, plus `String s = null; s.length();` in `greet()` | both errors are reported, so the first error isn't hiding anything in `greet()` |
| `private static final @Nullable String greeting;` set in a `static {}` block from a method call | error at `greeting.toUpperCase()`, so the declared annotation is what counts, whatever the initializer |

So the Javadoc is wrong to say the narrowed version would produce "a nullness warning on that dereference, inside the target itself". To NullAway, `greet()` is the same in the input, in the `= null` version and in the current expected output. The only difference is a new error on the declaration of `greeting`, which isn't a target. Specimin promises behavior preservation for the targets only, so an error outside the target isn't a regression, and it doesn't justify leaving the static-initializer logic ungated on `--targetField`. You should remove this test and add that gate, unless you disagree with this analysis.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
kelloggm: See my comment on that test case; in summary, its claims about NullAway are not correct. Specimin only promises behavior preservation for the targets, and a new warning _outside_ the target is therefore ok. So, this should be gated on `--targetField`, I think.

kelloggm (changes requested):

src/test/java/org/checkerframework/specimin/NullAwayFieldReadByTargetMethodPreservedTest.java
mushfiqiit: I removed the test and added the gate.

src/test/java/org/checkerframework/specimin/CheckerFrameworkStaticInitializersNotPreservedTest.java
mushfiqiit: I removed the test.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
mushfiqiit: I understand now. I gated it on targetField.

src/main/java/org/checkerframework/specimin/StandardTypeRuleDependencyMap.java
```diff
@@ -97,11 +100,16 @@ public class StandardTypeRuleDependencyMap implements TypeRuleDependencyMap {
...
    * @param fqnToCompilationUnits The map of type FQNs to their compilation units.
    * @param modularityModel The modularity model to use.
+   * @param targetFieldNames The fully-qualified names of the target fields, each formatted as
+   *     {@code Qualified.Class#fieldName}.
    */
   public StandardTypeRuleDependencyMap(
-      Map<String, CompilationUnit> fqnToCompilationUnits, ModularityModel modularityModel) {
+      Map<String, CompilationUnit> fqnToCompilationUnits,
+      ModularityModel modularityModel,
+      List<String> targetFieldNames) {
```
kelloggm: The fact that you need to pass this list in here is suspicious. How does a target field's initializer _usually_ get preserved? It must not be handled here, if you have to pass `targetFieldNames`. Check how they are usually preserved and make your intervention there rather than passing additional information around, if you can.

mushfiqiit: I revised the PR and attempted to do the same thing in TargetMemberFinderVisitor.java without any parameter passing.

kelloggm (approved):
LGTM

